PR #2, opened 2020-10-24
CI: add a simple github action

bilelmoussaoui:
Let's see if this builds fine

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -1,0 +1,132 @@
+on:
+  push:
+    branches: [master]
+  pull_request:
+
+name: CI
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    container:
+      image: ubuntu:20.10
+    strategy:
+      matrix:
+        rust:
+          - stable
+          - beta
+          - nightly
+          - "1.40.0"
+    steps:
+      - run: apt-get update -y
+      - run: apt-get install -y libgtk-3-dev libglib2.0-dev libgraphene-1.0-dev git xvfb curl libcairo-gobject2 libcairo2-dev
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: ${{ matrix.rust }}
+          override: true
+      # atk
+      - name: "atk: tests"
+        run: xvfb-run --auto-servernum cargo test --manifest-path atk/Cargo.toml --all-features
+      - name: "atk-sys: tests"
+        run: xvfb-run --auto-servernum cargo test --manifest-path atk/sys/Cargo.toml --all-features
+      - name: "atk: build"
+        run: cargo build --manifest-path atk/Cargo.toml --all-features
+      # cairo
+      - name: "cairo: tests"
+        run: xvfb-run --auto-servernum cargo test --manifest-path cairo/Cargo.toml --features "png,pdf,svg,ps,use_glib,v1_16,freetype,script,xcb,xlib,win32-surface"
+      - name: "cairo-sys: tests"
+        run: xvfb-run --auto-servernum cargo test --manifest-path cairo/sys/Cargo.toml --features "png,pdf,svg,ps,use_glib,v1_16,freetype,script,xcb,xlib,win32-surface"
+      - name: "cairo: build"
+        run: cargo build --manifest-path cairo/Cargo.toml --features "png,pdf,svg,ps,use_glib,v1_16,freetype,script,xcb,xlib,win32-surface"
+      # gdk
+      - name: "gdk: tests"
+        run: xvfb-run --auto-servernum cargo test --manifest-path gdk/Cargo.toml --features v3_24
+      - name: "gdk: build"
+        run: cargo build --manifest-path gdk/Cargo.toml --features v3_24
+      # gdk-pixbuf
+      - name: "gdk-pixbuf: tests"
+        run: xvfb-run --auto-servernum cargo test --manifest-path gdk-pixbuf/Cargo.toml --all-features
+      - name: "gdk-pixbuf-sys: tests"
+        run: xvfb-run --auto-servernum cargo test --manifest-path gdk-pixbuf/sys/Cargo.toml --all-features
+      - name: "gdk-pixbuf: build"
+        run: cargo build --manifest-path gdk-pixbuf/Cargo.toml --all-features
+      # gdkx11
+      - name: "gdkx11: tests"
+        run: xvfb-run --auto-servernum cargo test --manifest-path gdkx11/Cargo.toml --all-features
+      - name: "gdkx11-sys: tests"
+        run: xvfb-run --auto-servernum cargo test --manifest-path gdkx11/sys/Cargo.toml --all-features
+      - name: "gdkx11: build"
+        run: cargo build --manifest-path gdkx11/Cargo.toml --all-features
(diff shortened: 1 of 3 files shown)


bilelmoussaoui:
The tests fails in a very weird way. It complains about a missing header ` gio/gdesktopappinfo.h`. The header is provided by glib2-devel per `dnf provides */gdesktopappinfo.h` which is installed... I'm a bit confused.

@GuillaumeGomez any idea?

bilelmoussaoui:
Ah it actually fails locally as well. Weird

sdroege:
You need to use the `gio-unix-2.0` pkg-config package for getting the right includes to find that (and other) headers.

sdroege:
Now only fails with the following:

```
<command-line>: error: invalid application of ‘sizeof’ to incomplete type ‘GSettingsBackend’ {aka ‘struct _GSettingsBackend’}
tests/layout.c:10:31: note: in expansion of macro ‘ABI_TYPE_NAME’
   10 |     printf("%zu\n%zu", sizeof(ABI_TYPE_NAME), alignof(ABI_TYPE_NAME));
      |                               ^~~~~~~~~~~~~
<command-line>: error: invalid application of ‘__alignof__’ to incomplete type ‘GSettingsBackend’ {aka ‘struct _GSettingsBackend’}
tests/layout.c:10:55: note: in expansion of macro ‘ABI_TYPE_NAME’
   10 |     printf("%zu\n%zu", sizeof(ABI_TYPE_NAME), alignof(ABI_TYPE_NAME));
      |                                                       ^~~~~~~~~~~~~
tests/layout.c: In function ‘main’:
<command-line>: error: ‘GSettingsBackendClass’ undeclared (first use in this function); did you mean ‘GSettingsBackend’?
<command-line>: note: in definition of macro ‘ABI_TYPE_NAME’
<command-line>: note: each undeclared identifier is reported only once for each function it appears in
<command-line>: note: in definition of macro ‘ABI_TYPE_NAME’
test cross_validate_layout_with_c ... FAILED
tests/constant.c: In function ‘main’:
<command-line>: error: ‘G_SETTINGS_BACKEND_EXTENSION_POINT_NAME’ undeclared (first use in this function); did you mean ‘G_TLS_BACKEND_EXTENSION_POINT_NAME’?
<command-line>: note: in definition of macro ‘ABI_CONSTANT_NAME’
<command-line>: note: each undeclared identifier is reported only once for each function it appears in
<command-line>: note: in definition of macro ‘ABI_CONSTANT_NAME’
```

This is a bug in gio: nothing includes `<gio/gsettingsbackend.h>`, and that's also not easy to include because of

```cpp
#error "You must define G_SETTINGS_ENABLE_BACKEND before including <gio/gsettingsbackend.h>."
```

sdroege:
See https://gitlab.gnome.org/GNOME/glib/-/issues/2239

GuillaumeGomez:
Once we're done with this, a big squash will be welcome. :)

sdroege:
You accidentally reverted https://github.com/gtk-rs/gtk-rs/pull/2/commits/db778f30b367733fafee9614ec3599d821ad9fe2, that's why the CI fails now. For this to stay around we need to fix `gir`.

sdroege:
Added it back now so we can see if everything passes now at least.

Can you create an issue for this problem in gir?

Also https://github.com/gtk-rs/gir/issues/979 is another issue that currently has changes to autogenerated code. We should clean those two up ASAP.

sdroege:
Still fails because
```
compilation command "cc" "-Wno-deprecated-declarations" "-D__USE_MINGW_ANSI_STDIO" "-pthread" "-I/usr/include/gio-unix-2.0" "-I/usr/include/libmount" "-I/usr/include/blkid" "-I/usr/include/glib-2.0" "-I/usr/lib/x86_64-linux-gnu/glib-2.0/include" "-DABI_CONSTANT_NAME=G_SETTINGS_BACKEND_EXTENSION_POINT_NAME" "tests/constant.c" "-o" "/tmp/abil2IMTc/constant" failed, exit code: 1
compilation command "cc" "-Wno-deprecated-declarations" "-D__USE_MINGW_ANSI_STDIO" "-pthread" "-I/usr/include/gio-unix-2.0" "-I/usr/include/libmount" "-I/usr/include/blkid" "-I/usr/include/glib-2.0" "-I/usr/lib/x86_64-linux-gnu/glib-2.0/include" "-DABI_TYPE_NAME=GSettingsBackendClass" "tests/layout.c" "-o" "/tmp/abiTlYfoq/layout" failed, exit code: 1
```

That constant and the class struct need to be ignored separately in -sys mode. @bilelmoussaoui you update that?

sdroege:
Ok, did that now. Let's see what fails next

sdroege:
>  Let's see what fails next

Various things in glib. I don't think we ever ran the -sys tests here at all... so maybe let's disable them and create an issue to run them later. Would be good to get some kind of CI in at least so we can progress with this repo.

@GuillaumeGomez @bilelmoussaoui?

bilelmoussaoui:
Sounds good to me, we can enable the tests incrementally for each crate till we get everything working in the future. I will update the merge request accordingly if that sounds good to @GuillaumeGomez

GuillaumeGomez:
I definitely agree! First let's try to have basic checks then we can add more when we have something working.

bilelmoussaoui:
I've updated the PR & opened https://github.com/gtk-rs/gtk-rs/issues/13 to track the various CI improvements we could land in the future

sdroege:
You can keep the non-sys tests running, those were definitely running successfully before with travis.

bilelmoussaoui:
> You can keep the non-sys tests running, those were definitely running successfully before with travis.

Per the travis file, I don't see any "tests" other than docs ones which I don't have any idea how to handle properly now.

sdroege:
Because it was all separate before and not a workspace, each `cargo test` only ran the tests of the individual crates. So the above would've run the `gtk` tests.

You need to run something like [this](https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/blob/0b5cc2c83e0f3a913ce2d822a9e529828374ed92/.gitlab-ci.yml#L193-226) here to test each crate individually and to also set the cargo features as needed. The -sys tests are btw [here](https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/blob/0b5cc2c83e0f3a913ce2d822a9e529828374ed92/.gitlab-ci.yml#L259-292).

sdroege:
Looks good to me, thanks

sdroege:
Let's just add `GSettingsBackend` back for now (see discussion on gitlab). Just remove those two commits :)

bilelmoussaoui:
It seems there's an issue with `cairo_gobject_extend_get_type` as `--all-features` enables the use_glib feature on cairo I guess? I've included libcairo-gobject but that doesn't seem to help.

bilelmoussaoui:
This should pass this time :) I have disabled the -sys tests for the problematic ones & added them to the CI issue,  I will try to tackle those as soon as I can.

GuillaumeGomez:
It passed! \o/ If it's ok for you, I think we can merge.

bilelmoussaoui:
All good, go ahead o/

GuillaumeGomez:
Thanks!

sdroege:
@bilelmoussaoui Can you create an issue for the gio-sys pkg-config problem for gir, so we can solve it properly? :)

PR #6, opened 2020-10-29
Regenerate

sdroege:


gdkx11/sys/Cargo.toml
```diff
@@ -60,6 +60,6 @@ tempfile = "3"
 [features]
 v3_16 = []
 v3_22 = ["v3_16"]
-v3_24 = ["v3_22"]
+v3_24 = ["v3_16"]
```
sdroege, line 63: Not sure what this is about. gir seems to believe there is no `v3_22`. @bilelmoussaoui ?

GuillaumeGomez:
The first regen! \o/ You didn't have any trouble?

sdroege:
> The first regen! \o/ You didn't have any trouble?

No, should I have seen problems? :)

gdkx11/sys/Cargo.toml
bilelmoussaoui: Hmm, it definitely have them here though https://github.com/gtk-rs/gtk-rs/blob/069e2a021a908e84f7e43de258c06ecd75cbb7ac/gdkx11/sys/src/lib.rs#L381-L384

sdroege: Probably because those only come from `Gir.toml`. Needs a fix in gir, I guess. See https://github.com/gtk-rs/gir/issues/979

I've manually changed this part now

GuillaumeGomez (approved):

PR #8, opened 2020-10-29
glib/functions: fix get_charset logic

lucab:
This tweaks `get_charset()` signature and logic to properly return
both the charset label and the UTF-8 boolean.

### DIFF
--- a/glib/src/auto/functions.rs
+++ b/glib/src/auto/functions.rs
@@ -571,18 +571,6 @@ pub fn get_application_name() -> Option<GString> {
     unsafe { from_glib_none(glib_sys::g_get_application_name()) }
 }
 
-pub fn get_charset() -> Option<GString> {
-    unsafe {
-        let mut charset = ptr::null();
-        let ret = from_glib(glib_sys::g_get_charset(&mut charset));
-        if ret {
-            Some(from_glib_none(charset))
-        } else {
-            None
-        }
-    }
-}
-
 pub fn get_codeset() -> GString {
     unsafe { from_glib_full(glib_sys::g_get_codeset()) }
 }
--- a/glib/src/functions.rs
+++ b/glib/src/functions.rs
@@ -1,4 +1,3 @@
-#[cfg(not(windows))]
 use glib_sys;
 #[cfg(any(feature = "v2_58", feature = "dox"))]
 #[cfg(not(windows))]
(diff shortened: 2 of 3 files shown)


lucab:
(The header at the top of this file says it is auto-generated, but I didn't actually see where else the logic I'm patching would be coming from)

sdroege:
There's a `generator.py` at the top-level :)

You'll have to edit the `glib/Gir.toml` to ignore this function (`GLib.*` is the "object" the contains all global functions) and then add a manual implementation to `glib/src/functions.rs` that basically does what you did here

glib/src/auto/functions.rs
```diff
@@ -571,16 +571,15 @@ pub fn get_application_name() -> Option<GString> {
...
-        }
-    }
+/// Obtain the character set for the current locale.
+///
+/// This retuns whether the locale's encoding is UTF-8, and the current
+/// charset if aavailable.
+pub fn get_charset() -> (bool, Option<GString>) {
+    let mut charset = ptr::null();
+    let is_utf8 = unsafe { from_glib(glib_sys::g_get_charset(&mut charset)) };
+    let charset = unsafe { from_glib_none(charset) };
```
sdroege, line 581: Wrap the whole block into a single `unsafe`. They're all conceptually the same unsafe block and the invariants that have to hold are now spanning multiple unsafe blocks. That's not good.

They should be kept as small as possible, but everything before/after the unsafe block should have normal Rust safety rules again so that everything is nicely wrapped up inside the unsafe block.

lucab:
@sdroege thanks for the pointers, I'll massage the PR in that direction.

lucab:
Amended, it should be slightly less nasty now (and I've checked that `generator.py` is not angry about it) :smiley:

glib/src/functions.rs
```diff
@@ -212,3 +210,17 @@ pub fn spawn_async_with_pipes<
...
+/// This returns whether the locale's encoding is UTF-8, and the current
+/// charset if available.
+pub fn get_charset() -> (bool, Option<GString>) {
+    let (is_utf8, charset);
+    let mut out_charset = ptr::null();
+    unsafe {
+        is_utf8 = from_glib(glib_sys::g_get_charset(&mut out_charset));
+        charset = from_glib_none(out_charset);
+    };
+    (is_utf8, charset)
```
sdroege, line 225: Now if you move the tuple inside the `unsafe` block, you can avoid that spurious variable declaration at the top :)

Personally I'd just put the whole function into a single unsafe block. Everything it does is unsafe apart from actually returning the values, but those you have to get out of the unsafe block anyway so you can just return them from there.

Looks good otherwise, thanks :)

lucab:
Amended to reshuffle the `unsafe` block. Thanks again for the prompt feedback and for maintaining this!

sdroege:
Thanks for fixing this, looks good to me now :)

PR #10, opened 2020-10-30
In glib_object_wrapper!, use generic Class type instead defining a struct

ids1024:
Update of https://github.com/gtk-rs/glib/pull/720 with corresponding changes to other crates.

This makes the `$rust_class_name` argument in `glib_wrapper!` optional. If specified, it defines a `type` alias for the class. At least optionally allowing it is required to avoid breaking lots of gir generated code, but perhaps it can be removed after a change to gir.

This removes the `ObjectType::RustClassType` associated type, since it can't be enforced/assumed by the type system that that is an instance of this generic struct.

The `IsClassFor` trait is also removed, superseded by this generic type.

This also changes the `IsSubclassable` trait to be implemented on the object struct rather than the class struct, since there otherwise would be foreign trait on foreign type errors.

### DIFF
--- a/glib/src/object.rs
+++ b/glib/src/object.rs
@@ -52,8 +52,6 @@ pub unsafe trait ObjectType:
     type GlibType: 'static;
     /// type of the FFI Class structure.
     type GlibClassType: 'static;
-    /// type of the Rust Class structure.
-    type RustClassType: 'static;
 
     fn as_object_ref(&self) -> &ObjectRef;
     fn as_ptr(&self) -> *mut Self::GlibType;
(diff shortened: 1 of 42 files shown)


sdroege:
Can you rebase on top of latest master? Then we get CI back :)

glib/src/object.rs
```diff
@@ -2513,6 +2513,11 @@ impl<T: ObjectType> Class<T> {
             }
         }
     }
+
+    /// Gets the contained type struct
+    pub fn inner(&mut self) -> &mut T::GlibClassType {
```
sdroege, line 2518: Instead of that, can we get `AsRef<T>` and `AsMut<T>`? If not, naming them `as_ref()` and `as_mut()` would seem nicer nonetheless

glib/src/object.rs
```diff
@@ -52,8 +52,6 @@ pub unsafe trait ObjectType:
     type GlibType: 'static;
     /// type of the FFI Class structure.
     type GlibClassType: 'static;
-    /// type of the Rust Class structure.
-    type RustClassType: 'static;
```
sdroege, line 56: So far we didn't require there to be a wrapper type for `ObjectSubclass` impls... maybe that's something we can also change in the future as it seems like it would make things easier, or not?

Also it would allow returning the type itself instead of the parent type from `ObjectSubclass::get_instance()` and similar weirdnesses that we have right now.

What do you think @ids1024? We can do that separate from this PR though, no need to block on this here.

gtk/src/subclass/application.rs
```diff
@@ -49,8 +48,8 @@ impl<T: GtkApplicationImpl> GtkApplicationImplExt for T {
     }
 }
 
-unsafe impl<T: GtkApplicationImpl> IsSubclassable<T> for ApplicationClass {
-    fn override_vfuncs(&mut self) {
+unsafe impl<T: GtkApplicationImpl> IsSubclassable<T> for Application {
+    fn override_vfuncs(class: &mut ::glib::object::Class<Self>) {
```
sdroege, line 52: I hope we get custom self types at some point, `self: &mut Class<Self>` (compare `Future::poll`) would look nicer :)

sdroege (commented):
Looks good to me otherwise. Thanks a lot!

gtk/src/subclass/application.rs
ids1024: The unstable arbitrary self types feature in Rust, as implemented, only works when the type implements `Deref` to `self`. So assuming that's how a stable feature would work, I don't think it would help here. Unfortunately.

glib/src/object.rs
ids1024: Good idea. That would probably be better.

glib/src/object.rs
ids1024: That could be useful. Similarly to `get_instance`, all of the `*Impl` traits have methods receiving the parent type, which a lot of implementations will immediately want to downcast.

I've been thinking of that recently. I wonder if subclassing should requiring defining a `glib_wrapper!` and specifying that as an associated type in `ObjectSubclass`. If that doesn't lead to recursive type errors, and doesn't make basic subclassing too hard, it could improve some of these things.

But yeah, that's separate and doesn't need to block this PR.

glib/src/object.rs
ids1024: Though that does have the downside that it would probably require callers to specify the type, since it can't be inferred that `as_ref()` can only work with one type.

gtk/src/subclass/application.rs
sdroege: Ah too bad. We can live with that, I guess

glib/src/object.rs
sdroege: Yeah passing it to the virtual methods was also what I was thinking of, that should be possible then. Can you create an issue for this?

glib/src/object.rs
ids1024: But I think that's fine, since it may be more readable if code using this (which would probably mostly be implementations of `IsSubclassable`) is explicit about the type anyway.

I've changed this.

glib/src/object.rs
ids1024: Opened https://github.com/gtk-rs/gtk-rs/issues/14

sdroege: Thanks!

sdroege:
I'll merge this later today or tomorrow once I've updated `gstreamer-rs` accordingly. Don't want it to fail once this is merged :)

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

glib/src/wrapper.rs
```diff
@@ -329,144 +329,156 @@ macro_rules! glib_wrapper {
...
         $(#[$attr:meta])*
-        pub struct $name:ident(Object<$ffi_name:ty, $rust_class_name:ident>);
+        pub struct $name:ident(Object<$ffi_name:ty $(, $rust_class_name:ident)?>);
 
         match fn {
             get_type => || $get_type_expr:expr,
         }
     ) => {
-        $crate::glib_object_wrapper!(@object [$($attr)*] $name, $ffi_name, ::std::os::raw::c_void, $rust_class_name, @get_type $get_type_expr, @extends [], @implements []);
+        $crate::glib_object_wrapper!(@object [$($attr)*] $name, $ffi_name, ::std::os::raw::c_void, @get_type $get_type_expr, @extends [], @implements []);
```
sdroege, line 338: @ids1024 Can you also do the `gir` change that makes it use the new variants without the explicit class name?

gtk/src/subclass/cell_renderer_text.rs
```diff
@@ -38,11 +37,11 @@ impl<T: CellRendererTextImpl> CellRendererTextImplExt for T {
...
 
-unsafe impl<T: CellRendererTextImpl> IsSubclassable<T> for CellRendererTextClass {
-    fn override_vfuncs(&mut self) {
-        <CellRendererClass as IsSubclassable<T>>::override_vfuncs(self);
+unsafe impl<T: CellRendererTextImpl> IsSubclassable<T> for CellRendererText {
+    fn override_vfuncs(class: &mut ::glib::object::Class<Self>) {
+        <CellRenderer as IsSubclassable<T>>::override_vfuncs(class);
         unsafe {
-            let klass = &mut *(self as *mut Self as *mut gtk_sys::GtkCellRendererTextClass);
+            let klass = &mut *(class as *mut _ as *mut gtk_sys::GtkCellRendererTextClass);
```
sdroege, line 44:
```suggestion
            let klass = &mut *(class.as_mut() as *mut _ as *mut gtk_sys::GtkCellRendererTextClass);
```

Might be nicer but doesn't really matter

sdroege: Especially no wildcard cast anymore :)

GuillaumeGomez: Oh I like it!

sdroege: `sed -E -i 's;klass as \*mut _ ;klass.as_mut() ;' */src/subclass/*.rs`

Changes this automatically btw

sdroege:
> I'll merge this later today or tomorrow once I've updated `gstreamer-rs` accordingly. Don't want it to fail once this is merged :)

That's waiting here now https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/621

gtk/src/subclass/cell_renderer_text.rs
sdroege: Updated the commit so once the CI is through we can merge it here.

ids1024: Actually, that should be equivalent to just `let klass = class.as_mut();`. Which is exactly what I was thinking with the `AsMut` implementation.

https://github.com/gtk-rs/gtk-rs/pull/16

sdroege: Oh even simpler. Thanks!

glib/src/wrapper.rs
ids1024: It doesn't look complicated to update `gir`, but I'm seeing issues running `generator.py` with `gir` master similar to those mentioned in https://github.com/gtk-rs/gtk-rs/pull/11

sdroege: Indeed, something's broken currently. Can you create an issue for gir otherwise?

PR #11, opened 2020-11-01
glib/translate: take advantage of Rust Option & Result types

fengalin:
This is the continuation of https://github.com/gtk-rs/glib/pull/718.

Requires:
- [x] https://github.com/gtk-rs/gir/pull/977
- [x] Update `gir` submodule + Regenerate `atk`, `pango` and `gtk`.
- [x] https://github.com/gdesmott/system-deps/pull/14
- [x] Remove patch section.


glib/src/translate.rs
```diff
@@ -277,6 +345,31 @@ impl ToGlib for Ordering {
     }
 }
 
+/// A Rust type `T` for which `Option<T>` translates to the same glib type as T.
+// The copy bound on `Self::GlibType` is requested by clippy by virtue of:
+// https://rust-lang.github.io/rust-clippy/master/index.html#declare_interior_mutable_const
```
sdroege, line 350: I don't think this makes sense here, but it also doesn't hurt as all types that implement `ToGlib` / `FromGlib` should be `Copy` anyway? I think you can also just require that in those traits directly.

glib/src/translate.rs
```diff
@@ -1081,56 +1174,104 @@ impl FromGlib<glib_sys::gboolean> for bool {
...
-            None
+    fn from_glib(val: G) -> Option<T> {
+        T::try_from_glib(val).ok()
+    }
+}
+
+/// Error type for `TryFromGlib` when the Glib value can be None or invalid.
+#[derive(Debug, Eq, PartialEq)]
+pub enum GlibNoneOrInvalidError<I> {
+    Invalid(I),
```
sdroege, line 1212: Maybe name this `E` instead of `I` to make it a little bit clearer that this is supposed to be an error type. Maybe even require `std::error::Error` for it

gtk/src/entry.rs
```diff
@@ -0,0 +1,21 @@
...
+
+impl<O: IsA<Entry>> EntryExtManual for O {
+    fn get_invisible_char(&self) -> Option<char> {
+        let ret = unsafe { gtk_sys::gtk_entry_get_invisible_char(self.as_ref().to_glib_none().0) };
+
+        if ret == 0 {
+            return None;
+        }
+
+        Some(TryFrom::try_from(ret).unwrap())
```
sdroege, line 19:
```suggestion
        Some(TryFrom::try_from(ret).expect("blabla"))
```

?

sdroege: Also couldn't this use the `FromGlib` impl here instead?

gtk/src/text_iter.rs
```diff
@@ -16,4 +17,14 @@ impl TextIter {
...
     }
+
+    pub fn get_char(&self) -> Option<char> {
+        let ret = unsafe { gtk_sys::gtk_text_iter_get_char(self.to_glib_none().0) };
+
+        if ret == 0 {
+            return None;
+        }
+
+        Some(TryFrom::try_from(ret).unwrap())
```
sdroege, line 28: Same question here

sdroege (commented):
Looks generally good to me. Please also rebase on latest master when fixing the remaining comments as that will then give us CI :)

sdroege:
@GuillaumeGomez Your opinion?

GuillaumeGomez:
Looks good to me. Just the remaining `unwrap()` to be replaced and the other comments you made. ;)

Thanks @fengalin !

glib/src/translate.rs
fengalin: Ok!

glib/src/translate.rs
fengalin: I didn't want to force the `Error` trait on implementers initially, but on a second thought, I guess that would be better so that we can use this in interfaces with `dyn Error`.

I keep believing the name `I` is useful here to indicate that it applies to the `Invalid` variant. With the `Error` bound, don't you think this would make it clear that this is an `Error`?

gtk/src/entry.rs
fengalin: The idea was to not implement `FromGlib<u32>` for `char` since this could fail :)

glib/src/translate.rs
sdroege: With the `Error` bound it seems fine

gtk/src/entry.rs
sdroege: Ok!

sdroege (approved):
Looks good to me. @GuillaumeGomez please first merge the gir PR, and then this here needs to get gir updated and can go in too

sdroege:
@fengalin Please update with https://github.com/gtk-rs/gir/pull/977 and then let's get this in :)

fengalin:
It's much more convenient to contribute with the new single repo :+1:

sdroege:
> It's much more convenient to contribute with the new single repo

Please use `generator.py` to regenerate everything :)

sdroege:
@fengalin Good to go? I'll merge once the gstreamer-rs MR is also ready so we don't have things broken for too long :)

fengalin:
> @fengalin Good to go?

No, it's not. Using `generator.py` showed 2 issues: `gobject` versions / features mapping are not generated in `glib-sys`'s `Cargo.toml` and I believe `system-deps` is buggy. The lib name is not inherited when overriding with a feature. I'm on those issues ATM...

fengalin:
Just pushed a commit so that you can see what I had to do so far to workaround some issues since running `generator.py` (probably not directly related to the script :)). See the commit message for details. I'll investigate some more.

fengalin:
I was able to build the examples with all features using last commit (and `system-deps` patch).

However, the `glib/sys` tests don't pass.

sdroege:
> However, the `glib/sys` tests don't pass.

How does it fail?

fengalin:
> > However, the `glib/sys` tests don't pass.
> 
> How does it fail?

Nevermind, I think it's related to mingw not being installed. It failed on master with the same errors.

fengalin:
@GuillaumeGomez @sdroege I think it's finally good to go!

gdkx11/sys/Cargo.toml
```diff
@@ -60,6 +64,6 @@ tempfile = "3"
 [features]
 v3_16 = []
 v3_22 = ["v3_16"]
-v3_24 = ["v3_22"]
+v3_24 = ["v3_16"]
```
sdroege, line 67: This is wrong, see https://github.com/gtk-rs/gir/issues/979

sdroege:
Otherwise seems good to me

gdkx11/sys/Cargo.toml
fengalin: The never ending story...

sdroege: That broke at some point in the past and for the last update we manually changed it back here. Do the same for now unless you feel like working on gir now :) It would seem better to first get this out of the way than to include more unrelated issues

fengalin: That's what I think too. I'd like to resume work on the original changes that motivated this PR.

fengalin:
All green!

glib/src/translate.rs
```diff
@@ -20,6 +20,72 @@
 //!     }
 //! ```
 //!
+//! Implementing `OptionToGlib` on a Rust type `T` allows specifying a sentinel to indicate
```
GuillaumeGomez, line 23: Please use intra link (now that they're stable... :p). So for instance:

```
[`OptionToGlib`]
```

glib/src/translate.rs
```diff
@@ -20,6 +20,72 @@
...
+//! a `None` value and auto-implementing `FromGlib` for `Option<T>`, which would not be
```
GuillaumeGomez, line 24: Same here (and for all the comment ;) ).

glib/src/translate.rs
```diff
@@ -1081,56 +1166,104 @@ impl FromGlib<glib_sys::gboolean> for bool {
...
+    //#[allow(clippy::match_like_matches_macro)]
+    pub fn is_none(&self) -> bool {
+        match self {
+            GlibNoneOrInvalidError::None => true,
+            _ => false,
+        }
+    }
+
+    /// Returns `true` if `self` is the `Invalid` variant.
+    // FIXME `matches!` was introduced in rustc 1.42.0, current MSRV is 1.40.0
```
GuillaumeGomez, line 1226: No, we will increase it in the next gtk-rs release. Please make the update and don't worry about it. (In a following PR if you prefer though?)

sdroege: Will we go to 1.47 or 1.48 directly?

Note that if you start requiring 1.47 now then the GStreamer CI will fail forever until 1.47.1 or 1.48 is released (because of a clippy bug...).

glib/src/translate.rs
fengalin: Are you kidding me?! :) Last changes I'll do for this MR... After that I'll go on strike like a regular French guy.

glib/src/translate.rs
GuillaumeGomez: iirc, we'll need the 1.48 if we migrate to the clone! proc-macro.

glib/src/translate.rs
GuillaumeGomez: You can do it in a follow-up PR, I don't mind. But then don't forget to open an issue haha.

glib/src/translate.rs
sdroege: > iirc, we'll need the 1.48 if we migrate to the clone! proc-macro.

Fine with me once 1.48 is out and confirmed to work properly :)

fengalin: Ok, so I keep it like this for now and open an issue.

GuillaumeGomez: Thanks!

GuillaumeGomez:
@sdroege It's good for me, so whenever you want.

fengalin:
> @sdroege It's good for me, so whenever you want.

No way! I'm doing the doc changes first :)

GuillaumeGomez:
Oh wowo. Please do so, don't want to be in the middle XD

glib/src/translate.rs
fengalin: https://github.com/gtk-rs/gtk-rs/issues/17

PR #12, opened 2020-11-02
glib-macros: Use absolute paths to the StaticType trait

sdroege:
Otherwise code using it derive macros needs to have it in scope already,
which is not very user-friendly.

----

CC @gdesmott


sdroege:
@GuillaumeGomez This should probably also get a bugfix release.

GuillaumeGomez (approved):
Good catch!

PR #16, opened 2020-11-05
Use `Object<T>::as_mut()` without unsafe, and a couple other small things

ids1024:


gtk/src/subclass/application.rs
```diff
@@ -82,13 +82,12 @@ unsafe impl<T: GtkApplicationImpl> IsSubclassable<T> for Application {
...
-            // Chain our startup handler in here
-            let klass = &mut class.as_mut().parent_class;
-            klass.startup = Some(application_startup::<T>);
-        }
+
+        let klass = class.as_mut();
+        klass.window_added = Some(application_window_added::<T>);
+        klass.window_removed = Some(application_window_removed::<T>);
+        // Chain our startup handler in here
+        let klass = &mut class.as_mut().parent_class;
```
sdroege, line 90: This could also become `let klass: &mut gio_sys::GApplicationClass = class.as_mut()` but that seems more complicated :)

sdroege (approved):

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #23, opened 2020-11-06
Various minor cleanups

sdroege:


ids1024:
This appears to revert the `gir` submodule to an earlier commit. I assume that was committed unintentionally?

sdroege:
> This appears to revert the `gir` submodule to an earlier commit. I assume that was committed unintentionally?

Yes, I hate git submodules :) Thanks for noticing!

bilelmoussaoui (approved):

sdroege:
Once https://github.com/gtk-rs/gir/pull/985 is in and we regenerate with that, we won't have any compiler warnings left with latest stable and nightly. How it should've always been.

Can we have a github action that complains about warnings (makes them errors) and annotates PRs with them somehow? @bilelmoussaoui ?

bilelmoussaoui:
Definitely, I will handle that once my other CI additions are in

atk/src/lib.rs
```diff
@@ -32,6 +32,7 @@ mod rt;
 #[cfg_attr(feature = "cargo-clippy", allow(let_and_return))]
 #[cfg_attr(feature = "cargo-clippy", allow(many_single_char_names))]
 #[cfg_attr(feature = "cargo-clippy", allow(wrong_self_convention))]
+#[allow(unused_imports)]
```
GuillaumeGomez, line 35: At some point it'd be nice if `gir` imports were fixed. :)

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

atk/src/lib.rs
sdroege: Indeed, but that's quite non-trivial to get right :)

PR #27, opened 2020-11-08
Make virtual methods take wrapper of type, not parent

ids1024:
Glib tests seem to be passing, but Gtk and such needs to be updated.

### DIFF
--- a/examples/src/bin/basic_subclass.rs
+++ b/examples/src/bin/basic_subclass.rs
@@ -39,6 +39,7 @@ mod imp_win {
 
     impl ObjectSubclass for SimpleWindow {
         const NAME: &'static str = "SimpleWindow";
+        type Type = super::SimpleWindow;
         type ParentType = gtk::ApplicationWindow;
         type Instance = subclass::simple::InstanceStruct<Self>;
         type Class = subclass::simple::ClassStruct<Self>;
@@ -57,10 +58,8 @@ mod imp_win {
         // Here we are overriding the glib::Objcet::contructed
         // method. Its what gets called when we create our Object
         // and where we can initialize things.
-        fn constructed(&self, obj: &glib::Object) {
+        fn constructed(&self, obj: &Self::Type) {
             self.parent_constructed(obj);
-
-            let self_ = obj.downcast_ref::<super::SimpleWindow>().unwrap();
 
             let headerbar = gtk::HeaderBar::new();
             let increment = gtk::Button::with_label("Increment!");
@@ -72,14 +71,14 @@ mod imp_win {
 
             // Connect our method `on_increment_clicked` to be called
             // when the increment button is clicked.
-            increment.connect_clicked(clone!(@weak self_ => move |_| {
-                let priv_ = SimpleWindow::from_instance(&self_);
+            increment.connect_clicked(clone!(@weak obj => move |_| {
+                let priv_ = SimpleWindow::from_instance(&obj);
                 priv_.on_increment_clicked();
             }));
 
-            self_.add(&label);
-            self_.set_titlebar(Some(&headerbar));
-            self_.set_default_size(640, 480);
+            obj.add(&label);
+            obj.set_titlebar(Some(&headerbar));
+            obj.set_default_size(640, 480);
 
             self.widgets
                 .set(WindowWidgets {
@@ -113,7 +112,7 @@ glib_wrapper! {
 }
 
 impl SimpleWindow {
-    pub fn new(app: &gtk::Application) -> Self {
+    pub fn new(app: &SimpleApplication) -> Self {
         glib::Object::new(Self::static_type(), &[("application", app)])
             .expect("Failed to create SimpleWindow")
             .downcast::<SimpleWindow>()
@@ -131,6 +130,7 @@ mod imp_app {
 
     impl ObjectSubclass for SimpleApplication {
         const NAME: &'static str = "SimpleApplication";
+        type Type = super::SimpleApplication;
         type ParentType = gtk::Application;
         type Instance = subclass::simple::InstanceStruct<Self>;
         type Class = subclass::simple::ClassStruct<Self>;
@@ -155,7 +155,7 @@ mod imp_app {
         // application is launched by the desktop environment and
         // aksed to present itself.
         fn activate(&self, app: &gio::Application) {
-            let app = app.downcast_ref::<gtk::Application>().unwrap();
+            let app = app.downcast_ref::<super::SimpleApplication>().unwrap();
             let priv_ = SimpleApplication::from_instance(app);
             let window = priv_
                 .window
@@ -175,7 +175,7 @@ mod imp_app {
         fn startup(&self, app: &gio::Application) {
             self.parent_startup(app);
 
-            let app = app.downcast_ref::<gtk::Application>().unwrap();
+            let app = app.downcast_ref::<super::SimpleApplication>().unwrap();
             let priv_ = SimpleApplication::from_instance(app);
             let window = SimpleWindow::new(&app);
             priv_
--- a/examples/src/bin/basic_subclass.rs
+++ b/examples/src/bin/basic_subclass.rs
@@ -154,7 +154,7 @@ mod imp_app {
         // `gio::Application::activate` is what gets called when the
         // application is launched by the desktop environment and
         // aksed to present itself.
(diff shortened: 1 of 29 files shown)


glib/src/subclass/types.rs
```diff
@@ -247,6 +247,8 @@ pub trait ObjectSubclass: Sized + 'static {
     /// Optional.
     const ABSTRACT: bool = false;
 
+    type Type: ObjectType + crate::object::ObjectSubclassIs<Subclass=Self>;
```
sdroege, line 250: Could also update `get_instance()` below now, and maybe other APIs. Every usage of `ObjectSubclass::ParentType` apart from the type registration should be `Type` instead.

glib/src/subclass/types.rs
```diff
@@ -247,6 +247,8 @@ pub trait ObjectSubclass: Sized + 'static {
...
+
     /// Parent Rust type to inherit from.
     type ParentType: ObjectType
```
sdroege, line 253: We already know the `ParentType` because of how `glib_wrapper!` works. Can we make use of that for making sure the types align correctly (without adding a new trait)?

glib/src/subclass/object.rs
```diff
@@ -513,6 +522,10 @@ mod test {
         }
     }
 
+    glib_wrapper! {
+        pub struct SimpleObjectWrapper(ObjectSubclass<SimpleObject>);
```
sdroege, line 526: Maybe create an `imp` submodule/etc here too so it's less confusing

glib/src/subclass/object.rs
```diff
@@ -349,6 +353,10 @@ mod test {
         }
     }
 
+    glib_wrapper! {
```
sdroege, line 356: Also here maybe an `imp` submodule too

glib/src/object.rs
```diff
@@ -2535,6 +2536,10 @@ pub unsafe trait ParentClassIs: ObjectType {
     type Parent: ObjectType;
 }
 
+pub unsafe trait ObjectSubclassIs: ObjectType {
```
sdroege, line 2539: Needs docs, which should point to `glib_wrapper!`

sdroege (commented):
Seems generally good to me, thanks for working on this!

glib/src/subclass/types.rs
ids1024: Then that would work with any glib wrapper that has the same parent, even if it's a different type, right? I think we want to enforce that `ObjectSubclass::Type` is exactly a wrapper around the `ObjectSubclass`, not a wrapper around it's parent, a type with the same parent, or a child. An unsafe marker trait like this seems to be a good way to do this.

Also I think it would be possible to have blanket implementations for `IsA<Object>`, `IsA<Widget>` using this, based on whether or not the `ObjectSubclass` implements `ObjectImpl`, `WidgetImpl`, etc. So it wouldn't be necessary to specify the parent types when defining the wrapper. The problem with that is that we couldn't have implementations that conflict or could theoretically conflict. (So we basically couldn't have blanket implementations of `IsA` for anything else.)

sdroege: Not being able to have any other blanket impls seems ok to me. This seems like it would be the most important blanket impl ;) I'm not sure how that could work though, but if you find a way to make it work.

Apart from parent classes there are also implemented interfaces to consider here btw. `gio::ReadStream` for example implements `gio::Seekable`, if you're looking for an example.

examples/src/bin/basic_subclass.rs
```diff
@@ -160,7 +162,7 @@ mod imp_app {
         // application is launched by the desktop environment and
         // aksed to present itself.
         fn activate(&self, app: &gio::Application) {
```
sdroege, line 164: This would also become

```suggestion
        fn activate(&self, app: &SimpleApplication) {
```

once everything is updated?

ids1024: Right. The next step is to update the `*Impl` traits in gtk/etc to be defined similarly to how `ObjectExt` is handled here.

ids1024:
Okay, now gio and gtk are also updated to use `Self::Type`.

gtk/src/subclass/container.rs
```diff
@@ -4,122 +4,136 @@ use glib::translate::*;
...
-pub trait ContainerImplExt {
-    fn parent_add(&self, container: &Container, widget: &Widget);
-    fn parent_remove(&self, container: &Container, widget: &Widget);
-    fn parent_check_resize(&self, container: &Container);
-    fn parent_set_focus_child(&self, container: &Container, widget: Option<&Widget>);
-    fn parent_child_type(&self, container: &Container) -> glib::Type;
-    fn parent_get_path_for_child(&self, container: &Container, widget: &Widget) -> WidgetPath;
-    fn parent_forall(&self, container: &Container, include_internals: bool, callback: &Callback);
+pub trait ContainerImplExt: ObjectSubclass {
+    fn parent_add(&self, container: &Self::Type, widget: &Widget);
```
sdroege, line 45: I was wondering if it could make sense to not pass `&Self::Type` at all here as it's kind of redundant but I think it still makes sense:
  * Usually you need it, so that would otherwise be an additional call to `self.get_instance()`
  * Getting it from `&self` involves reference counting: not a usability problem as that's handled automatically in Rust, but it means additional atomic integer operations that could be avoided otherwise

As such, let's keep it like this for now :)

sdroege:
Looks good to me, but as usual, I'll merge this once I updated everything in GStreamer so that we don't end up with stuff not compiling. That might take a few days as this requires quite a few changes everywhere.

I assume you don't have a script to do the work either :)

gtk/src/subclass/container.rs
sdroege: Also for people using an IDE or LSP plugin for their editor this comes for free. The signature of the trait functions would be autocompleted so typing the `&Self::Type` is not much of an effort.

sdroege:
First part is here https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/629

ids1024:
> so that we don't end up with stuff not compiling.

The hard part is updating the definitions of `*Impl` traits, but that shouldn't be required to keep compiling. It would just make those traits inconsistent with the `glib`/`gtk` ones.

> I assume you don't have a script to do the work either :)

Not fully scripted, but after doing it a few times I did find `sed` helpful for types with a lot of virtual methods:

Manually added `use glib::Cast;`
Manually add bound `ContainerImplExt: ObjectSubclass`
`sed -i 's/container.to_glib_none()/container.unsafe_cast_ref::<Container>().to_glib_none()/' container.rs`
`sed -i 's/\&wrap/wrap.unsafe_cast_ref()/' container.rs`
`sed -i 's/\&Container/\&Self::Type/' container.rs`

ids1024:
It would be nice to have some abstraction handling a lot of the boilerplate for defining these `*Impl` traits. Though the obvious solution is a macro, and another arcane macro generating a bunch of unsafe code doesn't quite appeal...

sdroege:
> Not fully scripted, but after doing it a few times I did find `sed` helpful for types with a lot of virtual methods:

Yeah that's how I did it now. Now I just have to update all the existing GStreamer plugins :sleeping:

gtk/src/subclass/container.rs
ids1024: > As such, let's keep it like this for now :)

Also the soundness issues with `get_instance()` (https://github.com/gtk-rs/gtk-rs/issues/46), which will need some sort of breaking API change to address. I'm personally happy to not be adding a bunch of calls to the function when it can easily be used to produce unsound behavior. And the way it works may have to change to fix the soundness issue anyway.

sdroege: Well, it's difficult to accidentally run into that soundness issue so I wouldn't worry too much about it.

We can also make it easily cause an assertion (or direct segfault because dereferencing some random pointer) if that happens (check that the GType inside the instance is the one we would expect), which should make it even less likely to happen in practice. Want to give that a try?

sdroege:
WIP for migrating all the GStreamer plugins is here: https://gitlab.freedesktop.org/gstreamer/gst-plugins-rs/-/merge_requests/429

I found that it's much nicer to move the `imp` module into a separate file next to a `mod.rs` that has the super module.

Issue #17, opened 2020-11-06
glib/translate: use matches! where applicable (see FIXME comments)


fengalin:
Requires MSRV update first. See https://github.com/gtk-rs/gtk-rs/pull/11#pullrequestreview-525025057.

sdroege:
Done by https://github.com/gtk-rs/gtk-rs/pull/30

PR #30, opened 2020-11-08
CI: add a clippy job

bilelmoussaoui:


.github/workflows/CI.yml
```diff
@@ -117,6 +117,26 @@ jobs:
...
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+          override: true
+      - run: rustup component add clippy
+      - uses: actions-rs/cargo@v1
+        with:
+          command: clippy
+          args: -- -D warnings
```
sdroege, line 138: For GStreamer I also do `-A clippy::redundant_pattern_matching -A clippy::single_match -A clippy::cast_lossless -A clippy::missing_safety_doc`, but that might just be a question of style :)

sdroege:
Good for me, once it passes. That might take some effort though, I don't think the crates above glib/gio are clippy clean. Maybe keep them as warnings for now until it's all cleaned up? That way we at least don't introduce new warnings.

bilelmoussaoui:
> Good for me, once it passes. That might take some effort though, I don't think the crates above glib/gio are clippy clean. Maybe keep them as warnings for now until it's all cleaned up? That way we at least don't introduce new warnings.

I will run clippy locally and see the amount of work needed, I'm willing to put some time on it to get all the warnings cleared.

bilelmoussaoui:
@sdroege this one should be ready

.github/workflows/CI.yml
bilelmoussaoui: I have added those per crate depending on what's used in auto, unless you prefer otherwise?

sdroege: Sure, that also works :)

atk/src/lib.rs
```diff
@@ -7,11 +7,11 @@
...
-#![cfg_attr(feature = "cargo-clippy", allow(let_unit_value))]
-#![cfg_attr(feature = "cargo-clippy", allow(new_without_default))]
-#![cfg_attr(feature = "cargo-clippy", allow(type_complexity))]
-#![cfg_attr(feature = "cargo-clippy", allow(trivially_copy_pass_by_ref))]
-#![cfg_attr(feature = "cargo-clippy", allow(derive_hash_xor_eq))]
+#![allow(clippy::let_unit_value)]
+#![allow(clippy::new_without_default)]
+#![allow(clippy::type_complexity)]
+#![allow(clippy::trivially_copy_pass_by_ref)]
+#![allow(clippy::derive_hash_xor_eq)]
```
sdroege, line 14: We should distinguish here by things in `auto` and elsewhere. In non-auto I would prefer to get them fixed or per case to add an `#[allow]` to it.

.github/workflows/CI.yml
sdroege: `#[allow(clippy::missing_safety_doc)]` I would add directly in `lib.rs` for everything though. These are bindings, everything is unsafe :P

.github/workflows/CI.yml
```diff
@@ -16,7 +16,7 @@ jobs:
           - stable
           - beta
           - nightly
-          - "1.40.0"
+          - "1.42.0"
```
sdroege, line 19: Why 1.42 instead of 1.41?

sdroege:
Seems ok to me otherwise, thanks!

atk/src/lib.rs
bilelmoussaoui: It seems atk doesn't need any of those so I've removed them

.github/workflows/CI.yml
bilelmoussaoui: I thought it was stabilized in 1.41 but it wasn't the case https://github.com/rust-lang/rust/pull/67659/

atk/src/lib.rs
bilelmoussaoui: I will do the same for other crates though.

.github/workflows/CI.yml
bilelmoussaoui: The thing is that a lot of glib is using that per function, not sure whether to remove all of those & set it on lib.rs for all the crates or not?

bilelmoussaoui: Especially that for example, gtk has only one exported unsafe function, maybe it should be done per crate?

atk/src/lib.rs
bilelmoussaoui: Should be done

.github/workflows/CI.yml
bilelmoussaoui: I went ahead and handled that per crate. For glib, gdk, cairo (the three crates that exports a lot of unsafe functions) has it set in lib.rs & for others it's set per function

.github/workflows/CI.yml
```diff
@@ -117,6 +117,26 @@ jobs:
...
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+          override: true
+      - run: rustup component add clippy
+      - uses: actions-rs/cargo@v1
+        with:
+          command: clippy
+          args: --all-features -- -D warnings
```
sdroege, line 138: You want to run with `--all-targets` so it also covers tests

examples/src/bin/clipboard_simple.rs
```diff
@@ -109,12 +109,11 @@ fn build_ui(application: &gtk::Application) {
...
         let clipboard = gtk::Clipboard::get(&gdk::SELECTION_CLIPBOARD);
         clipboard.request_text(|_, t| {
-            if t.is_some() {
-                let t = t.unwrap();
+            if let Some(t) = t {
                 if t.len() >= 4 {
                     GLOBAL.with(|global| {
                         if let Some(ref ui) = *global.borrow() {
-                            ui.button_a1.set_active(t.chars().nth(0).unwrap() == '1');
+                            ui.button_a1.set_active(t.starts_with('1'));
```
sdroege, line 116: This is very strange code :) How about

```rust
let t = t.chars().collect::<Vec<_>>();
if t.len() >= 4 {
    [...]
    ui.button_a1.set_active(t[0] == '1');
    [...]
}
```

sdroege (commented):
Looks good to me otherwise now

examples/src/bin/clipboard_simple.rs
GuillaumeGomez: Indexing on strings is a very bad idea. If you have a multi-bytes character as the first one, you'll panic.

sdroege: Yes, read again what I wrote. That's a `Vec<char>` :P

GuillaumeGomez: Then nevermind. :3

sdroege:
Put a `Fixes https://github.com/gtk-rs/gtk-rs/issues/17` into the commit message somewhere please :)

.github/workflows/CI.yml
sdroege: Works for me

sdroege:
Good, the rust github action is currently broken...

.github/workflows/CI.yml
```diff
@@ -117,6 +117,26 @@ jobs:
           command: fmt
           args: --all -- --check
 
+  clippy:
```
sdroege, line 120: graphene-sys fails because the ABI tests don't generate any layout tests... which seems to be problematic

```
error: unused import: `graphene_sys::*`
 --> graphene/sys/tests/abi.rs:8:5
  |
8 | use graphene_sys::*;
  |     ^^^^^^^^^^^^^^^
  |
  = note: `-D unused-imports` implied by `-D warnings`

error: unused imports: `align_of`, `size_of`
  --> graphene/sys/tests/abi.rs:11:16
   |
11 | use std::mem::{align_of, size_of};
   |                ^^^^^^^^  ^^^^^^^
```

sdroege: @GuillaumeGomez

GuillaumeGomez: Not sure to get the problem here or how I can help...

sdroege: `graphene/sys/tests/abi.rs` contains any of the structs as it should, and you can get that problem if you try to build the tests :)

But I meant to CC @bilelmoussaoui, not you. Sorry

graphene/sys/tests/abi.rs
```diff
@@ -2,6 +2,8 @@
 // from gir-files (https://github.com/gtk-rs/gir-files)
 // DO NOT EDIT
 
+#![allow(unused)]
```
sdroege, line 5: Discussion about this one is https://github.com/gtk-rs/gtk-rs/pull/30#discussion_r525060701

@bilelmoussaoui You want to look into this, or we keep it like above for now and create yet another gir issue for later? :)

bilelmoussaoui: Let's keep it as is and create an issue for gir

bilelmoussaoui:
@sdroege clippy seems to finally be happy :tada:

.github/workflows/CI.yml
```diff
@@ -223,6 +223,26 @@ jobs:
...
+    container:
+      image: ubuntu:20.10
+    steps:
+      - run: apt-get update -y
+      - run: apt-get install -y libgtk-3-dev libglib2.0-dev libgraphene-1.0-dev git xvfb curl libcairo-gobject2 libcairo2-dev
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: nightly
```
sdroege, line 238: As mentioned a few times on IRC yesterday, running nightly clippy is not a good idea. It contains unstabilized lints and is regularly changing, which is suboptimal for a CI. We don't want random contributor's PRs to fail just because last night a new lint was added and some completely unrelated code triggers it.

gdk/src/change_data.rs
```diff
@@ -35,4 +35,8 @@ impl<'a> ChangeData<'a> {
             ChangeData::UChar(_) | ChangeData::UShort(_) | ChangeData::ULong(_) => 1,
         }
     }
+
+    pub fn is_empty(&'a self) -> bool {
```
sdroege, line 39: Without `'a` maybe?

glib/src/subclass/object.rs
```diff
@@ -31,7 +31,7 @@ pub trait ObjectImpl: ObjectSubclass + ObjectImplExt {
     ///
     /// This is called whenever the property value of the specific subclass with the
     /// given index should be returned.
-    fn get_property(&self, _obj: &Self::Type, _id: usize) -> Result<Value, ()> {
+    fn get_property(&self, _obj: &Self::Type, _id: usize) -> Value {
```
sdroege, line 34: Need to update gstreamer-rs / gst-plugins-rs before this can be merged, otherwise things will be broken for a while :)

sdroege: I mean: I'll do that once this PR here is ready but don't merge it before that.

glib/src/variant_type.rs
```diff
@@ -150,7 +151,7 @@ impl VariantTy {
             if ok && end == limit {
                 Ok(&*(type_string.as_bytes() as *const [u8] as *const VariantTy))
             } else {
-                Err(())
+                Err(glib_bool_error!("Invalid type {} string", type_string))
```
sdroege, line 154:
```suggestion
                Err(glib_bool_error!("Invalid type string: '{}'", type_string))
```

.github/workflows/CI.yml
sdroege, line 19: 1.48 is due tomorrow, so maybe let's just go with that tomorrow instead of updating the CI twice :)

.github/workflows/CI.yml
bilelmoussaoui: I'm not sure whether we can get back to using stable for clippy, unless we removed the --all-features and we go with manually speechifying the highest feature ?

sdroege: I think that's the only option, yes

.github/workflows/CI.yml
sdroege: That happened now

@bilelmoussaoui can you update this one?

bilelmoussaoui: Done

.github/workflows/CI.yml
sdroege: This one is the only one left from what I can see, then we can merge it

glib/src/subclass/object.rs
sdroege: 37 lines to change, I'll take care of that when merging :)

.github/workflows/CI.yml
bilelmoussaoui: I think it should be ready, let's see how the CI goes

bilelmoussaoui:
@sdroege should be ready

sdroege:
Thanks, will merge in a few minutes :)

PR #31, opened 2020-11-08
WIP: Refactor Value and Property handling

sdroege:
This will handle https://github.com/gtk-rs/glib/issues/667 and https://github.com/gtk-rs/glib/issues/637

### DIFF
--- a/glib/src/value.rs
+++ b/glib/src/value.rs
@@ -6,10 +6,6 @@
 //! isn't known at compile time but once created a `Value` can't change its
 //! type.
 //!
-//! [`TypedValue`](struct.TypedValue.html) has a statically known type and
-//! dereferences to `Value` so it can be used everywhere `Value` references are
-//! accepted.
-//!
 //! [`SendValue`](struct.SendValue.html) is a version of [`Value`](struct.Value.html)
 //! that can only store types that implement `Send` and as such implements `Send` itself. It
 //! dereferences to `Value` so it can be used everywhere `Value` references are accepted.
@@ -21,10 +17,10 @@
 //!
 //! ```
 //! use glib::prelude::*; // or `use gtk::prelude::*;`
-//! use glib::{TypedValue, Value};
+//! use glib::Value;
 //!
-//! // Value and TypedValue implement From<&i32>, From<&str>
-//! // and From<Option<&str>>. Another option is the `ToValue` trait.
+//! // Value implements From<&i32>, From<&str> // and From<Option<&str>>.
+//! // Another option is the `ToValue` trait.
 //! let mut num = 10.to_value();
 //! let mut hello = Value::from("Hello!");
 //! let none: Option<&str> = None;
@@ -78,11 +74,9 @@
 //! ```
 
 use libc::{c_char, c_void};
-use std::borrow::Borrow;
 use std::error;
 use std::ffi::CStr;
 use std::fmt;
-use std::marker::PhantomData;
 use std::mem;
 use std::ops::Deref;
 use std::ptr;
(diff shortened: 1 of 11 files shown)


sdroege:
For signals we can probably have some associated trait function that passes in some kind of "signal creation context" plus a builder-style API on it for making it easier to create signals. I'll experiment with that too.

sdroege:
Initial WIP version of the Value traits refactoring is here now. Needs all the impls to be updated and to be confirmed that it actually works correctly everywhere.

Some feedback on the traits would be useful already.

sdroege:
> Some feedback on the traits [in the last WIP commit] would be useful already.

@fengalin maybe from you? :)

fengalin:
I'll take a look later today (afk atm). It seems like a neat improvement indeed.

sdroege:
> I'll take a look later today (afk atm). It seems like a neat improvement indeed.

Thanks, no need to hurry :) The goal is basically to have a direct mapping between types, param specs and the Value API. Based on that we can derive subclass property handling code (the pspecs and getters/setters just based on the type), signal handler code (automatic Value <-> Rust type conversion for the parameters) and other nice things that are currently not possible.

sdroege:
`src/value.rs` should be finished now except for the TODO comments (implementing ownership transferring `TryFrom<Value> for T` and `From<T> for Value`).

Overall there is more code to implement per type now for working with `Value`s, but at least the majority of the cases for this are going to be covered by the `glib_wrapper!` macro (objects, boxed, shared) and gir (`src/auto/flags.rs` and `src/auto/enums.rs`). Nonetheless, if someone has an idea how to simplify it, please let me know.

glib/src/value.rs
```diff
@@ -80,44 +76,216 @@
...
+
+impl<'r, T: ValueTypeNullable> ValueType for Option<T>
+where
+    Self: for<'s> FromValue<'s, Error = WrongValueTypeError>,
+{
+    type Type = T::Type;
+}
+
+/// A type for which a `ParamSpec` exists and that can be used for object properties.
+pub trait HasParamSpec: ValueType {
```
fengalin, line 117: Nice! (also the following ones oc).

Minor nit though :) Since this is more about a type that allows defining `ParamSpec`s with specificities, `Has` might suggest that a particular `ParamSpec` is associated with the type. Maybe we could use `CanParamSpec` instead (and `CanParamSpecWithDefault`, ...)

glib/src/value.rs
```diff
@@ -80,44 +76,216 @@
...
+pub struct WrongValueTypeError {
     pub actual: Type,
     pub requested: Type,
 }
 
-impl GetError {
-    pub fn new_type_mismatch(actual: Type, requested: Type) -> Self {
-        GetError { actual, requested }
+impl WrongValueTypeError {
+    pub fn check<T: StaticType>(value: &Value) -> Result<(), WrongValueTypeError> {
```
fengalin, line 165: Something feels wrong with this one. Shouldn't `WrongValueTypeError` result from a `check` somewhere else? Maybe `Value::check_type`?

glib/src/value.rs
```diff
@@ -246,9 +351,9 @@ impl Value {
     }
 
     /// Tries to transform the value into a value of the target type
-    pub fn transform<T: StaticType + SetValue>(&self) -> Option<Value> {
+    pub fn transform<T: ValueType>(&self) -> Option<Value> {
```
fengalin, line 354: Maybe that would be worth changing the API and returning a `Result` so that we can easily `Display` the type mismatch.

fengalin (commented):
I only looked rapidly so far. Will take a deeper dive tomorrow hopefully.

glib/src/value.rs
sdroege: Good idea

glib/src/value.rs
sdroege: That would make it public `Value` API, which I thought might be confusing for users. But can be done :)

glib/src/value.rs
sdroege: `Create` or `Construct` maybe? It's not that this value "can" param spec, but you can "create"/"construct" one. Seems like a better verb or what do you think?

glib/src/value.rs
```diff
@@ -80,44 +76,216 @@
...
+    /// This exists only for handling nullable types.
+    // FIXME: Should default to Self once associated type defaults are stabilized
+    // https://github.com/rust-lang/rust/issues/29661
+    type Type: StaticType;
+}
+
+/// A type that can be stored in `Value`s and is nullable.
+///
+/// These are types were storing an `Option` is valid. Examples are `String` and all object types.
+pub trait ValueTypeNullable:
```
sdroege, line 104:
```suggestion
pub trait ValueTypeOptional:
```

Maybe that's better / more consistent? Also for the other usages of "nullable" in here

glib/src/value.rs
fengalin: Maybe it doesn't need to be made `pub`, if it's only checked from the module / crate.

glib/src/value.rs
fengalin: Oc! I only tried something short so far. I'll come back here if I find something else.

glib/src/value.rs
sdroege: It has to be called by all `FromValue` impls to check the type (in addition to whatever else they might need to check)

glib/src/value.rs
fengalin: Yes! Initially, I wrote this exact suggestion and I stepped back when I understood it was intended for expressing glib C null-ability. But since it will be visible within Rust code only, I think `Optinal` is better.

glib/src/value.rs
fengalin: Indeed. Another solution, if you want to avoid the confusion, would be to add an empty `ValueType`, so we would use `ValueType::check::<T>(value)` but we get turbofished (didn't confirmed), or simply have it on `StaticType` somehow: `StaticType::check(value)`

glib/src/value.rs
fengalin: How about `UseableAsParam`,  `UseableAsParamWithDefault` & `UseableAsParamWithMinMax`? I think this captures the meaning better than the `ParamSpec` orientation. We could also use `CanUseAsParam*`, shorted :)

sdroege:
Ok, no further comments here so I proceed :)

glib/src/value.rs
sdroege: `UseableAsParam` it is

glib/src/value.rs
sdroege: Well, this is basically `Value::is()` (which is implemented in terms of that currently) but for that it wouldn't make sense to return a `Result` :) I'll keep it as it is for now, all other options seem equally strange.

sdroege:
Ok, next step is to update gir and then the other crates and then get all tests pass :sleeping:

sdroege:
I wonder if I should split this PR so that part of it can already get in and doesn't bitrot while I need more time to finish the other parts.

The changes to the property handling can go in already and are useful and nice separately too. @bilelmoussaoui @GuillaumeGomez opinions?

examples/src/bin/listbox_model.rs
```diff
@@ -365,17 +333,42 @@ mod row_data {
         // This maps between the GObject properties and our internal storage of the
         // corresponding values of the properties.
         impl ObjectImpl for RowData {
-            fn set_property(&self, _obj: &Self::Type, id: usize, value: &glib::Value) {
-                let prop = &PROPERTIES[id];
+            fn properties() -> Vec<glib::ParamSpec> {
```
bilelmoussaoui, line 336: Does it has to return a Vec here?

sdroege: What else?

bilelmoussaoui: What I meant is whether it can return a `&'static [glib::ParamSpec]` ? :)

sdroege: No because creating a paramspec is not a const function. You wouldn't be able to initialize that array with anything.

sdroege: Well, you could go with storing it in a `Lazy` inside the function and returning it from that... but that seems overly complicated but arguably more correct.

Would you prefer that?

sdroege: Well, let's change it to that. Then at least it gives reliable results if called multiple times

bilelmoussaoui: >  Well, you could go with storing it in a Lazy inside the function and returning it from that... but that seems overly complicated but arguably more correct.

Well, if the plan is to provide some nice macro around all of this, I would rather go for correctness as it's already a bit complex to do all the properties stuff

sdroege: This is implemented in https://github.com/gtk-rs/gtk-rs/pull/270 now, which supersedes this PR.

I'll update this PR to only include the `Value` changes once the other one is merged.

sdroege:
I'm working currently on finishing this btw. The next PR is going to have the refactored traits but none of the new pspec-related traits yet (I think we'll want to keep them for the release afterwards).

sdroege:
I have glib compiling again, and gio almost.

sdroege:
Current WIP is at https://github.com/gtk-rs/gtk-rs/pull/454

sdroege:
Let's close this one for now.

Issue #28, opened 2020-11-08
Crash because log_domain is NULL.


chengchangwu:
It seems `gst_video_info_from_caps` may generate a null `log_domain` because `G_LOG_DOMAIN` is not defined while building gst-plugins-base. I'm using gstreamer 1.16.

[crash-because-of-log-domain-null.log](https://github.com/gtk-rs/gtk-rs/files/5505918/crash-because-of-log-domain-null.log)

chengchangwu:
I can make a PR, changing `log_domain` to a empty str or and "<unknown>" when `log_domain` is null. But I'm not sure it's the right way to do."

sdroege:
See https://gitlab.freedesktop.org/gstreamer/gstreamer/-/issues/634 but a `NULL` log domain is perfectly valid, see `glib/gmessages.h`:

```cpp
#ifndef G_LOG_DOMAIN
#define G_LOG_DOMAIN    ((gchar*) 0)
#endif  /* G_LOG_DOMAIN */
```

Can you create a PR that changes the API to work with an `Option` for the log domain (in all functions and the callback)?

For the stable release we should indeed map it to `"<null>"`. I'll fix that once the fix for git master is in.

chengchangwu:
I'll try to make a PR to change the API.

sdroege:
Great, thanks :) Let me know if you need help

chengchangwu:
I'm trying to build a local version of gtk-rs and gstreamer-rs. For gstreamer-rs I have followed the instructions of gstreamer-rs/DEVELOPMENT. For gtk-rs I only do `python3 generator.py`。I've modified my application's Cargo.toml to use my local versions of gtk-rs and gstreamer. But the compilation failed with many errors. Any example of the error is,

```
error: no rules expected the token `,`
  --> /home/ccwu/git/rust/gstreamer-rs/gstreamer/src/auto/allocator.rs:11:82
   |
11 |     pub struct Allocator(Object<gst_sys::GstAllocator, gst_sys::GstAllocatorClass, AllocatorClass>) @extends Object;
   |                                                                                  ^ no rules expected this token in macro call
```

What I am missing? I've never build my local version of gtk-rs and gstreamer-rs before.

sdroege:
[...]

PR #35, opened 2020-11-09
Allow log_domain to be null.

chengchangwu:
Hope this has fixed issue #28.

### DIFF
--- a/glib/src/log.rs
+++ b/glib/src/log.rs
@@ -130,18 +130,14 @@ pub fn log_set_handler<P: Fn(Option<&str>, LogLevel, &str) + Send + Sync + 'stat
         message: *const libc::c_char,
         user_data: glib_sys::gpointer,
     ) {
+        let log_domain: Borrowed<Option<GString>> = from_glib_borrow(log_domain);
         let message: Borrowed<GString> = from_glib_borrow(message);
         let callback: &P = &*(user_data as *mut _);
-        if log_domain.is_null() {
-            (*callback)(None, from_glib(log_level), message.as_str());
-        } else {
-            let log_domain: Borrowed<GString> = from_glib_borrow(log_domain);
-            (*callback)(
-                Some(log_domain.as_str()),
-                from_glib(log_level),
-                message.as_str(),
-            );
-        }
+        (*callback)(
+            (*log_domain).as_deref(),
+            from_glib(log_level),
+            message.as_str(),
+        );
     }
     let log_func = Some(log_func_func::<P> as _);
     unsafe extern "C" fn destroy_func<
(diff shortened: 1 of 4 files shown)


glib/src/log.rs
```diff
@@ -114,27 +114,39 @@ fn to_log_flags(fatal: bool, recursion: bool) -> u32 {
...
         log_domain: *const libc::c_char,
         log_level: glib_sys::GLogLevelFlags,
         message: *const libc::c_char,
         user_data: glib_sys::gpointer,
     ) {
-        let log_domain: Borrowed<GString> = from_glib_borrow(log_domain);
         let message: Borrowed<GString> = from_glib_borrow(message);
         let callback: &P = &*(user_data as *mut _);
-        (*callback)(log_domain.as_str(), from_glib(log_level), message.as_str());
+        if log_domain.is_null() {
```
sdroege, line 135: You can do `let log_domain: Borrowed<Option<GString>> = from_glib_borrow(log_domain)` and then use it as `log_domain.as_deref()` or similar

sdroege: No need for the separate `NULL` check

sdroege (commented):

glib/src/log.rs
chengchangwu: I've fixed that in the next commit.

chengchangwu:
Why can the github pipeline report more errors than I can find with `cargo test`.

chengchangwu:
I do not know how to change type of `log_domain` in `g_log` macro and fix the errors reported by `cargo test`.

glib/src/log.rs
```diff
@@ -304,30 +318,30 @@ pub fn log_default_handler(log_domain: &str, log_level: LogLevel, message: Optio
 /// ```no_run
 /// use glib::{LogLevel, g_log};
 ///
-/// g_log!("test", LogLevel::Debug, "test");
-/// g_log!("test", LogLevel::Message, "test");
+/// g_log!(Some("test"), LogLevel::Debug, "test");
```
sdroege, line 321: For the macro there should simply be another variant that works without domain so you can do

```rust
g_log!(LogLevel::Debug, "test");
g_log!("domain", LogLevel::Debug, "test");
```

sdroege: @GuillaumeGomez Can you help with that maybe?

GuillaumeGomez: Taking a look!

GuillaumeGomez:
Updated!

glib/src/log.rs
```diff
@@ -331,17 +331,68 @@ pub fn log_default_handler(log_domain: Option<&str>, log_level: LogLevel, messag
...
 /// g_log!(Some("test"), LogLevel::Warning, "test: {} {}", x, "a",);
 /// ```
+///
+/// To be noted that the log domain is optional:
+///
+/// ```no_run
+/// use glib::{LogLevel, g_log};
+///
+/// // As you can see: no log domain:
+/// g_log!("test", LogLevel::Message, "test");
```
sdroege, line 341: You actually provide a log domain here :)

glib/src/log.rs
```diff
@@ -581,10 +632,10 @@ macro_rules! g_info {
 /// ```
 #[macro_export]
 macro_rules! g_debug {
-    ($log_domain:expr, $format:expr, $($arg:expr),* $(,)?) => {{
+    ($log_domain:expr, $format:literal, $($arg:expr),* $(,)?) => {{
```
sdroege, line 635: The examples in the docs above everywhere still use `Some("test")` as log domain. That should allow omitting the log domain completely, or passing it as `"test"` without the `Option` wrapping :)

glib/src/log.rs
GuillaumeGomez: Woups! Good catch. :)

glib/src/log.rs
GuillaumeGomez: Oh right, it's clearly not mandatory.

GuillaumeGomez: Well, not anymore at least.

GuillaumeGomez:
I removed `Some` from the examples.

PR #135, opened 2020-11-10
inet_socket_address: Add std::net::SocketAddr conversions

abdulrehman-git:


gio/src/inet_socket_address.rs
```diff
@@ -0,0 +1,24 @@
...
+    }
+}
+
+impl From<SocketAddr> for InetSocketAddress {
+    fn from(addr: SocketAddr) -> Self {
+        InetSocketAddress::from_socket_address(addr)
+    }
+}
+
+impl Into<SocketAddr> for InetSocketAddress {
```
sdroege, line 19:
```suggestion
impl From<InetSocketAddress> for SocketAddr {
```

Would be better and also possible, or not?

gio/src/inet_socket_address.rs
```diff
@@ -0,0 +1,24 @@
+use InetAddress;
+use InetSocketAddress;
+use InetSocketAddressExt;
+
+use std::net::SocketAddr;
+
+impl InetSocketAddress {
+    pub fn from_socket_address(addr: SocketAddr) -> Self {
```
sdroege, line 8: I think this is not needed. The `From` impl below does exactly that already :)

gio/src/inet_socket_address.rs
```diff
@@ -0,0 +1,24 @@
...
+impl From<SocketAddr> for InetSocketAddress {
+    fn from(addr: SocketAddr) -> Self {
+        InetSocketAddress::from_socket_address(addr)
+    }
+}
+
+impl Into<SocketAddr> for InetSocketAddress {
+    fn into(self) -> SocketAddr {
+        //FIXME: unwrap is unsound here, may use TryInto?
+        SocketAddr::new(self.get_address().unwrap().into(), self.get_port())
```
sdroege, line 22: No, that's OK for now. This can't possibly return `None` and I'm going to fix that once https://gitlab.gnome.org/GNOME/glib/-/merge_requests/1708 is merged into gio.

sdroege (commented):
Looks generally good to me apart from those minor things, thanks!

gio/src/inet_socket_address.rs
abdulrehman-git: I think that's better :). Let me update.

gio/src/inet_socket_address.rs
abdulrehman-git: Ah Okay :).

gio/src/inet_socket_address.rs
sdroege: I've updated it for everything else here: https://github.com/gtk-rs/gtk-rs/pull/136

sdroege (approved):

sdroege:
Thanks!

PR #136, opened 2020-11-10
Replace all Into impls with the corresponding From impls

sdroege:
This is more flexible and still provides the old API in addition.


sdroege:
Much easier to do such gardening tasks now that we have it all merged into a single repo.

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #139, opened 2020-11-11
migration cleanup

bilelmoussaoui:
One thing I've noticed is that each crate had a different readme, we could probably try to expose the list of features there, a quick link to the crates.io/docs for the specific crate and probably the C/upstream projects links.


sdroege:
Yes making them consistent would be useful. What you suggest makes sense to me

bilelmoussaoui:
Right, I will try to do what I have in mind for a specific crate and see how that goes.

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

sdroege:
This was not finished yet :P

bilelmoussaoui:
Can be done separately :P

bilelmoussaoui:
I should have probably tagged it as WIP to avoid such situations in the future ^^

GuillaumeGomez:
Arf sorry. >< Based on the code I just assumed it was, my bad.

PR #140, opened 2020-11-12
Doc cfg attr

GuillaumeGomez:
Follow-up of https://github.com/gtk-rs/gir/pull/988


GuillaumeGomez:
@bilelmoussaoui Running with `--all-features` on stable is going to be problematic with this. Should I disable it for the stable channel in here?

bilelmoussaoui:
I guess we will have to manually switch to the highest feature there's?

pangocairo/src/lib.rs
```diff
@@ -3,6 +3,7 @@
 // Licensed under the MIT license, see the LICENSE file or <https://opensource.org/licenses/MIT>
 
 #![allow(deprecated)]
+#![cfg_attr(feature = "dox", feature(doc_cfg))]
```
sdroege, line 6: This only works with nightly or not?

gdk-pixbuf/src/pixbuf.rs
```diff
@@ -344,6 +344,7 @@ impl Pixbuf {
     }
 
     #[cfg(any(feature = "v2_32", feature = "dox"))]
+    #[cfg_attr(feature = "dox", doc(cfg(feature = "v2_32")))]
```
sdroege, line 347: Do you have a script to update all these?

pangocairo/src/lib.rs
GuillaumeGomez: Only for nightly, which is why I added this line.

gdk-pixbuf/src/pixbuf.rs
GuillaumeGomez: No, by hand.

pangocairo/src/lib.rs
sdroege: If someone wants to build docs locally that's a bit annoying, but I guess acceptable :)

GuillaumeGomez:
@bilelmoussaoui I fixed the CI. Do you think it's fine as is?

bilelmoussaoui:
@GuillaumeGomez do we have to disable running tests/builds on stable/beta/v1.40 ? otherwise the `if: ....` parts should be removed as you have switched to the latest feature there's for each crate already

GuillaumeGomez:
The feature I added is nightly only, which means we can't run with `dox` nor `--all-features`. Which is why I splitted both runs where `--all-features` was used.

sdroege:
@GuillaumeGomez What's missing here?

GuillaumeGomez:
The confirmation from @bilelmoussaoui I guess?

bilelmoussaoui:
Let's wait till the CI can work again

PR #153, opened 2020-11-16
Reexport sys dependencies

GuillaumeGomez:


atk/src/lib.rs
```diff
@@ -18,7 +18,9 @@ extern crate libc;
 #[macro_use]
 extern crate bitflags;
 
-extern crate atk_sys;
+#[doc(hidden)]
+pub extern crate atk_sys;
+pub use atk_sys as ffi;
```
sdroege, line 23: Why do we re-export each under both names?

GuillaumeGomez: I did the same we did in cairo, but I'm fine with reexporting as is.

sdroege: Having it twice publically re-exported seems wrong (unless that's already used by a macro externally, which might be the case for `glib_sys` at least). I would go with

```rust
extern crate atk_sys;
pub use atk_sys as ffi;
```

The first because that's what the autogenerated code expects, the second for external users.

sdroege: And then also change cairo :)

GuillaumeGomez:
```
error: extern crate `gdk_sys` is private, and cannot be re-exported (error E0365), consider declaring with `pub`
  --> gdk/src/lib.rs:11:9
   |
11 | pub use gdk_sys as ffi;
   |         ^^^^^^^^^^^^^^
   |
   = note: `#[deny(pub_use_of_private_extern_crate)]` on by default
   = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
   = note: for more information, see issue #34537 <https://github.com/rust-lang/rust/issues/34537>

```

So what do you prefer? :p

sdroege: Let's keep like you did for now then until we switch to Rust 2018 :)

sdroege: I.e. feel free to merge this one here

GuillaumeGomez: Ok!

PR #154, opened 2020-11-17
Also regen gobject if needed

GuillaumeGomez:
It's all because gobject has a GIR file named `Gir_gobject.toml`, which didn't match what `generator.py` was expecting. :)


ids1024:
Ah yes, that would explain why I had to manually change `gobject/auto` in 659b2f5e45eff3efac7ec69459577004d3a560b3.

GuillaumeGomez:
Now you know. :3

PR #155, opened 2020-11-18
Replace manual enum functions with autogenerated ones

MarijnS95:
This is the gtk-rs counterpart of https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/630, following the introduction of autogenerated associated function support in https://github.com/gtk-rs/gir/pull/991.

Turns out that, as far as I was able to run the tests (after extracting them from `CI.yml`...), the only functions that were defined by hand are in pango's `Gravity` enum :tada: 
(testing with `--all-features` doesn't work because that inadvertently removes functions through `cfg(not())` which are used by examples)

Note that this includes https://github.com/gtk-rs/gir/pull/994 to get rid of `doc(cfg())` on match arms and expressions. Solves a ton of warnings (When compiling with `dox`) but at the same convolutes this PR. See the second commit for an actually relevant change.

Let's hold off merging until `gir` is ready.


sdroege (approved):
Looks generally good to me, thanks!

MarijnS95:
@sdroege Pushed!

Turns out I have more `write!(f, ...)` -. `f.write_str(...)` changes lying around but that results in a whopping 400 files changed. I don't think that's worth taking in assuming format strings like `"{}"` or literals without formatting arguments are optimized out?

sdroege:
> I don't think that's worth taking in assuming format strings like `"{}"` or literals without formatting arguments are optimized out?

I don't think that assumption is necessarily correct :) But better put those into a separate PR, not as part of this one. That's an easy to review change.

MarijnS95:
@sdroege Hence the question mark, I'm curious yet have never looked into the resulting assembly from all these formatting macros. I assumed it to be pretty efficient but that might have been massively wrong :)

sdroege:
> @sdroege Hence the question mark, I'm curious yet have never looked into the resulting assembly from all these formatting macros. I assumed it to be pretty efficient but that might have been massively wrong :)

Well, IMHO it's also nicer to read code-wise so definitely a useful change even if it compiles to the same code. It actually might as the format string is handled at compile-time.

MarijnS95:
Finally got around to writing the three variants in godbolt: https://godbolt.org/z/v3xoxK

Note that this is might be hard to relate as `to_str` is inlined and we don't have an external C function call the compiler knows nothing about. But then again we're all about the formatting, let's see:

`write!(f, "{}", self.to_str())` is by far the worst. Besides `write!()` expanding to `.write_fmt` which needs to set up an `std::fmt::Arguments` object, even on `-C opt-level=3` `::fmt` is called on the `&'static str` as well.

Next is writing a literal with `write!(f, "My value")`. This might be used in a `match self { EnumVariant => write!(f, "Enum variant"), }`. As above, this has to set up `std::fmt::Arguments` for `.write_fmt`.

Finally there is `.write_str()`. This is what `.write_fmt` uses internally so it's obvious we just got rid of all the extra work around formatting, which must inevitably be faster.
In addition it's easier to use in `self.write_str(self.to_str())` and `self.write_str(match self { EnumVariant => "Enum variant", })` becomes more elegant.

sdroege:
Ok, so as expected I guess? :)

MarijnS95:
@sdroege Yup. It's not all too suboptimal. The function pointer `Display::fmt` for `&str` is pushed into the `Arguments` object which is called from `write_fmt`. The implementation simply calls `.pad()` which expands to a plain `write_str` when there are no padding options in the current formatter. We'd be getting rid of the virtual functions and extra boilerplate, not much else besides cosmetics. Don't think that's a change worth making.

gdk/src/auto/device_pad.rs
```diff
@@ -51,22 +39,16 @@ impl<O: IsA<DevicePad>> DevicePadExt for O {
...
-    #[cfg(any(feature = "v3_22", feature = "dox"))]
-    #[cfg_attr(feature = "dox", doc(cfg(feature = "v3_22")))]
     fn get_n_features(&self, feature: DevicePadFeature) -> i32 {
         unsafe {
             ffi::gdk_device_pad_get_n_features(self.as_ref().to_glib_none().0, feature.to_glib())
         }
     }
 
-    #[cfg(any(feature = "v3_22", feature = "dox"))]
-    #[cfg_attr(feature = "dox", doc(cfg(feature = "v3_22")))]
```
GuillaumeGomez, line 69: This shouldn't have been removed based on the documentation: https://developer.gnome.org/gdk3/stable/GdkDevicePad.html#gdk-device-pad-get-n-groups

gdk/src/auto/device_pad.rs
```diff
@@ -51,22 +39,16 @@ impl<O: IsA<DevicePad>> DevicePadExt for O {
...
     fn get_group_n_modes(&self, group_idx: i32) -> i32 {
         unsafe { ffi::gdk_device_pad_get_group_n_modes(self.as_ref().to_glib_none().0, group_idx) }
     }
 
-    #[cfg(any(feature = "v3_22", feature = "dox"))]
-    #[cfg_attr(feature = "dox", doc(cfg(feature = "v3_22")))]
```
GuillaumeGomez, line 61: Same.

gdk/src/auto/device_pad.rs
```diff
@@ -51,22 +39,16 @@ impl<O: IsA<DevicePad>> DevicePadExt for O {
         }
     }
 
-    #[cfg(any(feature = "v3_22", feature = "dox"))]
-    #[cfg_attr(feature = "dox", doc(cfg(feature = "v3_22")))]
```
GuillaumeGomez, line 55: Same.

atk/src/auto/enums.rs
```diff
@@ -211,6 +211,23 @@ pub enum RelationType {
...
+        assert_initialized_main_thread!();
+        unsafe { from_glib(ffi::atk_relation_type_for_name(name.to_glib_none().0)) }
+    }
+
+    pub fn get_name(self) -> Option<glib::GString> {
+        assert_initialized_main_thread!();
+        unsafe { from_glib_none(ffi::atk_relation_type_get_name(self.to_glib())) }
+    }
+
+    pub fn register(name: &str) -> RelationType {
```
sdroege, line 225: That seems problematic and will probably need manual bindings to be properly usable.

atk/src/auto/enums.rs
```diff
@@ -469,6 +486,28 @@ pub enum Role {
...
+        assert_initialized_main_thread!();
+        unsafe { from_glib_none(ffi::atk_role_get_localized_name(self.to_glib())) }
+    }
+
+    pub fn get_name(self) -> Option<glib::GString> {
+        assert_initialized_main_thread!();
+        unsafe { from_glib_none(ffi::atk_role_get_name(self.to_glib())) }
+    }
+
+    pub fn register(name: &str) -> Role {
```
sdroege, line 505: And this

atk/src/auto/enums.rs
```diff
@@ -1069,6 +1108,23 @@ pub enum StateType {
...
+        assert_initialized_main_thread!();
+        unsafe { from_glib(ffi::atk_state_type_for_name(name.to_glib_none().0)) }
+    }
+
+    pub fn get_name(self) -> Option<glib::GString> {
+        assert_initialized_main_thread!();
+        unsafe { from_glib_none(ffi::atk_state_type_get_name(self.to_glib())) }
+    }
+
+    pub fn register(name: &str) -> StateType {
```
sdroege, line 1122: And this

atk/src/auto/enums.rs
```diff
@@ -1295,6 +1351,28 @@ pub enum TextAttribute {
...
+        assert_initialized_main_thread!();
+        unsafe { from_glib_none(ffi::atk_text_attribute_get_name(self.to_glib())) }
+    }
+
+    pub fn get_value(self, index_: i32) -> Option<glib::GString> {
+        assert_initialized_main_thread!();
+        unsafe { from_glib_none(ffi::atk_text_attribute_get_value(self.to_glib(), index_)) }
+    }
+
+    pub fn register(name: &str) -> TextAttribute {
```
sdroege, line 1370: And this

gtk/src/auto/widget.rs
```diff
@@ -91,7 +91,7 @@ pub trait WidgetExt: 'static {
 
     fn create_pango_layout(&self, text: Option<&str>) -> pango::Layout;
 
-    //fn destroyed<P: IsA<Widget>>(&self, widget_pointer: &P);
+    //fn destroyed<P: IsA<Widget>>(&self, widget_pointer: P);
```
sdroege, line 94: Why did this change? :)

pango/src/auto/enums.rs
```diff
@@ -123,6 +125,16 @@ pub enum AttrType {
     __Unknown(i32),
 }
 
+impl AttrType {
+    pub fn get_name(self) -> Option<glib::GString> {
+        unsafe { from_glib_none(ffi::pango_attr_type_get_name(self.to_glib())) }
+    }
+
+    pub fn register(name: &str) -> AttrType {
```
sdroege, line 133: Problematic too

sdroege:
Otherwise looks good

gtk/src/auto/widget.rs
MarijnS95: I think that's https://github.com/gtk-rs/gir/pull/991/commits/08c5ff7f6c390e15638cd7023917c55222fccc73.

sdroege: Hm maybe something broken with commented out code. It doesn't have any effect on any of the actually generated code :man_shrugging: Let's ignore that then or what do you think? Also @GuillaumeGomez?

sdroege (approved):

gtk/src/auto/widget.rs
MarijnS95: > It doesn't have any effect on any of the actually generated code

It changes all enum functions to take `self` by move/copy instead of `&self`.

sdroege: Sure, I mean it has no effect on previously generated code (i.e. it doesn't break anything) :)

MarijnS95: I wonder why this function is not generated; is it the double pointer?

When someone looks into it I guess this is going to be a `&mut *mut P`, right?

sdroege: Yes that's the reason here

sdroege:
@MarijnS95 Your latest gir changes don't affect this one here? Otherwise please also update it :)

MarijnS95:
@sdroege Nope, all clear here :)

sdroege:
@MarijnS95 Please update this one with the latest gir, and then let's get it in :)

MarijnS95:
@sdroege Thanks again! Updated with the latest hashes :)

sdroege (approved):

PR #160, opened 2020-11-20
CI: don't use --workspace for docs

bilelmoussaoui:
Otherwise we get examples there as well, see https://gtk-rs.org/gtk-rs/gtk


sdroege:
@GuillaumeGomez Please fix the clone macro and push into this PR too :)

PR #161, opened 2020-11-20
Switch 2018

GuillaumeGomez:
I'll update/rebase this PR once https://github.com/gtk-rs/gir/pull/996 has been merged.

PR #165, opened 2020-11-21
Switch 2018

GuillaumeGomez:
Take over of #161.


sdroege:
@GuillaumeGomez I'll leave fixing the remainder of the CI to you. At least the normal build is fine now so I can continue, only tests and clippy are not.

PR #170, opened 2020-11-22
generator: Convert to pathlib, fix wrong path passed to various executables

MarijnS95:
`def_check_submodule` was always passing `'gir'` even when called with `gir-files`, and the `depth` used to derive whether `Gir.toml` resides in a `sys` crate is inaccurate when specifying eg. `**/sys` on the command line (as that'd be level 0).

In addition, convert path handling to `pathlib` whose paths are constructed and checked during argument parsing.

Please double-check if I overlooked a path anywhere that's still a string instead of `Path`.

EDIT: With pathlib we might also replace the depth-first search in `regen_crates` with `path.glob` or `path.rglob` on `**/Gir*.toml`?

### DIFF
--- a/generator.py
+++ b/generator.py
@@ -147,9 +147,6 @@ def main():
     if conf.gir_files_path == DEFAULT_GIR_FILES_DIRECTORY:
         if def_check_submodule(conf.gir_files_path, conf) == FAILURE:
             return 1
-    elif not isdir(conf.gir_files_path):
-        print("`{}` dir doesn't exist. Aborting...".format(conf.gir_files_path))
-        return 1
 
     if conf.gir_path == DEFAULT_GIR_PATH:
         if not build_gir_if_needed(def_check_submodule(DEFAULT_GIR_DIRECTORY, conf)):
--- a/generator.py
+++ b/generator.py
@@ -72,12 +72,12 @@ def build_gir_if_needed(updated_submodule):
     return True
 
 
-def regen_crates(path, conf, level=0):
+def regen_crates(path, conf):
     if path.is_dir():
-        for entry in path.iterdir():
-            if level < 3 and not regen_crates(entry, conf, level + 1):
+        for entry in path.rglob("Gir*.toml"):
+            if not regen_crates(entry, conf):
                 return False
-    elif path.name.startswith("Gir") and path.suffix == ".toml":
+    elif path.match("Gir*.toml"):
         print('==> Regenerating "{}"...'.format(path))
 
         args = [conf.gir_path, '-c', path, '-o', path.parent, '-d', conf.gir_files_path]
@@ -93,6 +93,9 @@ def regen_crates(path, conf, level=0):
             if not ask_yes_no_question('Do you want to continue?', conf):
                 return False
         print('<== Done!')
+    else:
+        print('==> {} is not a valid Gir*.toml file'.format(path))
+        return False
     return True
 
 
--- a/generator.py
+++ b/generator.py
@@ -1,7 +1,6 @@
 #!/usr/bin/env python3
 
-from os import listdir
-from os.path import isfile, isdir, join
+from pathlib import Path
 import argparse
 import subprocess
 import sys
@@ -11,9 +10,9 @@ NOTHING_TO_BE_DONE = 0
 NEED_UPDATE = 1
 FAILURE = 2
 
-DEFAULT_GIR_FILES_DIRECTORY = './gir-files'
-DEFAULT_GIR_DIRECTORY = './gir/'
-DEFAULT_GIR_PATH = DEFAULT_GIR_DIRECTORY + 'target/release/gir'
+DEFAULT_GIR_FILES_DIRECTORY = Path('./gir-files')
+DEFAULT_GIR_DIRECTORY = Path('./gir/')
+DEFAULT_GIR_PATH = DEFAULT_GIR_DIRECTORY / 'target/release/gir'
 
 
 def run_command(command, folder=None):
@@ -37,7 +36,7 @@ def update_workspace():
 def ask_yes_no_question(question, conf):
     question = '{} [y/N] '.format(question)
     if conf.yes:
-        print(question)
+        print(question + 'y')
         return True
     if sys.version_info[0] < 3:
         line = raw_input(question)
@@ -47,7 +46,7 @@ def ask_yes_no_question(question, conf):
 
 
 def def_check_submodule(submodule_path, conf):
-    if len(listdir(submodule_path)) != 0:
+    if any(submodule_path.iterdir()):
         return NOTHING_TO_BE_DONE
     print('=> Initializing {} submodule...'.format(submodule_path))
     if not run_command(['git', 'submodule', 'update', '--init', submodule_path]):
@@ -78,18 +77,16 @@ def build_gir_if_needed(updated_submodule):
 
 
 def regen_crates(path, conf, level=0):
-    for entry in listdir(path):
-        entry_file = join(path, entry)
-        if isdir(entry_file):
-            if level < 2 and not regen_crates(entry_file, conf, level + 1):
+    for entry in path.iterdir():
+        if entry.is_dir():
+            if level < 2 and not regen_crates(entry, conf, level + 1):
                 return False
-        elif entry.startswith("Gir") and entry.endswith(".toml"):
-            print('==> Regenerating "{}"...'.format(entry_file))
+        elif entry.name.startswith("Gir") and entry.suffix == ".toml":
+            print('==> Regenerating "{}"...'.format(entry))
 
-            args = [conf.gir_path, '-c', entry_file, '-o', path, '-d', conf.gir_files_path]
+            args = [conf.gir_path, '-c', entry, '-o', entry.parent, '-d', conf.gir_files_path]
             if level > 1:
-                args.append('-m')
-                args.append('sys')
+                args.extend(['-m', 'sys'])
             error = False
             try:
                 error = not run_command(args)
@@ -107,11 +104,14 @@ def parse_args():
     parser = argparse.ArgumentParser(description='Helper to regenerate gtk-rs crates using gir.',
                                      formatter_class=argparse.ArgumentDefaultsHelpFormatter)
 
-    parser.add_argument('path', nargs="*", default='.',
+    parser.add_argument('path', nargs="*", default=[Path('.')],
+                        type=Path,
                         help='Paths in which to look for Gir.toml files')
     parser.add_argument('--gir-files', dest="gir_files_path", default=DEFAULT_GIR_FILES_DIRECTORY,
+                        type=Path,
                         help='Path of the gir-files folder')
     parser.add_argument('--gir-path', default=DEFAULT_GIR_PATH,
+                        type=Path,
                         help='Path of the gir executable to run')
     parser.add_argument('--yes', action='store_true',
                         help=' Always answer `yes` to any question asked by the script')
@@ -127,14 +127,14 @@ def main():
     if conf.gir_files_path == DEFAULT_GIR_FILES_DIRECTORY:
         if def_check_submodule(conf.gir_files_path, conf) == FAILURE:
             return 1
-    elif not isdir(conf.gir_files_path):
+    elif not conf.gir_files_path.is_dir():
         print("`{}` dir doesn't exist. Aborting...".format(conf.gir_files_path))
         return 1
 
     if conf.gir_path == DEFAULT_GIR_PATH:
         if not build_gir_if_needed(def_check_submodule(DEFAULT_GIR_DIRECTORY, conf)):
             return 1
-    elif not isfile(conf.gir_path):
+    elif not conf.gir_path.is_file():
         print("`{}` file doesn't exist. Aborting...".format(conf.gir_path))
         return 1
 
--- a/generator.py
+++ b/generator.py
@@ -11,8 +11,9 @@ NOTHING_TO_BE_DONE = 0
 NEED_UPDATE = 1
 FAILURE = 2
 
-DEFAULT_GIR_DIRECTORY = 'gir-files'
-DEFAULT_GIR_PATH = './gir/target/release/gir'
+DEFAULT_GIR_FILES_DIRECTORY = './gir-files'
+DEFAULT_GIR_DIRECTORY = './gir/'
+DEFAULT_GIR_PATH = DEFAULT_GIR_DIRECTORY + 'target/release/gir'
 
 
 def run_command(command, folder=None):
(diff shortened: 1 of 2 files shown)


sdroege:
CC @sophie-h

MarijnS95:
Yeah, thanks @sophie-h for doing the argparse conversion!  I thought about that ever since the previous revision landed on `gstreamer-rs` past week.

generator.py
```diff
@@ -77,18 +73,16 @@ def build_gir_if_needed(updated_submodule):
...
-        if isdir(entry_file):
-            if level < 2 and not regen_crates(entry_file, conf, level + 1):
+    for entry in path.iterdir():
+        if entry.is_dir():
+            if level < 2 and not regen_crates(entry, conf, level + 1):
                 return False
-        elif entry.startswith("Gir") and entry.endswith(".toml"):
-            print('==> Regenerating "{}"...'.format(entry_file))
+        elif entry.name.startswith("Gir") and entry.suffix == ".toml":
+            print('==> Regenerating "{}"...'.format(entry))
```
GuillaumeGomez, line 81: Just to confirm: it prints the "full path" right?

MarijnS95: Yup, `entry.name` is used intentionally in the line above to get just the file name.

generator.py
```diff
@@ -11,32 +10,29 @@
 NEED_UPDATE = 1
 FAILURE = 2
 
-DEFAULT_GIR_DIRECTORY = 'gir-files'
-DEFAULT_GIR_PATH = './gir/target/release/gir'
+DEFAULT_GIR_FILES = Path('./gir-files')
```
GuillaumeGomez, line 13: I suggest `DEFAULT_GIR_FILES_DIRECTORY` instead to be consistent with `DEFAULT_GIR_DIRECTORY`.

generator.py
```diff
@@ -11,32 +10,29 @@
...
-    except:
-        return False
+    return run_command(['cargo', 'build', '--release'], 'gir')
 
 
 def ask_yes_no_question(question, conf):
     question = '{} [y/N] '.format(question)
     if conf.yes:
-        print(question)
+        print(question + 'y')
```
GuillaumeGomez, line 35: Why adding the `y` there?

generator.py
MarijnS95: Makes sense, will update :)

generator.py
MarijnS95: I'd rather not print this at all. It's a bit weird to see `[y/N]` (fullcaps `N` meaning default no show up) but the program continuing, with seemingly no answer given.

generator.py
MarijnS95: How about the command-line option, is `--gir-files` okay with you?

generator.py
GuillaumeGomez: Well, I guess it makes sense in a way. Let's go for that then!

generator.py
GuillaumeGomez: :+1:

generator.py
GuillaumeGomez: The option yes, but the field not really. Could you change it to `gir_files_path` please (same as `gir_path` :) )?

bilelmoussaoui:
Would be nice to send a PR for the same changes to gtk4-rs if possible. Otherwise I can backport them once merged here

generator.py
MarijnS95: That name is derived from the flag, butt can be overridden with `dest=`.

On that note I don't find `--gir-path` very expressive, should we perhaps rename that to just `--gir` or `--gir-exe(cutable)`? We can also add some short-hands like `-g`, `-f`, '-y` etc.

MarijnS95:
> Would be nice to send a PR for the same changes to gtk4-rs if possible. Otherwise I can backport them once merged here

@bilelmoussaoui Sure thing, I can apply the same commits there as soon as this passes review :)

generator.py
GuillaumeGomez: Please don't add short-hands. I voluntarily didn't add them. As for the name, I find `--gir-path` good enough but others opinions can't hurt. cc @bilelmoussaoui @sdroege

bilelmoussaoui: --gir-path is good for me as well, but no hard opinions
@sophie-h any opinions?

MarijnS95: Renamed the field to `gir_files_path`. I'm not using `--gir-path` yet so I have no stake in the choice :)

sdroege: Let's add short versions of the parameters in a separate PR as this apparently is contentious and needs to be discussed, while everything else in this PR is easy. (FWIW I don't care either way)

sdroege (approved):

sdroege:
Looks good to me but why is `regen_check` failing on the CI? Looks like the gir-files submodule is out of sync with the code. Does one of the changes here accidentally update the submodule automatically from `generator.py`? :)

generator.py
MarijnS95: @sdroege Fyi this is not about shorthands as much as it is about the naming of `--gir-path`. But whatever, this is a small development utility, not a public facing API :)

sdroege: I also don't care either way for that one :)

MarijnS95:
@sdroege The same happens in the `argparse` PR: https://github.com/gtk-rs/gtk-rs/runs/1438378458

`actions/checkout` in CI.yml is not set to `submodules: recursive` (or `true`) meaning that `./generator.py` gets to initialize them. Because `--yes` is passed, it'll always _pull_ `master` which is not what `gir-files` points to currently. Consequently, `git` wants you to be more specific about the `pull` strategy as well:

```
HEAD is now at 062b3d6 Merge pull request #995 from sdroege/remove-setvalueoptional-trait-bound
warning: Pulling without specifying how to reconcile divergent branches is
discouraged. You can squelch this message by running one of the following
commands sometime before your next pull:

  git config pull.rebase false  # merge (the default strategy)
  git config pull.rebase true   # rebase
  git config pull.ff only       # fast-forward only

You can replace "git config" with "git config --global" to set a default
preference for all repositories. You can also pass --rebase, --no-rebase,
or --ff-only on the command line to override the configured default per
invocation.
```

So:
- Does it even make sense to pass `--yes` in CI?
- Does it even make sense to have `pull -f origin master` in the generator?
  - Either this should be a manual action;
  - Or `reset`, or `--ff-only`.
- Have a separate flag for saying yes to continue after an error, and to update submodules to master.

sdroege:
I don't think the submodules should be updated ever in the CI. That defeats the whole point.

MarijnS95:
@sdroege Looks like it passes now. We have to sync the submodules directly from `CI.yml` otherwise the script will do the initial sync and ask for submodules, which fails: https://github.com/gtk-rs/gtk-rs/runs/1441060184?check_suite_focus=true

If we want to test this part of the codepath we might as well: `yes n | python3 generator.py`, what do you prefer?

sdroege:
Something for @bilelmoussaoui and @GuillaumeGomez to decide :)

GuillaumeGomez:
We could simply add a `--no` which would answer `no` to all questions. What do you think? Like that, the CI can't be stuck and we don't update anything that shouldn't be.

MarijnS95:
@GuillaumeGomez In all honesty I feel like removing `input` altogether, in favor of explicit options: `--continue-on-error` and `--update-submodules-to-master` or something like that.

GuillaumeGomez:
I prefer keeping the input. Generally it means that an issue occurred and I'd like to be able to keep a given state and choosing whether or not to continue depending on the error.

bilelmoussaoui:
> @sdroege The same happens in the `argparse` PR: https://github.com/gtk-rs/gtk-rs/runs/1438378458
> 
> `actions/checkout` in CI.yml is not set to `submodules: recursive` (or `true`) meaning that `./generator.py` gets to initialize them. Because `--yes` is passed, it'll always _pull_ `master` which is not what `gir-files` points to currently, resulting in the diff. Consequently, `git` wants you to be more specific about the `pull` strategy as well:
> 
> ```
> HEAD is now at 062b3d6 Merge pull request #995 from sdroege/remove-setvalueoptional-trait-bound
> warning: Pulling without specifying how to reconcile divergent branches is
> discouraged. You can squelch this message by running one of the following
> commands sometime before your next pull:
> 
>   git config pull.rebase false  # merge (the default strategy)
>   git config pull.rebase true   # rebase
>   git config pull.ff only       # fast-forward only
> 
> You can replace "git config" with "git config --global" to set a default
> preference for all repositories. You can also pass --rebase, --no-rebase,
> or --ff-only on the command line to override the configured default per
> invocation.
> ```
> 
> So:
> 
>     * Does it even make sense to pass `--yes` in CI?
> 
>     * Does it even make sense to have `pull -f origin master` in the generator?
>       
>       * Either this should be a manual action;
>       * Or `reset`, or `--ff-only`.
> 
>     * Have a separate flag for saying yes to continue after an error, and to update submodules to master.

We use an action to checkout and clone a repository, you can pass a parameter to it and ask it to clone the repository recursively

```yaml

  generator:
    name: regen check
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v2
        with:
          submodules: recursive
      - run: python3 generator.py --yes
      - run: git diff -R --exit-code
``` 
See https://github.com/actions/checkout for the various options you can pass to it

MarijnS95:
> I prefer keeping the input. Generally it means that an issue occurred and I'd like to be able to keep a given state and choosing whether or not to continue depending on the error.

@GuillaumeGomez We can do both, the input for the error and the flag to forcibly update the submodule to `master`?

@bilelmoussaoui See the last commit and my suggestion to add `submodules: recursive` in _exactly_ the comment you quoted :stuck_out_tongue_winking_eye:

sdroege:
So all good here now or ...? :)

bilelmoussaoui (approved):

sdroege:
@GuillaumeGomez ?

GuillaumeGomez (approved):

sdroege:
Ok, then this only needs someone to fix the merge conflicts :)

MarijnS95:
@sdroege Someone, not me. @GuillaumeGomez [requested](https://github.com/gtk-rs/gtk-rs/pull/170#discussion_r528364147) 2 days ago to rename `DEFAULT_GIR_FILES_DIRECTORY` which I promptly did, but went ahead and made + self-approve [the same change](https://github.com/gtk-rs/gtk-rs/pull/174) a couple hours ago plus some other renames, conflicting with most commits in this PR. Not nice. Could have asked or waited until this PR landed, would have been easier for both of us.

sdroege:
That's annoying and not very nice indeed. @GuillaumeGomez can you merge this PR then? You broke it :)

GuillaumeGomez:
Arf sorry, completely forgot about this one. I was doing a gir change and noticed that. I'm very sorry.

If you want I can fix the conflict? Again, I'm very sorry I forgot about it...

generator.py
```diff
@@ -147,6 +147,9 @@ def main():
     if conf.gir_files_path == DEFAULT_GIR_FILES_DIRECTORY:
         if def_check_submodule(conf.gir_files_path, conf) == FAILURE:
             return 1
+    elif not isdir(conf.gir_files_path):
+        print("`{}` dir doesn't exist. Aborting...".format(conf.gir_files_path))
+        return 1
```
MarijnS95: Careful, this is dealt with by `def directory_path(path):` above, and should otherwise be `conf.gir_files_path.is_dir()`.

MarijnS95:
@GuillaumeGomez No worries, if it's a genuine oversight then that's okay.  And merging is easier than rebasing, too :D

generator.py
GuillaumeGomez: Arf! Removing useless code I added back.

generator.py
```diff
@@ -127,7 +127,7 @@ def parse_args():
     parser.add_argument('path', nargs="*", default=[Path('.')],
                         type=valid_path,
                         help='Paths in which to look for Gir.toml files')
-    parser.add_argument('--gir-files', dest="gir_files_path", default=DEFAULT_GIR_FILES_DIRECTORY,
+    parser.add_argument('--gir-files-directory', dest="gir_files_path", default=DEFAULT_GIR_FILES_DIRECTORY,
```
MarijnS95, line 130: Do we want to persist using `dest="gir_files_path"` or switch to `gir_files_directory` from your change?

GuillaumeGomez: I kept your wording. But both are fine for me.

GuillaumeGomez:
And I removed the unnecessary condition I added back. I'm really sorry about this mess. As soon as CI is green, I merge. Thanks a lot for this! I discovered a lot of new things about python in this PR too. :)

MarijnS95:
@GuillaumeGomez Cool, thanks for fixing it up! Conversely I'm learning a lot about gir/gstreamer/gtk in Rust and the support code around it, thanks for the ride yall!

sdroege:
@MarijnS95 Thanks for the improvements :)

generator.py
```diff
@@ -11,32 +10,29 @@
...
 
 
 def ask_yes_no_question(question, conf):
     question = '{} [y/N] '.format(question)
     if conf.yes:
-        print(question)
+        print(question + 'y')
         return True
     if sys.version_info[0] < 3:
         line = raw_input(question)
```
sdroege, line 38: Oops, there is no function `raw_input()` anywhere. Glad we're using Python so nobody noticed this ;)

MarijnS95: @sdroege Should be in Python 2, which is what this code falls back to?

Then again Python 2 is deprecated for almost a year now, and the shebang specifies Python 3, might as well remove this in a followup PR :)

sdroege: Oh I just had my editor complain that this function doesn't exist (and I don't have Python 2 installed)

bilelmoussaoui: The code uses `Path` & some new fancy python 3 features, it should be safe to remove the python 2 fallback I would say.

GuillaumeGomez: wouuuups

sdroege: Not a problem really, but we should remove the Python 2 backwards compat code. This is 2020.

GuillaumeGomez: Absolutely, it's just that I'm so used to write this code that I didn't pay attention. Sending a PR.

MarijnS95:
There's one more thing I have in the pipeline on my gstreamer-rs fork: autoformat this file with [black](https://github.com/psf/black) to make future alterations easier. Anyone interested? Keep in mind this will normalize (currently mixed between `'` and `"`) quotation usage.

GuillaumeGomez:
Please go ahead!

MarijnS95:
@GuillaumeGomez Done! Hope that's the last of 'em, and we can finally get back to writing Rust :)

GuillaumeGomez:
Thanks a lot!

PR #175, opened 2020-11-24
Removing python2 code

GuillaumeGomez:
There is no point in keeping this check considering that we don't support python 2.

cc @MarijnS95 @sdroege


sdroege (approved):
That was the only Python 2 compat code?

GuillaumeGomez:
As far as I can tell, yes. I never intended to support python 2 so it's mostly a bad habit...

PR #179, opened 2020-11-26
glib::Binding improvements

sdroege:
Fixes some bugs and makes it automatically thread-safe once https://gitlab.gnome.org/GNOME/glib/-/merge_requests/1745 is in and the user is using GLib 2.68 or newer.


glib/Gir_GObject.toml
```diff
@@ -29,16 +29,24 @@ name = "GObject.Binding"
 status = "generate"
 trait = false
 concurrency = "send+sync"
-       # Fixed in GLib 2.66.2
        [[object.function]]
        name = "get_source"
-           [object.function.return]
-           nullable = true
-       # Fixed in GLib 2.66.2
+       # Not-threadsafe, see https://gitlab.gnome.org/GNOME/glib/-/merge_requests/1745
```
GuillaumeGomez, line 34:
```suggestion
       # Not thread-safe, see https://gitlab.gnome.org/GNOME/glib/-/merge_requests/1745
```

glib/Gir_GObject.toml
```diff
@@ -29,16 +29,24 @@ name = "GObject.Binding"
...
+       # Using the GObject properties will be automatically safe with GLib 2.68.
+       ignore = true
        [[object.function]]
        name = "get_target"
-           [object.function.return]
-           nullable = true
+       # Not-threadsafe, see https://gitlab.gnome.org/GNOME/glib/-/merge_requests/1745
```
GuillaumeGomez, line 39:
```suggestion
       # Not thread-safe, see https://gitlab.gnome.org/GNOME/glib/-/merge_requests/1745
```

glib/Gir_GObject.toml
sdroege: Thanks, fixed both

glib/src/gobject/binding.rs
```diff
@@ -0,0 +1,38 @@
+// Copyright 2020, The Gtk-rs Project Developers.
+// See the COPYRIGHT file at the top-level directory of this distribution.
+// Licensed under the MIT license, see the LICENSE file or <https://opensource.org/licenses/MIT>
+use crate::translate::*;
```
GuillaumeGomez, line 4: Please add an empty line after the license comment. :)

GuillaumeGomez (approved):

glib/src/gobject/binding.rs
sdroege: Done

GuillaumeGomez:
Then it's ready to go. Thanks!

PR #181, opened 2020-11-28
glib: implement FromGlibContainer for GString

bilelmoussaoui:
Needed by https://github.com/gtk-rs/gtk4-rs/pull/89

### DIFF
--- a/glib/src/gstring.rs
+++ b/glib/src/gstring.rs
@@ -404,6 +404,80 @@ impl<'a> ToGlibPtr<'a, *mut c_char> for GString {
     }
 }
 
+#[doc(hidden)]
+impl<'a> FromGlibContainer<*const c_char, *const i8> for GString {
+    unsafe fn from_glib_none_num(ptr: *const i8, num: usize) -> Self {
+        if num == 0 || ptr.is_null() {
+            return Self::from("");
+        }
+        let mut bytes = Vec::with_capacity(num + 1);
+        let slice = std::slice::from_raw_parts(ptr as *const u8, num);
+        bytes.extend_from_slice(slice);
+        bytes.push(0);
+
+        CStr::from_bytes_with_nul_unchecked(bytes.as_slice()).into()
+    }
+
+    unsafe fn from_glib_container_num(ptr: *const i8, num: usize) -> Self {
+        if num == 0 || ptr.is_null() {
+            return GString::from("");
+        }
+        GString(Inner::Foreign(ptr as *mut _, num))
+    }
+
+    unsafe fn from_glib_full_num(ptr: *const i8, num: usize) -> Self {
+        if num == 0 || ptr.is_null() {
+            return GString::from("");
+        }
+        GString(Inner::Foreign(ptr as *mut _, num))
+    }
+}
+
+#[doc(hidden)]
+impl<'a> FromGlibContainer<*const c_char, *mut i8> for GString {
+    unsafe fn from_glib_none_num(ptr: *mut i8, num: usize) -> Self {
+        FromGlibContainer::from_glib_none_num(ptr as *const i8, num)
+    }
+
+    unsafe fn from_glib_container_num(ptr: *mut i8, num: usize) -> Self {
+        FromGlibContainer::from_glib_none_num(ptr as *const i8, num)
+    }
+
+    unsafe fn from_glib_full_num(ptr: *mut i8, num: usize) -> Self {
+        FromGlibContainer::from_glib_none_num(ptr as *const i8, num)
+    }
+}
+
+#[doc(hidden)]
+impl<'a> FromGlibContainer<*const c_char, *const u8> for GString {
+    unsafe fn from_glib_none_num(ptr: *const u8, num: usize) -> Self {
+        FromGlibContainer::from_glib_none_num(ptr as *const i8, num)
+    }
+
+    unsafe fn from_glib_container_num(ptr: *const u8, num: usize) -> Self {
+        FromGlibContainer::from_glib_none_num(ptr as *const i8, num)
+    }
+
+    unsafe fn from_glib_full_num(ptr: *const u8, num: usize) -> Self {
+        FromGlibContainer::from_glib_none_num(ptr as *const i8, num)
+    }
+}
+
+#[doc(hidden)]
+impl<'a> FromGlibContainer<*const c_char, *mut u8> for GString {
+    unsafe fn from_glib_none_num(ptr: *mut u8, num: usize) -> Self {
+        FromGlibContainer::from_glib_none_num(ptr as *const i8, num)
+    }
+
+    unsafe fn from_glib_container_num(ptr: *mut u8, num: usize) -> Self {
+        FromGlibContainer::from_glib_none_num(ptr as *const i8, num)
+    }
+
+    unsafe fn from_glib_full_num(ptr: *mut u8, num: usize) -> Self {
+        FromGlibContainer::from_glib_none_num(ptr as *const i8, num)
+    }
+}
+
 impl GlibPtrDefault for GString {
     type GlibType = *const c_char;
 }
@@ -450,7 +524,7 @@ impl_from_glib_container_as_vec_string!(GString, *mut c_char);
 #[cfg(test)]
 #[allow(clippy::blacklisted_name)]
 mod tests {
-    use super::GString;
+    use super::{FromGlibContainer, GString};
     use std::ffi::CString;
 
     #[test]
@@ -509,6 +583,32 @@ mod tests {
         let v: &[u8] = b"foo";
         let s: GString = Vec::from(v).into();
         assert_eq!(s.as_str(), "foo");
+    }
+
+    #[test]
+    fn test_from_glib_container() {
+        unsafe {
+            let test_a: GString = FromGlibContainer::from_glib_container_num(
+                ffi::g_strdup("hello_world".as_ptr() as *const i8),
+                5,
+            );
+            assert_eq!("hello", test_a.as_str());
+
+            let test_b: GString = FromGlibContainer::from_glib_none_num("hello_world".as_ptr(), 5);
+            assert_eq!("hello", test_b.as_str());
+
+            let test_c: GString = FromGlibContainer::from_glib_none_num(std::ptr::null::<i8>(), 0);
+            assert_eq!("", test_c.as_str());
+
+            let test_d: GString = FromGlibContainer::from_glib_none_num("".as_ptr(), 0);
+            assert_eq!("", test_d.as_str());
+
+            let test_e: GString = FromGlibContainer::from_glib_container_num(
+                ffi::g_strdup(std::ptr::null::<i8>()),
+                0,
+            );
+            assert_eq!("", test_e.as_str());
+        }
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


glib/src/gstring.rs
```diff
@@ -404,6 +404,48 @@ impl<'a> ToGlibPtr<'a, *mut c_char> for GString {
     }
 }
 
+#[doc(hidden)]
+impl<'a> FromGlibContainer<*const c_char, *const i8> for GString {
+    unsafe fn from_glib_none_num(ptr: *const i8, num: usize) -> Self {
```
bilelmoussaoui, line 409: To be honest, I'm not very sure if this was the right approach at all :) I'm a bit confused what would be the differences between the 3 methods in such case?

sdroege: `none` doesn't take ownership of the input pointer, the other two would (and are the same because `char` is a basic type without memory management).

glib/src/gstring.rs
```diff
@@ -404,6 +404,48 @@ impl<'a> ToGlibPtr<'a, *mut c_char> for GString {
...
+        let mut bytes = Vec::with_capacity(num);
+
+        for i in 0..num {
+            let b = std::ptr::read(ptr.add(i)) as u8;
+            bytes.push(b);
+        }
+        CStr::from_bytes_with_nul_unchecked(bytes.as_slice()).into()
+    }
+
+    unsafe fn from_glib_container_num(ptr: *const i8, num: usize) -> Self {
```
sdroege, line 422: `container` and `full` could simply create a `GString(Inner::Foreign(ptr as *mut _, num))` without copying or anything

glib/src/gstring.rs
```diff
@@ -404,6 +404,48 @@ impl<'a> ToGlibPtr<'a, *mut c_char> for GString {
...
+        if num == 0 {
+            return Self::from("");
+        }
+        let mut bytes = Vec::with_capacity(num);
+
+        for i in 0..num {
+            let b = std::ptr::read(ptr.add(i)) as u8;
+            bytes.push(b);
+        }
+        CStr::from_bytes_with_nul_unchecked(bytes.as_slice()).into()
```
sdroege, line 419: Does this actually work? I believe that function requires the vec to be NUL terminated or otherwise it's UB. You could add a 0 byte in the end though

glib/src/gstring.rs
```diff
@@ -404,6 +404,48 @@ impl<'a> ToGlibPtr<'a, *mut c_char> for GString {
...
+    unsafe fn from_glib_none_num(ptr: *const i8, num: usize) -> Self {
+        if num == 0 {
+            return Self::from("");
+        }
+        let mut bytes = Vec::with_capacity(num);
+
+        for i in 0..num {
+            let b = std::ptr::read(ptr.add(i)) as u8;
+            bytes.push(b);
+        }
```
sdroege, line 418:
```rust
let slice = std::slice::from_raw_parts(ptr as *const u8, num);
bytes.extend_from_slice(slice);
```

That does the same as your loop but more efficiently and shorter.

bilelmoussaoui: Ah I somehow got confused by the docs of slice::from_raw_parts where it says that num shouldn't be the number of bytes but rather the number of elements.

sdroege: That's the same for the `FromGlibContainer` trait btw (and for `u8` / `i8` it's all the same anyway).

glib/src/gstring.rs
bilelmoussaoui: Should be working now

glib/src/gstring.rs
bilelmoussaoui: I've added a very simple test just to check if everything was working fine but the `container` & `full` one fails with a `free(): invalid pointer`, I wonder whether what I wrote is just completely wrong on that test or something isn't working properly? :)

glib/src/gstring.rs
bilelmoussaoui: Noted, thanks!

glib/src/gstring.rs
```diff
@@ -404,6 +404,29 @@ impl<'a> ToGlibPtr<'a, *mut c_char> for GString {
     }
 }
 
+#[doc(hidden)]
+impl<'a> FromGlibContainer<*const c_char, *const i8> for GString {
```
sdroege, line 408: For completeness, can you also implement it for `*mut _` and `u8`?

glib/src/gstring.rs
```diff
@@ -404,6 +404,29 @@ impl<'a> ToGlibPtr<'a, *mut c_char> for GString {
...
+    unsafe fn from_glib_none_num(ptr: *const i8, num: usize) -> Self {
+        if num == 0 {
+            return Self::from("");
+        }
+        let mut bytes = Vec::with_capacity(num);
```
sdroege, line 413:
```suggestion
        let mut bytes = Vec::with_capacity(num + 1);
```

For the 0 :)

glib/src/gstring.rs
```diff
@@ -511,6 +534,21 @@ mod tests {
...
     }
 
+    #[test]
+    fn test_from_glib_container() {
+        unsafe {
+            let a: GString =
+                FromGlibContainer::from_glib_container_num("hello_world".as_ptr() as *const i8, 5);
+            assert_eq!("hello", a.as_str());
+            let b: GString =
+                FromGlibContainer::from_glib_full_num("hello_world".as_ptr() as *const i8, 5);
```
sdroege, line 544:
```suggestion
                FromGlibContainer::from_glib_full_num(glib_sys::g_strdup("hello_world".as_ptr() as *const i8), 5);
```

Or similar. The problem is that you pass a static (and definitely not heap allocated with the GLib allocator) string in there but say that `GString` can take ownership of it. That doesn't work :)

Same for the container variant as they're really the same here.

glib/src/gstring.rs
sdroege: You can implement that all by calling into the other implementation by casting the raw pointers.

glib/src/gstring.rs
sdroege: See comments below

glib/src/gstring.rs
sdroege: Otherwise you'd allocate `num` bytes, copy the string, append `0` and that appending would allocate another memory area on the heap of at least `num+1` bytes, copy everything first, then free the old memory, then append the `0`.

glib/src/gstring.rs
bilelmoussaoui: > You can implement that all by calling into the other implementation by casting the raw pointers.

I was going to copy paste them :P

glib/src/gstring.rs
bilelmoussaoui: I should have looked at the other tests first..

glib/src/gstring.rs
bilelmoussaoui: Interesting, thanks as always!

glib/src/gstring.rs
```diff
@@ -404,6 +404,74 @@ impl<'a> ToGlibPtr<'a, *mut c_char> for GString {
...
+
+        CStr::from_bytes_with_nul_unchecked(bytes.as_slice()).into()
+    }
+
+    unsafe fn from_glib_container_num(ptr: *const i8, num: usize) -> Self {
+        GString(Inner::Foreign(ptr as *mut _, num))
+    }
+
+    unsafe fn from_glib_full_num(ptr: *const i8, num: usize) -> Self {
+        GString(Inner::Foreign(ptr as *mut _, num))
```
sdroege, line 426: I wonder if these work correctly for empty strings or some fixes would be needed. Can you add a test for empty strings in the different variants?

sdroege: I.e. `NULL` / `0` and `$random_pointer` / `0` should give the same results for `none` and `full`

bilelmoussaoui: I have added more cases to the test

glib/src/gstring.rs
```diff
@@ -511,6 +579,32 @@ mod tests {
...
+
+            let c: GString =
+                FromGlibContainer::from_glib_none_num(std::ptr::null() as *const i8, 0);
+            assert_eq!("", c.as_str());
+
+            let d: GString = FromGlibContainer::from_glib_none_num("".as_ptr(), 0);
+            assert_eq!("", d.as_str());
+
+            let e: GString = FromGlibContainer::from_glib_container_num(
+                ffi::g_strdup(std::ptr::null() as *const i8),
```
sdroege, line 601: `g_strdup()` returns `NULL` if passed `NULL` or not? And the cast should be unneeded

glib/src/gstring.rs
```diff
@@ -511,6 +579,32 @@ mod tests {
...
+                0,
```
sdroege, line 602: And I think this actually is UB because it calls `slice::from_raw_parts()` somewhere with `NULL`. See https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html

The `as_str()` and other functions further above would need to special-case `NULL` / `0` and simply return a `""` in that case.

glib/src/gstring.rs
bilelmoussaoui: Yes, per the function documentation 
> Duplicates a string. If str is NULL it returns NULL. The returned string should be freed with g_free() when no longer needed.

The cast is needed because otherwise I keep getting 
```
 FromGlibContainer::from_glib_none_num(std::ptr::null(), 0);
     |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ cannot infer type for struct `gstring::GString`
```

glib/src/gstring.rs
bilelmoussaoui: Does that has to happen on all other functions? otherwise I think we can just add a check to `full` & `container` implementation, the same way I did for the `none` one

glib/src/gstring.rs
sdroege: `std::ptr::null::<i8>()` might be nicer than a cast then :)

bilelmoussaoui: Aha. Thanks! I wasn't aware I can do this :)

glib/src/gstring.rs
sdroege: Well, `GString::from_glib_full(ptr::null())` seems to be allowed and would end up in exactly the same situation (before your changes already).

Probably best to handle that generally somewhere.

bilelmoussaoui: It should be handled now

glib/src/gstring.rs
```diff
@@ -62,13 +61,15 @@ impl GString {
...
     /// length, underneath the `GString` instance for the duration of the borrow.
     unsafe fn new_borrowed(ptr: *const c_char) -> Borrowed<Self> {
-        assert!(!ptr.is_null());
         Borrowed::new(GString(Inner::Foreign(ptr as *mut _, libc::strlen(ptr))))
     }
 
     pub fn as_str(&self) -> &str {
         let cstr = match self {
             GString(Inner::Foreign(ptr, length)) => unsafe {
+                if ptr.is_null() || length == &0 {
```
sdroege, line 70: Why `&0` instead of `0`? :)

bilelmoussaoui: Because length is &usize and there's no PartialEq impl for that

sdroege: Ok!

sdroege (approved):

PR #184, opened 2020-11-30
Fix clone macro by allowing returned value with type parameters

### LINKED ISSUE #183: Confusing clone! macro error message
I upgraded to gtk-rs master, now I get this "confusing" error message :-)

```
error: proc macro panicked
   --> ../src/ui/package_widgets/action_button.rs:112:23
    |
112 |           let closure = clone!(@weak self.builder as builder, @weak package => @default-return None::<glib::Value>, move |_:&[glib::Value]|{
    |  _______________________^
113 | |             Self::update_stack(builder.clone(), package.clone());
114 | |             None
115 | |         });
    | |__________^
    |
    = help: message: Expected `,` after `@default-return None::<`, found `,`
```

No idea if this can be improved a bit ^^

GuillaumeGomez:
Fixes #183.

It also puts back the clone macro check on CI.

cc @haecker-felix

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -123,12 +123,20 @@ jobs:
       - name: "glib: build nightly"
         run: cargo build --manifest-path glib/Cargo.toml --all-features
         if: matrix.rust == 'nightly'
+      - name: "glib: tests glib macros nightly"
+        working-directory: glib
+        run: python3 clone_tests/clone_compilation_errors.py
+        if: matrix.rust == 'nightly'
       #not nightly
       - name: "glib: tests"
         run: xvfb-run --auto-servernum cargo test --manifest-path glib/Cargo.toml --features "v2_66"
         if: matrix.rust != 'nightly'
       - name: "glib: build"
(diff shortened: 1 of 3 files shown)


.github/workflows/CI.yml
```diff
@@ -123,13 +123,21 @@ jobs:
       - name: "glib: build nightly"
         run: cargo build --manifest-path glib/Cargo.toml --all-features
         if: matrix.rust == 'nightly'
+      - name: "glib: tests glib macros nightly"
+        working-directory: glib
+        run: python3 clone_tests/clone_compilation_errors.py
+        if: matrix.rust == 'nightly'
```
sdroege, line 129: Why only on nightly?

sdroege (commented):
Seems ok to me otherwise, thanks :)

.github/workflows/CI.yml
GuillaumeGomez: Arf! I forgot to remove this check on the second.

sdroege (approved):

PR #188, opened 2020-12-01
Subclassing for GtkButton, backported from gtk4-rs

ids1024:
Was added to gtk4-rs in https://github.com/gtk-rs/gtk4-rs/pull/57

This doesn't include the deprecated virtual methods in gtk3 that are not in gtk4. Those could be added, but new projects probably shouldn't be using them anyway.


sdroege:
Thanks!

PR #189, opened 2020-12-02
Backport composite templates from gtk4-rs

BrainBlasted:
Backports support for composite templates from gtk4-rs.

We can't use gtk-macros as the name of the proc macro crate due to a conflict with an existing gtk-macros crate.


GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #191, opened 2020-12-04
Add ListModel subclassing support

bilelmoussaoui:


gio/src/subclass/list_store.rs
```diff
@@ -0,0 +1,13 @@
...
+
+use crate::ListStore;
+use glib::subclass::object::ObjectImpl;
+use glib::Object;
+
+pub trait ListStoreImpl: ObjectImpl {}
+
+unsafe impl<T: ListStoreImpl> IsSubclassable<T> for ListStore {
+    fn override_vfuncs(class: &mut glib::Class<Self>) {
+        <Object as IsSubclassable<T>>::override_vfuncs(class);
```
sdroege, line 11: This is not actually sub-classable :)

```
G_DECLARE_FINAL_TYPE(GListStore, g_list_store, G, LIST_STORE, GObject)
```

There's no public class struct. How does this currently compile at all? The `GListStoreClass` struct seems to only exist in the .gir file but not in the headers. That seems concerning and should probably be fixed (by removing it from the .gir file!). Can you create an issue for that (against glib) and remove it from here?

gio/src/subclass/list_model.rs
```diff
@@ -0,0 +1,55 @@
+use crate::auto::ListModel;
+use glib::subclass::prelude::*;
+use glib::translate::*;
+use glib::Cast;
+
+pub trait ListModelImpl: ObjectImpl {
+    fn get_item_type(&self, list_model: &Self::Type) -> glib::Type;
```
sdroege, line 7:
```
 * The item type of a #GListModel can not change during the life of the
 * model.
```

Maybe something we want to enforce somehow? This could cause big problems otherwise :)

sdroege (commented):
Looks good to me otherwise. Maybe you want to extend the `listbox_model` example with usage of this? You could store all the items in a custom model around a `Vec` or `BTreeMap` or something :)

gio/src/subclass/list_store.rs
bilelmoussaoui: Ah my bad. I knew you cannot subclass it (from Vala days) but the generated class struct confused me a I guess. Opened an issue https://gitlab.gnome.org/GNOME/glib/-/issues/2261

gio/src/subclass/list_model.rs
bilelmoussaoui: I have no idea how I enforce such thing, any suggestions?

sdroege: You could store the first type returned in object qdata and check against that afterwards, and also make sure that all items ever returned are of that type or a subclass of it.

sdroege: I.e. check at runtime and panic otherwise.

The alternatively would be to make this a `pub trait ListModelImpl<T: StaticType + IsA<Object>>` or something scary like that. Not sure how much headache that causes when implementing it though. Maybe bad, maybe not at all :)

bilelmoussaoui:
> Looks good to me otherwise. Maybe you want to extend the `listbox_model` example with usage of this? You could store all the items in a custom model around a `Vec` or `BTreeMap` or something :)

As a different example or by modifying the current one?

gio/src/subclass/list_model.rs
bilelmoussaoui: I have tired the at runtime check, seems to compile fine ^^ I will try adding an example and see if it works as expected.  Although I had to enforce `glib::ObjectType` on whatever that implements the `ListModel` interface.

sdroege:
> As a different example or by modifying the current one?

Modifying the current one. Using a `gio::ListStore` is boring

gio/src/subclass/list_model.rs
```diff
@@ -0,0 +1,83 @@
...
+    let instance = &*(list_model as *mut T::Instance);
+    let imp = instance.get_impl();
+
+    let type_ = imp
+        .get_item_type(from_glib_borrow::<_, ListModel>(list_model).unsafe_cast_ref())
+        .to_glib();
+    // Store the type so we can enforce that it doesn't change.
+    let key = glib::Quark::from_string("type");
+
+    match imp.get_qdata(key) {
```
sdroege, line 38: I completely forgot that we have non-sys bindings for qdata. Even better :) You might want to use those also in the `GIOStream` subclassing

gio/src/subclass/list_model.rs
```diff
@@ -0,0 +1,83 @@
...
+unsafe extern "C" fn list_model_get_item<T: ObjectExt + ListModelImpl>(
+    list_model: *mut ffi::GListModel,
+    position: u32,
+) -> *mut glib::gobject_ffi::GObject {
+    let instance = &*(list_model as *mut T::Instance);
+    let imp = instance.get_impl();
+
+    let type_ = imp
+        .get_qdata::<glib::Type>(glib::Quark::from_string("type"))
+        .expect("ListModel should define the item-type");
```
sdroege, line 69: Maybe `get_item_type()` was not called before this, in which case you probably want to initialize it here

gio/src/subclass/list_model.rs
```diff
@@ -0,0 +1,83 @@
...
+    list_model: *mut ffi::GListModel,
+) -> glib::ffi::GType {
+    let instance = &*(list_model as *mut T::Instance);
+    let imp = instance.get_impl();
+
+    let type_ = imp
+        .get_item_type(from_glib_borrow::<_, ListModel>(list_model).unsafe_cast_ref())
+        .to_glib();
+    // Store the type so we can enforce that it doesn't change.
+    let key = glib::Quark::from_string("type");
```
sdroege, line 36: `type` is very generic and could easily conflict with something a user is setting on this, for example.

Also might make sense to initialize the quark in a `once_cell::sync::Lazy`

gio/src/subclass/list_model.rs
sdroege: > Although I had to enforce glib::ObjectType on whatever that implements the ListModel interface.

Yeah I think we're missing something somewhere to bind all those traits together.

gio/src/subclass/list_model.rs
```diff
@@ -0,0 +1,83 @@
...
+        list_model_iface.get_item_type = Some(list_model_get_item_type::<T>);
+        list_model_iface.get_n_items = Some(list_model_get_n_items::<T>);
+        list_model_iface.get_item = Some(list_model_get_item::<T>);
+    }
+}
+
+unsafe extern "C" fn list_model_get_item_type<T: ObjectExt + ListModelImpl>(
+    list_model: *mut ffi::GListModel,
+) -> glib::ffi::GType {
+    let instance = &*(list_model as *mut T::Instance);
```
sdroege, line 29: It's the instance that needs to be `ObjectType` and on which you can call `set`/`get_qdata`. Not sure how this works at all at runtime currently :)

gio/src/subclass/list_store.rs
sdroege: That probably also means that we need to extend the "final type" heuristic in gir to consider instance structs without any fields at all as indicator. Can you create an issue for that?

bilelmoussaoui:
Hmm so my attempt with updating the example doesn't seem to actually work, it instantly panics (due to assert)
```
thread 'main' panicked at 'assertion failed: self.is::<T>()', /home/bilelmoussaoui/Projects/gtk-rs/glib/src/object.rs:299:9
stack backtrace:
   0: std::panicking::begin_panic
             at /home/bilelmoussaoui/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/panicking.rs:505
   1: glib::object::Cast::unsafe_cast_ref
             at /home/bilelmoussaoui/Projects/gtk-rs/glib/src/object.rs:299
   2: glib::object::Cast::upcast_ref
             at /home/bilelmoussaoui/Projects/gtk-rs/glib/src/object.rs:152
   3: <listbox_model::model::Model as core::convert::AsRef<gio::auto::list_model::ListModel>>::as_ref
             at /home/bilelmoussaoui/Projects/gtk-rs/glib/src/object.rs:990
   4: <O as gtk::auto::list_box::ListBoxExt>::bind_model::{{closure}}
             at /home/bilelmoussaoui/Projects/gtk-rs/gtk/src/auto/list_box.rs:551
   5: core::option::Option<T>::map
             at /home/bilelmoussaoui/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/option.rs:450
   6: <O as gtk::auto::list_box::ListBoxExt>::bind_model
             at /home/bilelmoussaoui/Projects/gtk-rs/gtk/src/auto/list_box.rs:551
   7: listbox_model::build_ui
             at ./src/bin/listbox_model.rs:117
   8: listbox_model::main::{{closure}}
             at ./src/bin/listbox_model.rs:268
   9: <O as gio::auto::application::ApplicationExt>::connect_activate::activate_trampoline
             at /home/bilelmoussaoui/Projects/gtk-rs/gio/src/auto/application.rs:555
  10: g_closure_invoke
  11: <unknown>
  12: g_signal_emit_valist
  13: g_signal_emit
  14: <unknown>
  15: g_application_run
  16: <O as gio::application::ApplicationExtManual>::run
             at /home/bilelmoussaoui/Projects/gtk-rs/gio/src/application.rs:20
  17: listbox_model::main
             at ./src/bin/listbox_model.rs:271
  18: core::ops::function::FnOnce::call_once
             at /home/bilelmoussaoui/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ops/function.rs:227
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```
The assert in object.rs doesn't give much context from a "gtk-rs user" perspective, maybe we should panic there with a proper reason message?

sdroege:
> Hmm so my attempt with updating the example doesn't seem to actually work, it instantly panics (due to assert)

Good that I asked for an example :)

> The assert in object.rs doesn't give much context from a "gtk-rs user" perspective, maybe we should panic there with a proper reason message?

Yeah that would be more helpful. Your problem is that you don't register with GObject that your type implements that interface, so the corresponding assertion ("is this thing a ListModel") fails.

You need basically [this](https://gitlab.freedesktop.org/gstreamer/gst-plugins-rs/-/blob/58786fa0b5f122f2ccd60277739fb1450b3a71f7/generic/file/src/filesrc/imp.rs#L148-151) but for the `ListModel` interface.

sdroege:
> You need basically [this](https://gitlab.freedesktop.org/gstreamer/gst-plugins-rs/-/blob/58786fa0b5f122f2ccd60277739fb1450b3a71f7/generic/file/src/filesrc/imp.rs#L148-151) but for the `ListModel` interface.

I'd like to add something more convenient for this, but not sure how that could look like. A type list as associated type (`type Interfaces = [gio::ListModel]`) on `ObjectSubclass` or something like that would do the job but that's very inconvenient with current Rust.

bilelmoussaoui:
> > Hmm so my attempt with updating the example doesn't seem to actually work, it instantly panics (due to assert)
> 
> Good that I asked for an example :)
> 
> > The assert in object.rs doesn't give much context from a "gtk-rs user" perspective, maybe we should panic there with a proper reason message?
> 
> Yeah that would be more helpful. Your problem is that you don't register with GObject that your type implements that interface, so the corresponding assertion ("is this thing a ListModel") fails.
> 
> You need basically [this](https://gitlab.freedesktop.org/gstreamer/gst-plugins-rs/-/blob/58786fa0b5f122f2ccd60277739fb1450b3a71f7/generic/file/src/filesrc/imp.rs#L148-151) but for the `ListModel` interface.

I think it would be nice to add that to the documentation then as I didn't found anything about interfaces on https://gtk-rs.org/gtk-rs/glib/subclass/index.html and I thought that it will "just work". I will send a PR for that :)

bilelmoussaoui:
Somehow I can't get my code to compile at all and it confuses me a lot because why Model wouldn't implement `glib::ObjectType` if it's a `glib::Object`? 
```

   Compiling gtk-rs-examples v0.0.1 (/home/bilelmoussaoui/Projects/gtk-rs/examples)
error[E0277]: the trait bound `model::imp::Model: glib::ObjectType` is not satisfied
  --> examples/src/bin/listbox_model.rs:57:14
   |
57 |         impl ListModelImpl for Model {
   |              ^^^^^^^^^^^^^ the trait `glib::ObjectType` is not implemented for `model::imp::Model`
   | 
  ::: /home/bilelmoussaoui/Projects/gtk-rs/gio/src/subclass/list_model.rs:7:51
   |
7  | pub trait ListModelImpl: ObjectImpl + ObjectExt + glib::ObjectType {
   |                                                   ---------------- required by this bound in `ListModelImpl`
```

gio/src/subclass/list_model.rs
```diff
@@ -0,0 +1,92 @@
...
+
+pub trait ListModelImpl: ObjectImpl {
+    fn get_item_type(&self, list_model: &Self::Type) -> glib::Type;
+    fn get_n_items(&self, list_model: &Self::Type) -> u32;
+    fn get_item(&self, list_model: &Self::Type, position: u32) -> Option<glib::Object>;
+}
+
+unsafe impl<T: ListModelImpl> IsImplementable<T> for ListModel
+where
+    <T as ObjectSubclass>::Type: IsA<glib::Object>,
```
sdroege, line 14: Instead of all these `where` clauses you could also just use the unsafe qdata API from `gobject_sys` btw

gio/src/subclass/list_model.rs
sdroege: I added a FIXME comment for that into the code

examples/src/bin/listbox_model.rs
```diff
@@ -19,6 +19,83 @@ use std::env::args;
...
+        }
+
+        pub fn append(&self, obj: &RowData) {
+            let self_ = imp::Model::from_instance(self);
+            self_.0.borrow_mut().push(obj.clone());
+        }
+
+        pub fn remove(&self, index: u32) {
+            let self_ = imp::Model::from_instance(self);
+            self_.0.borrow_mut().remove(index as usize);
```
sdroege, line 94: You need to call `self.items_changed()` here with the correct parameters so the view knows that the model changed

examples/src/bin/listbox_model.rs
```diff
@@ -19,6 +19,83 @@ use std::env::args;
...
+        pub fn new() -> Model {
+            glib::Object::new(Self::static_type(), &[])
+                .expect("Failed to create Model")
+                .downcast()
+                .expect("Created Model is of wrong type")
+        }
+
+        pub fn append(&self, obj: &RowData) {
+            let self_ = imp::Model::from_instance(self);
+            self_.0.borrow_mut().push(obj.clone());
```
sdroege, line 89: You need to call `self.items_changed()` here with the correct parameters so the view knows that the model changed

gio/src/subclass/list_model.rs
bilelmoussaoui: I just ended up using get_item_type instead of reading it from the qdata directly

examples/src/bin/listbox_model.rs
```diff
@@ -19,6 +19,86 @@ use std::env::args;
...
+            let index = self_.0.borrow().len() - 1;
```
sdroege, line 90: That will become `usize::MAX` for the first item because of wraparound... not sure if that's intended :)

sdroege, line 90: Maybe only borrow once here instead of twice, but make sure to drop the guard object before emitting the signal

gio/src/subclass/list_model.rs
```diff
@@ -0,0 +1,92 @@
...
+{
+    let instance = &*(list_model as *mut T::Instance);
+    let imp = instance.get_impl();
+    let model = from_glib_borrow::<_, ListModel>(list_model);
+
+    let item = imp.get_item(model.unsafe_cast_ref(), position);
+
+    item.map(|i| {
+        if let Some(type_) = model.get_qdata::<glib::Type>(*LIST_ITEM_TYPE_QUARK) {
```
sdroege, line 84: Why don't you just call `get_item_type()` here? Then you don't have to do the qdata things in multiple functions

sdroege:
Looks ok to me otherwise

gio/src/subclass/list_model.rs
bilelmoussaoui: I have updated it on a separate commit :)

gio/src/subclass/list_store.rs
bilelmoussaoui: https://github.com/gtk-rs/gir/issues/1013

bilelmoussaoui: That actually means there's some widgets/objects that shouldn't be really subclassable... I wonder how Vala or other languages do check that?

sdroege: You mean widgets that we already allow to subclass but that shouldn't? For Vala it would fail to compile because there is no instance struct in the C headers (because Vala goes via C in the end), same for the C++ (mm) bindings. Others probably just don't care? :) Try subclassing `GtkListBox` from Python maybe, but I'm 99% sure it actually works.

sdroege: Which other widgets btw?

Also I think if you would try actually subclassing `ListBox` from Rust the way you had it before, you would likely get a critical warning at runtime because we would create a too small C instance struct (as the actual one is not known from the .gir file).

examples/src/bin/listbox_model.rs
```diff
@@ -19,6 +19,90 @@ use std::env::args;
...
+        impl ObjectSubclass for Model {
+            const NAME: &'static str = "Model";
+            type Type = super::Model;
+            type ParentType = glib::Object;
+            type Instance = subclass::simple::InstanceStruct<Self>;
+            type Class = subclass::simple::ClassStruct<Self>;
+
+            glib_object_subclass!();
+
+            fn type_init(type_: &mut subclass::InitializingType<Self>) {
```
sdroege, line 45: Maybe add an explanatory comment here

examples/src/bin/listbox_model.rs
```diff
@@ -19,6 +19,90 @@ use std::env::args;
...
+        }
+
+        pub fn append(&self, obj: &RowData) {
+            let self_ = imp::Model::from_instance(self);
+            let index = {
+                let mut data = self_.0.borrow_mut();
+                data.push(obj.clone());
+                data.len() - 1
+            };
+            self.items_changed(index as u32, 0, 1);
```
sdroege, line 95: Maybe the `items_changed()` and that we do the borrowing above in a separate block could also use some explanatory comment

gio/src/subclass/list_model.rs
```diff
@@ -0,0 +1,92 @@
...
+
+pub trait ListModelImpl: ObjectImpl {
+    fn get_item_type(&self, list_model: &Self::Type) -> glib::Type;
+    fn get_n_items(&self, list_model: &Self::Type) -> u32;
+    fn get_item(&self, list_model: &Self::Type, position: u32) -> Option<glib::Object>;
+}
+
+unsafe impl<T: ListModelImpl> IsImplementable<T> for ListModel
+where
+    <T as ObjectSubclass>::Type: IsA<glib::Object>,
```
sdroege, line 15: As `IOStream` works without those `where` clauses, they're probably not actually needed here? I mean, the wrapper we get down there is actually a `gio::ListModel`.

gio/src/subclass/list_model.rs
```diff
@@ -0,0 +1,92 @@
...
+
+unsafe extern "C" fn list_model_get_item_type<T: ListModelImpl>(
+    list_model: *mut ffi::GListModel,
+) -> glib::ffi::GType
+where
+    <T as ObjectSubclass>::Type: IsA<glib::Object>,
+{
+    let instance = &*(list_model as *mut T::Instance);
+    let imp = instance.get_impl();
+    let model = from_glib_borrow::<_, ListModel>(list_model);
```
sdroege, line 40: For consistency with the `IOStream` code, maybe also call it `wrap` here and elsewhere :)

gio/src/subclass/list_model.rs
```diff
@@ -0,0 +1,92 @@
...
+
+    let type_ = imp.get_item_type(model.unsafe_cast_ref()).to_glib();
+
+    // Store the type so we can enforce that it doesn't change.
+    match model.get_qdata(*LIST_ITEM_TYPE_QUARK) {
+        Some(old_type) => {
+            assert_ne!(
```
sdroege, line 47:
```suggestion
            assert_eq!(
```

Or not?

sdroege (commented):
Some more minor changes but then it can go in.

Thanks also for updating the `IOStream` code!

gio/src/subclass/list_model.rs
bilelmoussaoui: The difference is `IOStream` is not an interface, so it doesn't needs the extra conditions as imposed by the `where` clauses

bilelmoussaoui:
I have addressed the remaining comments and added a bit of docs to the example, thanks for the review as always :)

gio/src/subclass/list_model.rs
sdroege: Ah of course. Makes sense :)

examples/src/bin/listbox_model.rs
```diff
@@ -19,6 +19,96 @@ use std::env::args;
...
+            const NAME: &'static str = "Model";
+            type Type = super::Model;
+            type ParentType = glib::Object;
+            type Instance = subclass::simple::InstanceStruct<Self>;
+            type Class = subclass::simple::ClassStruct<Self>;
+
+            glib_object_subclass!();
+
+            // Called right after class_init and allows a GObject to specify
+            // which interfaces it implement, in this case gio::ListModel
```
sdroege, line 46: Actually it's called even before `class_init`. On the first call to `ObjectSubclass::get_type()` :)

examples/src/bin/listbox_model.rs
```diff
@@ -19,6 +19,96 @@ use std::env::args;
...
+                .expect("Failed to create Model")
+                .downcast()
+                .expect("Created Model is of wrong type")
+        }
+
+        pub fn append(&self, obj: &RowData) {
+            let self_ = imp::Model::from_instance(self);
+            let index = {
+                // Borrow the data only once and ensure the borrow guard is dropped
+                // before we emit the items_changed signal
```
sdroege, line 94: ... because usually the view could call `get_item()` / `get_n_items()` directly from the signal handler to update its state

sdroege:
Please just merge after those two comments are updated a bit. Sorry to be a bit picky here but as this is the first example we have in gtk-rs directly that makes use of interfaces, we'll probably point people to it often so better to explain more details :)

gio/src/subclass/list_store.rs
bilelmoussaoui: > Which other widgets btw?
> 
> Also I think if you would try actually subclassing `ListBox` from Rust the way you had it before, you would likely get a critical warning at runtime because we would create a too small C instance struct (as the actual one is not known from the .gir file).

I don't think there are any currently, at least not in gtk-rs/gtk4-rs, I will have to check libhandy stuff though.

bilelmoussaoui:
It seems I can't merge stuff in gtk-rs, maybe it has to be approved first?

sdroege (approved):

sdroege:
@bilelmoussaoui Can you merge now? Otherwise @GuillaumeGomez please check permissions :)

bilelmoussaoui:
Can't seem to merge it, it says "You’re not authorized to merge this pull request."

GuillaumeGomez (approved):

GuillaumeGomez:
I was reviewing so just at the right time. :D

Thanks!

PR #194, opened 2020-12-07
gio/sys: resolve winapi reference

wonchulee:
It was broken to build gio package on windows platform.

16 |         self::winapi::shared::ws2def::AF_UNSPEC;
   |               ^^^^^^ could not find `winapi` in `self`

I checked it on windows 10.

### DIFF
--- a/gio/sys/src/manual.rs
+++ b/gio/sys/src/manual.rs
@@ -12,17 +12,15 @@ pub type GSocketMsgFlags = libc::c_int;
 
 #[cfg(target_family = "windows")]
 mod windows_constants {
-    pub const G_SOCKET_FAMILY_INVALID: super::GSocketFamily =
-        self::winapi::shared::ws2def::AF_UNSPEC;
-    pub const G_SOCKET_FAMILY_UNIX: super::GSocketFamily = self::winapi::shared::ws2def::AF_UNIX;
-    pub const G_SOCKET_FAMILY_IPV4: super::GSocketFamily = self::winapi::shared::ws2def::AF_INET;
-    pub const G_SOCKET_FAMILY_IPV6: super::GSocketFamily = self::winapi::shared::ws2def::AF_INET6;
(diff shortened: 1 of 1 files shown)


gio/sys/src/manual.rs
```diff
@@ -10,19 +10,33 @@ pub use self::windows_constants::*;
...
+    pub const G_SOCKET_FAMILY_IPV4: super::GSocketFamily = super::winapi::shared::ws2def::AF_INET;
+    pub const G_SOCKET_FAMILY_IPV6: super::GSocketFamily = super::winapi::shared::ws2def::AF_INET6;
 
     pub const G_SOCKET_MSG_NONE: super::GSocketMsgFlags = 0;
-    pub const G_SOCKET_MSG_OOB: super::GSocketMsgFlags = self::winapi::um::winsock2::MSG_OOB;
-    pub const G_SOCKET_MSG_PEEK: super::GSocketMsgFlags = self::winapi::um::winsock2::MSG_PEEK;
+    pub const G_SOCKET_MSG_OOB: super::GSocketMsgFlags = super::winapi::um::winsock2::MSG_OOB;
+    pub const G_SOCKET_MSG_PEEK: super::GSocketMsgFlags = super::winapi::um::winsock2::MSG_PEEK;
     pub const G_SOCKET_MSG_DONTROUTE: super::GSocketMsgFlags =
-        self::winapi::um::winsock2::MSG_DONTROUTE;
```
sdroege, line 25: Instead of all these changes, you can also keep the old code and replace `self::winapi::...` with `winapi::...`.

That's the only problem here :)

sdroege (changes requested):

gio/sys/src/manual.rs
wonchulee: Indeed! fixed it, thanks

sdroege (approved):
Thanks!

PR #200, opened 2020-12-07
Implement derive macro for Downgrade/Upgrade traits

andy128k:
I do not know if this is a good idea or not to use `gtk` crate in doc-tests of `glib-macros` crate. So, all examples are totally artificial and marked as `ignored`.

### DIFF
--- a/glib-macros/src/lib.rs
+++ b/glib-macros/src/lib.rs
@@ -3,6 +3,7 @@
 // Licensed under the MIT license, see the LICENSE file or <https://opensource.org/licenses/MIT>
 
 mod clone;
+mod downgrade_derive;
 mod gboxed_derive;
 mod genum_derive;
 mod gflags_attribute;
@@ -309,4 +310,69 @@ pub fn gflags(attr: TokenStream, item: TokenStream) -> TokenStream {
     let gtype_name = parse_macro_input!(attr as LitStr);
     let gen = gflags_attribute::impl_gflags(&input, &gtype_name);
     gen.into()
+}
+
+/// Macro for deriving implementations of [`glib::clone::Downgrade`] and
+/// [`glib::clone::Upgrade`] traits and a weak type.
+///
+/// # Examples
+///
+/// ## New Type Idiom
+///
+/// ```rust,ignore
+/// #[derive(glib::Downgrade)]
+/// pub struct FancyLabel(gtk::Label);
+///
+/// impl FancyLabel {
+///     pub fn new(label: &str) -> Self {
+///         Self(gtk::LabelBuilder::new().label(label).build())
+///     }
+///
+///     pub fn flip(&self) {
+///         self.0.set_angle(180.0 - self.0.get_angle());
+///     }
+/// }
+///
+/// let fancy_label = FancyLabel::new("Look at me!");
+/// let button = gtk::ButtonBuilder::new().label("Click me!").build();
+/// button.connect_clicked(clone!(@weak fancy_label => move || fancy_label.flip()));
+/// ```
+///
+/// ## Generic New Type
+///
+/// ```rust,ignore
+/// #[derive(glib::Downgrade)]
+/// pub struct TypedEntry<T>(gtk::Entry, std::marker::PhantomData<T>);
+///
+/// impl<T: ToString + FromStr> for TypedEntry<T> {
+///     // ...
+/// }
+/// ```
+///
+/// ## Structures and Enums
+///
+/// ```rust,ignore
+/// #[derive(Clone, glib::Downgrade)]
+/// pub struct ControlButtons {
+///     pub up: gtk::Button,
+///     pub down: gtk::Button,
+///     pub left: gtk::Button,
+///     pub right: gtk::Button,
+/// }
+///
+/// #[derive(Clone, glib::Downgrade)]
+/// pub enum DirectionButton {
+///     Left(gtk::Button),
+///     Right(gtk::Button),
+///     Up(gtk::Button),
+///     Down(gtk::Button),
+/// }
+/// ```
+///
+/// [`glib::clone::Downgrade`]: clone/trait.Downgrade.html
+/// [`glib::clone::Upgrade`]: clone/trait.Upgrade.html
+#[proc_macro_derive(Downgrade)]
+pub fn downgrade(input: TokenStream) -> TokenStream {
+    let input = parse_macro_input!(input as DeriveInput);
+    downgrade_derive::impl_downgrade(input)
 }
--- a/glib-macros/src/downgrade_derive/enums.rs
+++ b/glib-macros/src/downgrade_derive/enums.rs
@@ -1,0 +1,140 @@
+// Copyright 2020, The Gtk-rs Project Developers.
+// See the COPYRIGHT file at the top-level directory of this distribution.
+// Licensed under the MIT license, see the LICENSE file or <https://opensource.org/licenses/MIT>
+
+use super::fields::{derive_downgrade_fields, DowngradeStructParts};
+use proc_macro::TokenStream;
+use quote::{format_ident, quote};
+use syn::{Generics, Ident};
+
+/// This function derives a weak type for a given strong enum and
+/// implementations of `Downgrade` and `Upgrade` traits.
+///
+/// # Example
+///
+/// ```rust,ignore
+/// #[derive(glib::Downgrade)]
+/// enum Choice {
+///     This(X, Y),
+///     That { x: X, y: Y },
+/// }
+/// ```
+///
+/// Here is what will be derived:
+///
+/// ```rust,ignore
+/// enum ChoiceWeak {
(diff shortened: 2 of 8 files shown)


glib-macros/src/downgrade_enum.rs
```diff
@@ -0,0 +1,74 @@
+use crate::downgrade_fields::{derive_downgrade_fields, DowngradeStructParts};
```
sdroege, line 1: Instead of prefixing the filenames, maybe move all this into a downgrade module. E.g. `downgrade::enum`

glib-macros/src/downgrade_struct.rs
```diff
@@ -0,0 +1,44 @@
...
+    let DowngradeStructParts {
+        weak_fields,
+        end_of_struct,
+        destruct,
+        downgrade,
+        upgrade,
+    } = derive_downgrade_fields(data_struct.fields);
+
+    let derived = quote! {
+        pub struct #weak_type #generics #weak_fields #end_of_struct
```
sdroege, line 22: It's slightly ugly that we let a new type pop up in the scope as part of the derive macro but I don't think there's anything better we can do here so...

sdroege:
Looks good to me otherwise. @GuillaumeGomez @bilelmoussaoui ?

glib-macros/src/downgrade_enum.rs
GuillaumeGomez, line 1: Missing the header too. Take a look at the other files.

glib-macros/src/downgrade_struct.rs
```diff
@@ -0,0 +1,44 @@
+use crate::downgrade_fields::{derive_downgrade_fields, DowngradeStructParts};
```
GuillaumeGomez, line 1: Missing file header.

glib-macros/src/lib.rs
```diff
@@ -310,3 +313,76 @@ pub fn gflags(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+
+/// Macro for deriving implementations of [`glib::Downgrade`] and
+/// [`glib::Upgrade`] traits and a weak type.
+///
+/// # Examples
+///
+/// ## New Type Idiom
+///
+/// ```rust,ignore
+/// #[derive(Downgrade)]
```
GuillaumeGomez, line 325: Isn't it supposed to be `#[derive(glib::Downgrade)]`?

glib-macros/src/lib.rs
```diff
@@ -310,3 +313,76 @@ pub fn gflags(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+///
+/// let fancy_label = FancyLabel::new("Look at me!");
+/// let button = gtk::ButtonBuilder::new().label("Click me!").build();
+/// button.connect_clicked(clone!(@weak fancy_label => move || fancy_label.flip()));
+/// ```
+///
+/// ## Generic New Type
+///
+/// ```rust,ignore
+/// #[derive(Downgrade)]
```
GuillaumeGomez, line 346: Isn't it supposed to be `#[derive(glib::Downgrade)]`?

glib-macros/src/lib.rs
```diff
@@ -310,3 +313,76 @@ pub fn gflags(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+///
+/// impl<T: ToString + FromStr> for TypedEntry<T> {
+///     // ...
+/// }
+/// ```
+///
+/// ## Structures and Enums
+///
+/// ```rust,ignore
+/// #[derive(Clone, Downgrade)]
```
GuillaumeGomez, line 357: Isn't it supposed to be `#[derive(glib::Downgrade)]`?

glib-macros/src/lib.rs
```diff
@@ -310,3 +313,76 @@ pub fn gflags(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+/// pub struct ControlButtons {
+///     pub up: gtk::Button,
+///     pub down: gtk::Button,
+///     pub left: gtk::Button,
+///     pub right: gtk::Button,
+/// }
+///
+/// #[derive(Clone, Downgrade)]
```
GuillaumeGomez, line 365: Isn't it supposed to be `#[derive(glib::Downgrade)]`?

glib-macros/src/lib.rs
sdroege: If you do `use glib::Downgrade` first then not, but good point... let's better be explicit here

glib-macros/src/downgrade_enum.rs
```diff
@@ -0,0 +1,74 @@
...
+use proc_macro::TokenStream;
+use quote::{format_ident, quote};
+use syn::Ident;
+
+pub fn derive_downgrade_for_enum(ident: Ident, data_enum: syn::DataEnum) -> TokenStream {
```
GuillaumeGomez, line 6: Just thought about this but could you at least put some small comments on each function explaining what it's doing please?

glib-macros/src/downgrade_struct.rs
andy128k: Alternative is to generate name like [gensym](https://crates.io/crates/gensym) does. But this will make build non-reproducible. Which is, i believe, unwanted.

Anyway, type leaks through `<T as glib::clone::Downgrade>::Weak`.

sdroege: Yes that's not really an improvement :) Let's keep it like this for now

glib-macros/src/downgrade_enum.rs
andy128k: Done

glib-macros/src/downgrade_struct.rs
andy128k: Done

glib-macros/src/lib.rs
andy128k: Done

glib-macros/src/lib.rs
andy128k: Done

glib-macros/src/lib.rs
andy128k: Done

glib-macros/src/downgrade_enum.rs
andy128k: Done

glib-macros/src/lib.rs
```diff
@@ -310,3 +311,68 @@ pub fn gflags(attr: TokenStream, item: TokenStream) -> TokenStream {
     let gen = gflags_attribute::impl_gflags(&input, &gtype_name);
     gen.into()
 }
+
+/// Macro for deriving implementations of [`glib::clone::Downgrade`] and
+/// [`glib::clone::Upgrade`] traits and a weak type.
```
sdroege, line 316: Are those traits not re-exported from the crate root, i.e. `glib::Downgrade` instead of `glib::clone::Downgrade`?

sdroege: if not then maybe we should do that instead, and also include them in the prelude :)

sdroege (commented):

glib-macros/src/lib.rs
andy128k: So, `glib::clone!` should also be included in the prelude.

sdroege: Macros are always exported from the root of the crate

andy128k: Assuming `derive(Downgrade)` covers most of scenarios (he-he), users barely need to import `Downgrade/Upgrade` traits.

andy128k: But `glib::clone` is not in the prelude. It would be strange if `glib::Downgrade` derive macro will be in the prelude and `glib::clone` not.

sdroege: That's not possible with macros, they're always at the crate root

andy128k: I added macros to the prelude.

But I didn't add traits. They cause a conflict in `glib/src/subclass/object.rs`. Should I also change `subclass/object.rs`?

sdroege: > I added macros to the prelude.

What difference does that make in practice, given that macros are always exported at the crate root?

> But I didn't add traits. They cause a conflict in glib/src/subclass/object.rs. Should I also change subclass/object.rs?

What conflict does it cause?

andy128k: I causes ambiguity between `glib::object::ObjectExt::downgrade` and `glib::clone::Downgrade::downgrade`.

sdroege: I see, let's keep it the way it is then. Please also don't put the macro into the prelude, thanks!

andy128k: I reverted changes in the prelude.

sdroege:
The CI is failing, but otherwise seems good to go for me

sdroege:
Thanks!

PR #201, opened 2020-12-07
export deps from gtk

### LINKED ISSUE #143: Re-export preludes of dependent crates and dependent crates themselves
See what @sagara- wrote [here](https://github.com/gtk-rs/examples/issues/327#issuecomment-727712496):

> It might be a good idea to re-export dependent crates (like gio, gdk, atk, etc) from gtk to avoid having to explicitly specify these libraries in Cargo.toml. rust-qt apparently does this: "Each crate re-exports its dependencies, so, for example, you can access qt_core as qt_widgets::qt_core without adding an explicit dependency." (https://rust-qt.github.io/qt/getting_started/). This would then allow importing gtk::gio::prelude::* in this example without needing the explicit gio dependency in Cargo.toml. In general unless you're writing library code for one of these layers directly I don't think importing them via gtk would really be an issue.

bilelmoussaoui:
Fixes #143

### DIFF
--- a/examples/src/bin/basic_subclass.rs
+++ b/examples/src/bin/basic_subclass.rs
@@ -5,9 +5,8 @@
 //! and `activate` and how to interact with the GObjects and their private
 //! structs.
 
-use gio::prelude::*;
-use gio::subclass::prelude::*;
 use gio::ApplicationFlags;
+use gtk::{gio, glib};
 
 use glib::clone;
 use glib::subclass;
--- a/examples/src/bin/builder_basics.rs
+++ b/examples/src/bin/builder_basics.rs
@@ -1,10 +1,9 @@
 //! # Builder Basics Sample
 //!
 //! This sample demonstrates how to use the builder with an imported glade file
-use gio::prelude::*;
-use glib::clone;
-use gtk::prelude::*;
 
+use gtk::glib;
+use gtk::prelude::*;
 use gtk::{ApplicationWindow, Builder, Button, MessageDialog};
 
 use std::env::args;
@@ -25,7 +24,7 @@ fn build_ui(application: &gtk::Application) {
         gtk::Inhibit(true)
     });
 
-    bigbutton.connect_clicked(clone!(@weak dialog => move |_| dialog.show_all()));
+    bigbutton.connect_clicked(glib::clone!(@weak dialog => move |_| dialog.show_all()));
     window.show_all();
 }
 
--- a/examples/src/bin/clone_macro.rs
+++ b/examples/src/bin/clone_macro.rs
@@ -1,8 +1,7 @@
 use std::cell::RefCell;
 use std::rc::Rc;
 
-use gio::prelude::*;
-use glib::clone;
+use gtk::glib;
 use gtk::{prelude::*, Application, ApplicationWindow, Button};
 
 #[derive(Default)]
@@ -30,7 +29,7 @@ fn main() {
     {
         let state2 = Rc::new(RefCell::new(State::new()));
 
-        application.connect_activate(clone!(@weak state, @strong state2 => move |app| {
+        application.connect_activate(glib::clone!(@weak state, @strong state2 => move |app| {
             state.borrow_mut().started = true;
 
             let window = ApplicationWindow::new(app);
@@ -38,7 +37,7 @@ fn main() {
             window.set_default_size(350, 70);
 
             let button = Button::with_label("Click me!");
-            button.connect_clicked(clone!(@weak state, @weak state2 => move |_| {
+            button.connect_clicked(glib::clone!(@weak state, @weak state2 => move |_| {
                 let mut state = state.borrow_mut();
                 let mut state2 = state2.borrow_mut();
                 println!("Clicked (started: {}): {} - {}!", state.started, state.count, state2.count);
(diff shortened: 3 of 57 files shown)


gtk/src/prelude.rs
```diff
@@ -4,8 +4,18 @@
 
 //! Traits and essential types intended for blanket imports.
 
+#[doc(hidden)]
+pub use atk::prelude::*;
```
GuillaumeGomez, line 8: Just wondering: do we really want to reexport every dependencies' prelude?

sdroege: Especially the preludes, yes.

sdroege:
Looks good to me but the same should also be done for the other crates or not? E.g. `gio` should re-export `glib`, `pango` should re-export `glib`, etc.

bilelmoussaoui:
> Looks good to me but the same should also be done for the other crates or not? E.g. `gio` should re-export `glib`, `pango` should re-export `glib`, etc.

I'm not sure how useful that would be? otherwise I can do that as well

sdroege:
> I'm not sure how useful that would be? otherwise I can do that as well

The same way how it's useful from gtk :) Why is gtk different?

Also it would be inconsistent otherwise.

sdroege:
@bilelmoussaoui How should we proceed here?

bilelmoussaoui:
I will update the PR today 😊

sdroege:
@bilelmoussaoui This conflicts now and also needs to be rebased

bilelmoussaoui:
@sdroege should be ready

gdkx11/src/lib.rs
```diff
@@ -5,6 +5,7 @@
 #![allow(clippy::type_complexity)]
 
 pub use ffi;
+pub use x11;
```
sdroege, line 8: No gdk and glib here? Also this has no prelude?

sdroege:
Looks good to me otherwise

gdkx11/src/lib.rs
bilelmoussaoui: Ah it seems like I missed them on gtk4-rs as  well ^^ I have added them & removed the pango/gdk-pixbuf dependencies from gdkx11 because it doesn't need them :)

sdroege (approved):

examples/src/bin/basic_subclass.rs
```diff
@@ -4,8 +4,8 @@
 //! and showcases how you can override virtual funcitons such as `startup`
 //! and `activate` and how to interact with the GObjects and their private
 //! structs.
+use gtk::{gio, glib};
```
GuillaumeGomez, line 7: Please put back an empty line before this import.

examples/src/bin/builder_basics.rs
```diff
@@ -1,8 +1,7 @@
 //! # Builder Basics Sample
 //!
 //! This sample demonstrates how to use the builder with an imported glade file
-use gio::prelude::*;
-use glib::clone;
+use gtk::glib;
```
GuillaumeGomez, line 4: Same.

examples/src/bin/clone_macro.rs
```diff
@@ -1,8 +1,7 @@
 use std::cell::RefCell;
```
GuillaumeGomez, line 1: Seems like we didn't add a module doc comment explaining what this example was about. :-/

examples/src/bin/gio_futures_await.rs
```diff
@@ -1,4 +1,5 @@
-use gio::prelude::*;
+use gtk::prelude::*;
```
GuillaumeGomez, line 1: Same here.

examples/src/bin/list_store.rs
```diff
@@ -1,5 +1,4 @@
-use gio::prelude::*;
-use glib::clone;
+use gtk::glib;
```
GuillaumeGomez, line 1: Same here. We should maybe create an issue about that...

examples/src/bin/multi_windows.rs
```diff
@@ -1,5 +1,4 @@
-use gio::prelude::*;
-use glib::clone;
+use gtk::glib;
```
GuillaumeGomez, line 1: Same here too... :'(

examples/src/bin/multithreading_context.rs
```diff
@@ -1,4 +1,4 @@
-use gio::prelude::*;
+use gtk::glib;
```
GuillaumeGomez, line 1: And here too.... Ok, opening an issue. XD

examples/src/bin/multi_windows.rs
bilelmoussaoui: All of those are unrelated to my changes though

GuillaumeGomez (commented):
Some nits but otherwise looks good.

examples/src/bin/multi_windows.rs
GuillaumeGomez: Yup yup, don't worry. It was mostly for me, sorry for the noise.

examples/src/bin/basic_subclass.rs
bilelmoussaoui: Oups, fixed

examples/src/bin/gio_futures_await.rs
bilelmoussaoui: there's no comment to have an empty line between :p

examples/src/bin/multi_windows.rs
bilelmoussaoui: aha i misunderstood :p I thought I had to fix those as well hehe

PR #202, opened 2020-12-07
glib/signal.rs: Add docstring for SignalHandlerId

A6GibKm:
I don't know if it makes sense to put this docstring here, but since non clone structs can be a little unwieldy I think an example could help some lost soul in the future.

### DIFF
--- a/glib/src/signal.rs
+++ b/glib/src/signal.rs
@@ -13,6 +13,36 @@ use std::mem;
 use std::num::NonZeroU64;
 
 /// The id of a signal that is returned by `connect`.
+///
+/// ```ignore
+/// use glib::SignalHandlerId;
+/// use gtk::prelude::*;
+/// use std::cell::RefCell;
+///
+/// struct Button {
+///     widget: gtk::Button,
+///     clicked_handler_id: RefCell<Option<SignalHandlerId>>,
+/// }
+///
+/// impl Button {
+///     fn new() -> Self {
+///         let widget = gtk::Button::new();
+///         let clicked_handler_id = RefCell::new(Some(widget.connect_clicked(|_button| {
+///             // Do something.
+///         })));
+///         Self {
+///             widget,
+///             clicked_handler_id,
+///         }
+///     }
+///
+///     fn disconnect(&self) {
+///         if let Some(id) = self.clicked_handler_id.borrow_mut().take() {
+///             self.widget.disconnect(id)
+///         }
+///     }
+/// }
+/// ```
 #[derive(Debug, Eq, PartialEq)]
 pub struct SignalHandlerId(NonZeroU64);
 


glib/src/signal.rs
```diff
@@ -13,6 +13,36 @@ use std::mem;
 use std::num::NonZeroU64;
 
 /// The id of a signal that is returned by `connect`.
+///
+/// ```rust
+/// use glib::SignalHandlerId;
+/// use gtk::prelude::*;
```
GuillaumeGomez, line 19: You're in glib, so this won't work when running `cargo test`. Maybe try with a signal from glib?

glib/src/signal.rs
```diff
@@ -13,6 +13,36 @@ use std::mem;
 use std::num::NonZeroU64;
 
 /// The id of a signal that is returned by `connect`.
+///
+/// ```rust
```
GuillaumeGomez, line 17:
```suggestion
/// ```
```

Explanations: by default, all code blocks are rust ones. :)

GuillaumeGomez:
I think it's the right place! Just need to fix the example and then we're good to go. By the way, if you want to test it locally, you can simply use `cargo test` in `glib` folder.

glib/src/signal.rs
sdroege: There are no signals in GLib :)

glib/src/signal.rs
```diff
@@ -13,6 +13,36 @@ use std::mem;
...
 /// The id of a signal that is returned by `connect`.
+///
+/// ```no_run
+/// use glib::SignalHandlerId;
+/// use gtk::prelude::*;
+/// use std::cell::RefCell;
+///
+/// struct Button {
+///     widget: gtk::Button,
+///     connection: RefCell<Option<SignalHandlerId>>,
```
sdroege, line 24:
```suggestion
///     clicked_handler_id: RefCell<Option<SignalHandlerId>>,
```

Or something like that

glib/src/signal.rs
sdroege: So I think this has to be `no_compile` unfortunately

glib/src/signal.rs
```diff
@@ -13,6 +13,36 @@ use std::mem;
...
+///
+/// impl Button {
+///     fn new() -> Self {
+///         let widget = gtk::Button::new();
+///         let clicked_handler_id = RefCell::new(Some(widget.connect_clicked(|_button| {
+///             // Do something.
+///         })));
+///         Self {
+///             widget,
+///             connection,
```
bilelmoussaoui, line 35:
```suggestion
///             clicked_handler_id,
```

glib/src/signal.rs
```diff
@@ -13,6 +13,36 @@ use std::mem;
 use std::num::NonZeroU64;
 
 /// The id of a signal that is returned by `connect`.
+///
+/// ```no_compile
```
GuillaumeGomez, line 17:
``````suggestion
```ignore
``````

It won't be interpreted as rust code otherwise.

sdroege (approved):

bilelmoussaoui (approved):

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #205, opened 2020-12-09
glib: Add a `Value::downcast_mut()` function

ids1024:
Useful since `TypedValue::set*` take `&mut self`.

### DIFF
--- a/glib/src/value.rs
+++ b/glib/src/value.rs
@@ -186,6 +186,26 @@ impl Value {
         }
     }
 
+    /// Tries to downcast to a `&mut TypedValue`.
+    ///
+    /// Returns `Some(&mut TypedValue<T>)` if the value carries a type corresponding
+    /// to `T` and `None` otherwise.
+    pub fn downcast_mut<'a, T: FromValueOptional<'a> + SetValue>(
+        &mut self,
+    ) -> Option<&mut TypedValue<T>> {
+        unsafe {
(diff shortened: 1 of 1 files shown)


sdroege:
Makes sense, but `TypedValue` was planned to get removed as part of https://github.com/gtk-rs/gtk-rs/pull/31 . Do you actually use this in a useful way? If so, how? In my experience keeping Rust values around in the code and calling `to_value()` where needed is more convenient.

ids1024:
As far as I can tell, without downcasting to a mutable `TypedValue`, there isn't a way to safe way to set the value of an existing `Value`. (`SetValue::set_value` is an unsafe unchecked way to do this).

I am trying to add a binding for `g_settings_bind_with_mapping`, which requires a callback to set the `Value` passed as an argument. I assume there are other APIs requiring something similar.

sdroege:
You'd assign a new value, that's effectively the same.

ids1024:
Ok, assigning like that seems to work.

I like the idea having typed versions of things like this, but I guess it's reasonable enough to remove it if no one can find cases where it's especially useful. It does complicate things.

sdroege:
Yeah I like it too in principle, but I never had use for it or where I could use it, not using it was actually simpler. It only seems to complicate things compared to just passing `Value`s around where needed and elsewhere keeping everything in a proper Rust type.

PR #206, opened 2020-12-09
Macro parser improvements

sdroege:
Backport of https://github.com/gtk-rs/gtk4-rs/pull/103


gtk3-macros/src/attribute_parser.rs
```diff
@@ -0,0 +1,180 @@
+use proc_macro2::Span;
```
GuillaumeGomez, line 1: No header in this crate?

sdroege: That's how it apparently always was. So we need to fix that after this PR, and also in gtk4 (CC @bilelmoussaoui )

GuillaumeGomez: Ok, I'll let you open the issue then. ;)

GuillaumeGomez:
Thanks!

sdroege:
@GuillaumeGomez You create an issue?

gtk3-macros/src/attribute_parser.rs
sdroege: Ah you said I should, ok!

PR #210, opened 2020-12-09
gio: Add `Settings::bind_with_mapping()` binding

ids1024:
Using boxed `Fn` mainly to have a reasonable way to pass `None`.

### DIFF
--- a/gio/src/settings.rs
+++ b/gio/src/settings.rs
@@ -1,11 +1,119 @@
-use crate::{Settings, SettingsExt};
+use crate::{Settings, SettingsBindFlags, SettingsExt};
+use glib::prelude::*;
+use glib::translate::{from_glib_borrow, from_glib_none, ToGlib, ToGlibPtr};
 use glib::variant::FromVariant;
 use glib::{BoolError, IsA, ToVariant};
+
+pub struct BindingBuilder<'a> {
+    settings: &'a Settings,
+    key: &'a str,
+    object: &'a glib::Object,
+    property: &'a str,
+    flags: SettingsBindFlags,
+    get_mapping: Option<Box<dyn Fn(&glib::Variant, glib::Type) -> Option<glib::Value>>>,
+    set_mapping: Option<Box<dyn Fn(&glib::Value, glib::VariantType) -> Option<glib::Variant>>>,
+}
+
+impl<'a> BindingBuilder<'a> {
+    pub fn flags(mut self, flags: SettingsBindFlags) -> Self {
+        self.flags = flags;
+        self
+    }
+
+    pub fn get_mapping<F: Fn(&glib::Variant, glib::Type) -> Option<glib::Value> + 'static>(
+        mut self,
+        f: F,
+    ) -> Self {
+        self.get_mapping = Some(Box::new(f));
+        self
+    }
+
+    pub fn set_mapping<
+        F: Fn(&glib::Value, glib::VariantType) -> Option<glib::Variant> + 'static,
+    >(
+        mut self,
+        f: F,
+    ) -> Self {
+        self.set_mapping = Some(Box::new(f));
+        self
+    }
+
+    pub fn build(self) {
+        type Mappings = (
+            Option<Box<dyn Fn(&glib::Variant, glib::Type) -> Option<glib::Value>>>,
+            Option<Box<dyn Fn(&glib::Value, glib::VariantType) -> Option<glib::Variant>>>,
+        );
+        unsafe extern "C" fn bind_with_mapping_get_trampoline(
+            value: *mut glib::gobject_ffi::GValue,
+            variant: *mut glib::ffi::GVariant,
+            user_data: glib::ffi::gpointer,
+        ) -> glib::ffi::gboolean {
+            let user_data = &*(user_data as *const Mappings);
+            let f = user_data.0.as_ref().unwrap();
+            let value = &mut *(value as *mut glib::Value);
+            if let Some(v) = f(&*from_glib_borrow(variant), value.type_()) {
+                *value = v;
+                true
+            } else {
+                false
+            }
+            .to_glib()
+        }
+        unsafe extern "C" fn bind_with_mapping_set_trampoline(
+            value: *const glib::gobject_ffi::GValue,
+            variant_type: *const glib::ffi::GVariantType,
+            user_data: glib::ffi::gpointer,
+        ) -> *mut glib::ffi::GVariant {
+            let user_data = &*(user_data as *const Mappings);
+            let f = user_data.1.as_ref().unwrap();
+            let value = &*(value as *const glib::Value);
+            f(value, from_glib_none(variant_type)).to_glib_full()
+        }
+        unsafe extern "C" fn destroy_closure(ptr: *mut libc::c_void) {
+            Box::<Mappings>::from_raw(ptr as *mut _);
+        }
+        let get_trampoline: Option<unsafe extern "C" fn(_, _, _) -> _> =
+            if self.get_mapping.is_none() {
+                None
+            } else {
+                Some(bind_with_mapping_get_trampoline)
+            };
+        let set_trampoline: Option<unsafe extern "C" fn(_, _, _) -> _> =
+            if self.set_mapping.is_none() {
+                None
+            } else {
+                Some(bind_with_mapping_set_trampoline)
+            };
+        let mappings: Mappings = (self.get_mapping, self.set_mapping);
+        unsafe {
+            ffi::g_settings_bind_with_mapping(
+                self.settings.to_glib_none().0,
+                self.key.to_glib_none().0,
+                self.object.to_glib_none().0,
+                self.property.to_glib_none().0,
+                self.flags.to_glib(),
+                get_trampoline,
+                set_trampoline,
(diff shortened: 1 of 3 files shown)


gio/src/settings.rs
```diff
@@ -23,6 +35,68 @@ impl<O: IsA<Settings>> SettingsExtManual for O {
...
+
+    fn bind_with_mapping<P: IsA<glib::Object>>(
+        &self,
+        key: &str,
+        object: &P,
+        property: &str,
+        flags: SettingsBindFlags,
+        get_mapping: Option<Box<dyn Fn(&mut glib::Value, &glib::Variant) -> bool>>,
+        set_mapping: Option<Box<dyn Fn(&glib::Value, &glib::VariantType) -> glib::Variant>>,
+    ) {
```
sdroege, line 47: Maybe a builder here would be nicer (see `glib::Object::bind_property()`) so those optional functions only have to be provided if necessary.

gio/src/settings.rs
```diff
@@ -23,6 +35,68 @@ impl<O: IsA<Settings>> SettingsExtManual for O {
...
+            f(value, &*from_glib_borrow(variant)).to_glib()
+        }
+        unsafe extern "C" fn bind_with_mapping_set_trampoline(
+            value: *const gobject_sys::GValue,
+            variant_type: *const glib_sys::GVariantType,
+            user_data: glib_sys::gpointer,
+        ) -> *mut glib_sys::GVariant {
+            let user_data = &*(user_data as *const Mappings);
+            let f = user_data.1.as_ref().unwrap();
+            let value = &*(value as *mut glib::Value);
```
sdroege, line 69:
```suggestion
            let value = &*(value as *const glib::Value);
```

gio/src/settings.rs
```diff
@@ -23,6 +35,68 @@ impl<O: IsA<Settings>> SettingsExtManual for O {
...
         self.set_value(key, &value.to_variant())
     }
+
+    fn bind_with_mapping<P: IsA<glib::Object>>(
+        &self,
+        key: &str,
+        object: &P,
+        property: &str,
+        flags: SettingsBindFlags,
+        get_mapping: Option<Box<dyn Fn(&mut glib::Value, &glib::Variant) -> bool>>,
```
sdroege, line 45: Should this maybe be `Fn(&glib::Variant) -> Option<glib::Value>`?

gio/src/settings.rs
```diff
@@ -23,6 +35,68 @@ impl<O: IsA<Settings>> SettingsExtManual for O {
...
+        set_mapping: Option<Box<dyn Fn(&glib::Value, &glib::VariantType) -> glib::Variant>>,
```
sdroege, line 46:
```suggestion
        set_mapping: Option<Box<dyn Fn(&glib::Value, &glib::VariantType) -> Option<glib::Variant>>>,
```

This is nullable according to the docs

gio/src/settings.rs
ids1024: If the function pointer is `NULL`, gio uses the default mapping. So unless there's an explicit way to invoke that default mapping, we need to pass a `NULL` function pointer to gio to support this.

sdroege: > a new GVariant holding the data from value , or NULL in case of an error

Say the docs. So returning `NULL` instead of a variant on errors seems allowed.

sdroege: You would convert `None` into `false` and `Some(_)` into `true` and copy over the value to the pointer passed to the trampoline.

sdroege: I mean, the function would work exactly the same way but in a more Rust-y way with an optional return value instead of an out parameter and "success" boolean return value.

ids1024: Ah, I see. So it would be `get_mapping: Option<Box<dyn Fn(&glib::Variant) -> Option<glib::Value>>`? (In the absence of changing to use a builder or such.) That should be nicer to work with.

sdroege: Yeah that's what I meant

gio/src/settings.rs
ids1024: Another possible option is to split this into 3 functions, `bind_with_get_mapping()`, `bind_with_set_mapping()`, and `bind_with_get_set_mapping()`. (Both `None` should already be handled by the existing `bind()` function).

sdroege: I would remove the existing bind and add a builder (that also makes the flags optional)

gio/src/settings.rs
ids1024: Though it might be necessary to also pass a `Type`, since I think in general a mapping might work with multiple property types... If supporting that is important.

sdroege: How would that type be used and where?

ids1024: I'm guessing the default mapping is like that. It could be useful to define a mapping that could be used in several places, but wouldn't be relevant in most cases. Just worth noting.

I was wondering whether there's any issue setting the value to the wrong type, but it seems to print a helpful warning:

> GLib-GObject-WARNING **: 08:45:28.918: unable to set property 'active' of type 'gboolean' from value of type 'gchararray'

So it is probably fine to just change this to return `Option<Value>`.

ids1024: Though there's also the matter of symmetry with `get_mapping`, which takes a `glib::VariantType`.

ids1024:
Now this commit is using:

```rust
get_mapping: Option<Box<dyn Fn(&glib::Variant, glib::Type) -> Option<glib::Value>>>,
set_mapping: Option<Box<dyn Fn(&glib::Value, glib::VariantType) -> Option<glib::Variant>>>,
```

Which seems like a reasonable Rust function signature, matching the capabilities of the C API. Most implementations will probably use closures ignoring the `Type` and `VariantType` arguments, but that's easy enough.

gio/src/settings.rs
```diff
@@ -23,6 +35,74 @@ impl<O: IsA<Settings>> SettingsExtManual for O {
...
+        );
+        unsafe extern "C" fn bind_with_mapping_get_trampoline(
+            value: *mut glib::gobject_ffi::GValue,
+            variant: *mut glib::ffi::GVariant,
+            user_data: glib::ffi::gpointer,
+        ) -> glib::ffi::gboolean {
+            let user_data = &*(user_data as *const Mappings);
+            let f = user_data.0.as_ref().unwrap();
+            let value = &mut *(value as *mut glib::Value);
+            if let Some(v) = f(&*from_glib_borrow(variant), value.type_()) {
```
sdroege, line 60: So the value is initialized with the correct target type?

ids1024: Yep.

> GSettingsBindGetMapping  
> The type for the function that is used to convert from GSettings to an object property. The value is already initialized to hold values of the appropriate type.

https://developer.gnome.org/gio/stable/GSettings.html

sdroege: You could then even put a Rust panic at the end of the trampoline if you want

gio/src/settings.rs
sdroege: Looks good to me apart from this

ids1024:
Updated to use a builder pattern. Still doesn't seem like a perfect API, but I suppose I'm at least happy to not be writing `Some(Box::new(...))` when using it.

gio/src/settings.rs
```diff
@@ -1,11 +1,115 @@
...
     fn set<U: ToVariant>(&self, key: &str, value: &U) -> Result<(), BoolError>;
+
+    #[doc(alias = "g_settings_bind")]
+    #[doc(alias = "g_settings_bind_with_mapping")]
+    fn bind<'a, P: IsA<glib::Object>>(
+        &'a self,
+        key: &'a str,
+        object: &'a P,
+        property: &'a str,
+        flags: SettingsBindFlags,
```
sdroege, line 111: I would move the flags into the builder and initialize with the default flags instead of having the user always provide flags

gio/src/settings.rs
sdroege: I.e. `assert!(v.type_().is_a(value.type_()))`

sdroege:
> Updated to use a builder pattern. Still doesn't seem like a perfect API, but I suppose I'm at least happy to not be writing `Some(Box::new(...))` when using it.

Yeah writing the `Some(Box::new())` in user code is always annoying, so this is a huge usability improvement IMHO.

gio/src/settings.rs
ids1024: Good point. Done.

sdroege (approved):

gio/src/settings.rs
sdroege: Let me know if you want to do something like that or just keep it as is. If you do, should maybe do the same for the other callback too

ids1024: Hm. I guess there's a somewhat broader question of whether gtk-rs should panic in cases like this, where gtk prints a warning but keeps working.

The helpful warning glib already print seems good enough, given it can't just be enforced by the type system (now if only Rust had full dependent typing...).

sdroege: I guess it's fine if it doesn't cause things to misbehave later. Let's keep it like this for now then.

PR #211, opened 2020-12-09
Remove get  for getters & properties where applicable

fengalin:
## Note to application developers

Applications developers should use [`fix-getters-calls`](https://crates.io/crates/fix-getters-calls) to ease migration of their applications. If you also wish to change your get functions definition to comply with the standards implemented here, use [`fix-getters-def`](https://crates.io/crates/fix-getters-def).

## Object

This is an attempt at fixing https://github.com/gtk-rs/gir/issues/963.

## Dependencies

- [gir PR](https://github.com/gtk-rs/gir/pull/1021).
- [`fix-getters`](https://github.com/fengalin/fix-getters).

The last repository is a pair of tools to automate manual code getters definitions and call sites update. Please take a look at the `README`s in this repo. They explain the assumptions and decisions I had to make to achieve this. Of course, these assumption are all subject to debate. The repository also provides a crate which implements getter renaming rules. See the discussion in the [gir PR](https://github.com/gtk-rs/gir/pull/1021).

## Content

This PR consists in the following commits with different reviewing needs:

- I had to resolve conflicts in auto-generated code as a result of get prefixes removals. They can be found in [this commit](https://github.com/fengalin/gtk-rs/commit/ccdb353ffc9d4f38127043bb2079778a64603018). I believe some of them might be subject to debate.
- `gir` is updated in [this commit](https://github.com/fengalin/gtk-rs/commit/65f0f7a4f0fce94bfdc6eb2a9b054f9cc5cbbd75). Nothing special here, discussion should occur in [this PR](https://github.com/gtk-rs/gir/pull/1021).
- The `regen` produces many changes in [this commit](https://github.com/fengalin/gtk-rs/commit/414dd85c678cbab41b2c1ed86a2b96a1edc28905), but those changes result from automatic rules. I did several iterations at the changes and at the functions which were kept unchanged, but it's most possible that I overlooked something. I think agreeing on the rules explained in [`fix-getters-rules`](https://github.com/fengalin/fix-getters/tree/main/rules) and the changes to `gir` should cover most of the concerns though.
- Manual code is automatically modified using the [`fix-getters`](https://github.com/fengalin/fix-getters) tools in [this commit](https://github.com/fengalin/gtk-rs/commit/1b1622e2ab0b9cae5a72e5a75d35efd2a0cbe26d). If we agree on what is described in the `README`s, I think we should be covered for this.
- Final manual adjustments can be found in [this commit](https://github.com/fengalin/gtk-rs/commit/a588cf60ca71b77e455c8e2ce53e89c0e014ee30). There are not many of them, so it shouldn't be a problem to review.

## Base + a request

This is based on a snapshot of master and I will keep this as a basis until we reach the decision to merge this because it's quite time consuming to update. If / when we decide to merge, I would like that no branch be merged in neither `gir` nor `gtk-rs` until we get this merged completely.

I'll update `gstreamer-rs` when we come close to a merge decision.


fengalin:
Requires some manual changes I didn't catch yet...

atk/src/auto/text.rs
```diff
@@ -45,22 +45,22 @@ pub trait TextExt: 'static {
...
 
     #[doc(alias = "atk_text_get_character_at_offset")]
     fn get_character_at_offset(&self, offset: i32) -> char;
 
     #[doc(alias = "atk_text_get_character_count")]
-    fn get_character_count(&self) -> i32;
+    fn character_count(&self) -> i32;
 
     #[doc(alias = "atk_text_get_character_extents")]
     fn get_character_extents(&self, offset: i32, coords: CoordType) -> (i32, i32, i32, i32);
```
bilelmoussaoui, line 57: Any reason why those were not renamed?

fengalin: See https://github.com/fengalin/fix-getters#get-functions-selection
This is open to discussion of course.

fengalin:
Thought I should clarify: this is WIP in the sense that it can't be merged as is because:

* It depends on other projects changes / release (see initial comment).
* I wish we would discuss the premises on which I based these changes and [`fix-getters`](https://github.com/fengalin/fix-getters) in particular.

Other than that, it's ready :)

gio/Gir.toml
```diff
@@ -718,6 +717,66 @@ status = "generate"
...
+    # matches what we expect from an iterator in Rust.
+    ignore= true
+
+[[object]]
+name = "Gio.Mount"
+status = "generate"
+    [[object.function]]
+    name = "get_default_location"
+        [object.function.return]
+        nullable = false
```
sdroege, line 762: Why all the `nullable = false` here?

gio/Gir.toml
```diff
@@ -1168,13 +1241,32 @@ name = "Gio.Vfs"
 status = "generate"
 concurrency = "send+sync"
 
+[[object]]
+name = "Gio.Volume"
+status = "generate"
+    [[object.function]]
+    name = "get_icon"
+        [object.function.return]
+        nullable = false
```
sdroege, line 1250: Why all the `nullable = false` here?

glib/src/variant.rs
```diff
@@ -178,7 +179,9 @@ impl Variant {
     ///
     /// Returns `Some` if self contains a `Variant`.
     #[inline]
-    pub fn variant(&self) -> Option<Variant> {
+    // FIXME reverted auto-rename, but the difference with `variant` is not obvious
+    // and this one looks more like a getter.
+    pub fn get_variant(&self) -> Option<Variant> {
```
sdroege, line 184: It's more a `try_get_as_variant()`, just like the `str()` one should be `try_get_as_str()` or similar (and same for array, etc below).

The `toml` crate calls these kind of functions `as_foo()` so let's maybe go with that?

sdroege: Also I don't know why these even exist. `variant.get::<glib::Variant>()` should do exactly the same if it was implemented? Let's keep them for now but rename into the style `toml` uses

glib/src/object.rs
```diff
@@ -1173,8 +1167,8 @@ pub trait ObjectExt: ObjectType {
     fn is<T: StaticType>(&self) -> bool;
 
     fn get_type(&self) -> Type;
```
sdroege, line 1169: Elsewhere we call this `fn type_(&self) -> ...`

sdroege (commented):
Looks all good to me otherwise. Thanks :)

gio/Gir.toml
fengalin: I can't tell now... and it doesn't seem to make any difference. Removing.

gio/Gir.toml
fengalin: I can't tell now... and it doesn't seem to make any difference. Removing.

glib/src/object.rs
fengalin: Yeah, I agree, this is the kind of stuff for which I wanted a general consensus. I added [this remark](https://github.com/fengalin/fix-getters/blob/main/rules/README.md#general-rules) in the `fix-getters-rules` README:

> [...] other solutions would be to generate `mut_`, `loop_`, `type_` or `r#mut`, `r#loop`, `r#type`, or even a mix depending on the keyword. E.g. we could keep `get_mut`, but use the quite common `type_` and `r#async` variantes. This is something I would like to discuss.

I'll update this in `fix-getters-rules` (and thus in Gir) in one go with others suggestions from other participants.

sdroege: `get_mut()` seems fine, the others should probably get a `_` appended. @GuillaumeGomez ?

glib/src/variant.rs
fengalin: What do you think about the comment for the one just above:

```rust
    /// Boxes value.
    #[inline]
    // FIXME this one might be renamed as `new_variant`. See also below.
    pub fn variant(value: &Variant) -> Self {
        unsafe { from_glib_none(ffi::g_variant_new_variant(value.to_glib_none().0)) }
    }
```

Shall we move to `new_variant` which would follow Rust conventions better IMO? See also the changes we talked about before for the [`glib::ParamSpec::_type_`](https://github.com/fengalin/gtk-rs/blob/3f99b56b3941a5adbf67528d36d10b4b3d40fc18/glib/src/param_spec.rs#L168).

sdroege: I was talking about both. I think the correct thing to do here would be `fn from_variant(v: Variant) -> Variant` and `fn as_variant(&self) -> Option<Variant>` and the same for the other functions here.

Even more correct would be to just use the traits but they're incompletely implemented right now so that's for a later time.

sdroege: The paramspec case is a bit different because there it's actually about creating a new instance not *from* e.g. a string but *for* a string. `new_string()` or `for_string()` seems good there.

glib/src/object.rs
fengalin: Note: full list of Reserved keyword for which we could debate ;)

https://github.com/fengalin/fix-getters/blob/ab8a8cceaf45b5ba98f8030541100437b1ecdb0f/rules/src/function.rs#L16

sdroege: `get_mut()` seems to be the only special case here, and there's `std::sync::Arc::get_mut()` so...

fengalin: I think `get_optional` and `get_some` make sense too.

sdroege: True

fengalin: I double checked. The keywords in the list can't be made into a call site name. E.g., you can't call `something.loop(...)`. So, beside the cases we mentioned previously, the keyword list above should be reviewed for candidates of alternatives such as `type_` or `r#async`.

fengalin: @GuillaumeGomez: what's your opinion? Which keywords should we substitute with which convention? So far, we agreed on:

- `get_mut`, `get_option` & `get_some`: unchanged.
- `get_type` -> `type_`.

Full list available from [here](https://github.com/fengalin/fix-getters/blob/ab8a8cceaf45b5ba98f8030541100437b1ecdb0f/rules/src/function.rs#L16). Most of them could be kept unchanged, but maybe we want more substitutes:

- `get_async` -> `r#async`?
- `loop` -> `loop_`?
- ...

GuillaumeGomez: I'd go with adding `_` at the end. Simpler for users to be able to write it in their code.

glib/src/variant.rs
fengalin: Should be fixed now.

glib/src/object.rs
fengalin: Done

glib/src/variant.rs
fengalin: Argh... Clippy doesn't like `as_*`: https://github.com/gtk-rs/gtk-rs/pull/211/checks?check_run_id=2314556400#step:21:99

glib/src/object.rs
bilelmoussaoui: there's also `get_gtype` that I think should be renamed to `type_` instead of `gtype`

fengalin: Thanks. Will check this tomorrow. Too tired...

glib/src/variant.rs
sdroege: `as_*` for the getter, `from_*` for the constructor. array, tuple, maybe are constructors, that's why clippy is unhappy

fengalin: `form_*` doesn't seem right to me since it's the resulting type which is concerned by `*`. My suggestion would be these:

```rust
fn as_variant(&self) -> Option<Variant>;

fn new_as_variant(value: &Variant) -> Self;
fn new_as_array<T: StaticVariantType>(children: &[Variant]) -> Self;
fn new_as_tuple(children: &[Variant]) -> Self;
fn new_as_maybe<T: StaticVariantType>(child: Option<&Variant>) -> Self
```

Clippy is Ok with this :)

sdroege: Why `new_as`? It's creating a new `Variant` **from** an "array" (or rather slice in Rust terminogy), for example :)

fengalin: Let's take `new_as_tuple` as an example. It's also creating a variant from an `array` but I thought the point was that the resulting `Variant` was identified as being a `tuple`. Am I wrong?

sdroege: The tuple one only takes a slice currently because we don't have variadic tuples in Rust yet. Ideally it would take a `(Variant, Variant, Variant, ...)` or whatever the syntax for that would be.

fengalin: Oh, I see!

glib/src/object.rs
fengalin: @sdroege: how do you feel about `ParamSpec::new_gtype`? Is it ok to rename it to `ParamSpec::new_type`?
That's the only candidate for a `gtype` substitution I could find in `gtk-rs/gtk-rs`.

sdroege: No strong opinion either way :)

fengalin: Is it helpful to the user knowing it is a `GType`?

sdroege: Well, it's a `glib::Type`, not a `glib::GType`

fengalin: Good! switching to `ParamSpec::new_type` then. Thanks!

PR #212, opened 2020-12-12
Add new pango attributes

sophie-h:


pango/src/attribute.rs
```diff
@@ -14,96 +20,153 @@ use crate::Weight;
...
+    #[cfg(any(feature = "v1_44", feature = "dox"))]
+    #[cfg_attr(feature = "dox", doc(cfg(feature = "v1_44")))]
+    #[doc(alias = "pango_attr_allow_breaks_new")]
+    pub fn new_allow_breaks(allow_breaks: bool) -> Option<Attribute> {
+        unsafe { from_glib_full(ffi::pango_attr_allow_breaks_new(allow_breaks.to_glib())) }
+    }
+
     #[cfg(any(feature = "v1_38", feature = "dox"))]
     #[cfg_attr(feature = "dox", doc(cfg(feature = "v1_38")))]
+    #[doc(alias = "pango_attr_background_alpha_new")]
```
sdroege, line 32: @GuillaumeGomez Why didn't your script add those?

pango/src/attribute.rs
```diff
@@ -14,96 +20,153 @@ use crate::Weight;
...
+    #[cfg_attr(feature = "dox", doc(cfg(feature = "v1_46")))]
+    #[doc(alias = "pango_attr_overline_color_new")]
+    pub fn new_overline_color(red: u16, green: u16, blue: u16) -> Option<Attribute> {
+        unsafe { from_glib_full(ffi::pango_attr_overline_color_new(red, green, blue)) }
+    }
+
+    #[cfg(any(feature = "v1_46", feature = "dox"))]
+    #[cfg_attr(feature = "dox", doc(cfg(feature = "v1_46")))]
+    #[doc(alias = "pango_attr_overline_new")]
+    pub fn new_overline(overline: Overline) -> Option<Attribute> {
```
sdroege, line 103: Are all these constructors actually `nullable`? That seems strange for the majority of them

sdroege (commented):
Also it seems like the `new_` prefix of all the constructors is a bit un-Rust-y and should probably be removed.

fengalin:
> Also it seems like the `new_` prefix of all the constructors is a bit un-Rust-y and should probably be removed.

What do you mean? Most Rust constructors use `new` AFAICT :) The problem here is that we construct a variant of the type. If we'd used `overline` instead of `new_overline` that would potentially clash with an `overline` getter. I had the same problem with `glib::Param` while aligning getters with Rust styles in https://github.com/gtk-rs/gtk-rs/pull/211. See `get_flags` which was "rightfully" renamed as `flags` when there was already a `flags` constructor which lead me to rename all those with a `new` prefix, which I believe makes it clearer what the intent is.

sdroege:
@fengalin Complicated. For "variant constructors" I saw in many libraries that they only use the name of the variant without `new_` as function name but that can indeed conflict with getters if the API is confusing enough like in paramspec (a `flags` variant and all pspecs also have flags... OTOH maybe there should be no `flags` constructor but it should be `ParamSpecFlags::new()` instead).

fengalin:
Yes, the `Type::new` would probably be the way to go, or maybe a builder pattern.

sdroege:
> Yes, the `Type::new` would probably be the way to go, or maybe a builder pattern.

They're actually different types that are returned there, so builder doesn't make too much sense. The main problem is that basically all other API wants a `ParamSpec` and not a `ParamSpecInt` (etc) so user code would have to cast or the `new` function would have to return something else than `Self`. Anyway, unrelated to this PR!

fengalin:
I was talking about this PR too :) We could introduce an `AttributeBuilder` capable of building an `Attribute` with the expected variant. Ex. `AttriubuteBuilder::overline` (or longer `AttriubuteBuilder::build_overline`) wouldn't clash with an `Attribute` getter.

sdroege:
That doesn't seem like an improvement in usability though :)

fengalin:
Let's agree to disagree then :) I think we should come up with a general approach for this pattern so that it would be the same for all these situations in the organization though.

sdroege:
Let's worry about that in a separate issue, can you create one? This one can go in as-is apart from the above questions then.

pango/src/attribute.rs
GuillaumeGomez: My script is looking for `sys::` and not `ffi::`. ;)

Another run will be needed then!

GuillaumeGomez: Oh also! Would be nice to have the C equivalent for the types too.

sdroege: We don't use `sys::` anywhere since before you ran the script :P

GuillaumeGomez: Hum... Then the bug must be somewhere else.

pango/src/attribute.rs
sophie-h: I would guess that none of them are nullable. I will remove Option from all of them.

sdroege:
Thanks!

PR #214, opened 2020-12-13
glib: Hook up dispose() in ObjectImpl

YaLTeR:


glib/src/subclass/object.rs
```diff
@@ -92,6 +100,20 @@ unsafe extern "C" fn constructed<T: ObjectImpl>(obj: *mut gobject_ffi::GObject)
     imp.constructed(&from_glib_borrow::<_, Object>(obj).unsafe_cast_ref());
 }
 
+unsafe extern "C" fn dispose<T: ObjectImpl>(obj: *mut gobject_ffi::GObject) {
+    let instance = &*(obj as *mut T::Instance);
+    let imp = instance.get_impl();
+
+    imp.dispose(&from_glib_borrow::<_, Object>(obj).unsafe_cast_ref());
+
+    // Chain up to the parent's dispose.
```
sdroege, line 109: Good idea to enforce it in this case and not let the subclass take care of chaining up itself. It's easily forgotten and then causes memory leaks :)

sdroege (approved):

PR #216, opened 2020-12-14
subclass: Hook up instance_init()

YaLTeR:
Add a new wrapper type, `InitializingObject`, to store the object pointer during `instance_init()`. Binding crates are expected to define traits and implement them for `InitializingObject` to expose safe methods. Notably, this is required for `gtk_widget_init_template()`.

### DIFF
--- a/glib/src/subclass/types.rs
+++ b/glib/src/subclass/types.rs
@@ -4,9 +4,9 @@
 
 //! Module that contains the basic infrastructure for subclassing `GObject`.
 
-use crate::object::{ObjectSubclassIs, ObjectType};
+use crate::object::{Cast, ObjectSubclassIs, ObjectType};
 use crate::translate::*;
-use crate::{Closure, SignalFlags, StaticType, Type, Value};
+use crate::{Closure, Object, SignalFlags, StaticType, Type, Value};
 use std::fmt;
 use std::marker;
 use std::mem;
@@ -369,6 +369,31 @@ pub trait ObjectSubclass: Sized + 'static {
     fn with_class(_klass: &Self::Class) -> Self {
         Self::new()
     }
+
+    /// Performs additional instance initialization.
+    ///
+    /// Called just after `with_class()`. At this point the initialization has not completed yet, so
+    /// only a limited set of operations is safe (see `InitializingObject`).
+    fn instance_init(_obj: &InitializingObject<Self::Type>) {}
+}
+
+/// An object that is currently being initialized.
+///
+/// Binding crates should use traits for adding methods to this struct. Only methods explicitly safe
+/// to call during `instance_init()` should be added.
+pub struct InitializingObject<T: ObjectType>(Borrowed<T>);
+
+impl<T: ObjectType> InitializingObject<T> {
+    /// Returns a reference to the object.
+    ///
+    /// # Safety
+    ///
+    /// The returned object has not been completely initialized at this point. Use of the object
+    /// should be restricted to methods that are explicitly documented to be safe to call during
+    /// `instance_init()`.
+    pub unsafe fn as_ref(&self) -> &T {
+        &self.0
+    }
 }
 
 unsafe extern "C" fn class_init<T: ObjectSubclass>(klass: ffi::gpointer, _klass_data: ffi::gpointer)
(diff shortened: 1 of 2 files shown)


glib/src/subclass/types.rs
```diff
@@ -425,6 +450,10 @@ unsafe extern "C" fn instance_init<T: ObjectSubclass>(
     let imp = T::with_class(klass);
 
     ptr::write(imp_storage, imp);
+
+    // Any additional instance initialization.
+    let obj = from_glib_none::<_, Object>(obj.cast());
```
sdroege, line 455:
```suggestion
    let obj = from_glib_borrow::<_, Object>(obj.cast());
```

I'm surprised that this didn't cause memory leaks or crashes with GTK :) You would be stealing the floating reference here.

YaLTeR: If I make it borrow then don't I need to store `&T` in `InitializingObject`, thus requiring a lifetime generic parameter? Just wanted to avoid that.

glib/src/subclass/types.rs
```diff
@@ -369,6 +369,31 @@ pub trait ObjectSubclass: Sized + 'static {
...
+
+impl<T: ObjectType> InitializingObject<T> {
+    /// Returns a reference to the object.
+    ///
+    /// # Safety
+    ///
+    /// The returned object has not been completely initialized at this point. Use of the object
+    /// should be restricted to methods that are explicitly documented to be safe to call during
+    /// `instance_init()`.
+    pub unsafe fn as_ref(&self) -> &T {
```
sdroege, line 394: I'm a bit worried here that we provide a normal reference. Anything with that reference is potentially fatal :)

But I guess that's ok, the callers fault if they call an unsafe function!

glib/src/subclass/types.rs
```diff
@@ -369,6 +369,31 @@ pub trait ObjectSubclass: Sized + 'static {
     fn with_class(_klass: &Self::Class) -> Self {
         Self::new()
     }
+
+    /// Performs additional instance initialization.
+    ///
+    /// Called just after `with_class()`. At this point the initialization has not completed yet, so
+    /// only a limited set of operations is safe (see `InitializingObject`).
+    fn instance_init(_obj: InitializingObject<Self::Type>) {}
```
sdroege, line 377:
```suggestion
    fn instance_init(_obj: &InitializingObject<Self::Type>) {}
```

You probably don't want it to be moved. That provides even more ways of doing wrong things :)

glib/src/subclass/types.rs
YaLTeR: Yeah, that's exactly the motivation for making it `unsafe`. See also `NonNull::as_ref()` etc.

glib/src/subclass/types.rs
sdroege: No, you would store `Borrow<T>` inside `InitializingObject`.

glib/src/subclass/types.rs
sdroege: Making it return a raw pointer would make it more clear but sure :)

glib/src/subclass/types.rs
YaLTeR: Can you cast `Borrow<T>`? I need to cast it from `Object` to `Self::Type`.

sdroege: Yes, see all the signal handlers: https://github.com/gtk-rs/gtk-rs/blob/f316f4025cde40953f2e47bcd58c8f400ca92bc7/gtk/src/auto/button.rs#L733

YaLTeR: Doesn't this grab an `&Object` from within `Borrowed<Object>` and cast that into `&T`? While what I'd need is `Borrowed<Object> -> Borrowed<T>`

sdroege: You can do the cast as part of `as_ref()`. Or you can also do `Borrowed::new(Object::from_glib_borrow(ptr).into_inner().unsafe_cast())` if you prefer that.

glib/src/subclass/types.rs
YaLTeR: Hm, what for example?

sdroege: In case of `Borrow<T>` you could carry around a dangling pointer, for example.

YaLTeR: Right, good point.

YaLTeR:
Updated.

sdroege (approved):

Issue #60, opened 2020-01-02
Clean up copyright/license headers


sdroege:
See https://github.com/rust-lang/rust/pull/43498

I mentioned that a few times before but let's actually fix this now. Any opinions @GuillaumeGomez @EPashkin ? :)

PR #217, opened 2020-12-14
More checks

GuillaumeGomez:
Based on https://github.com/gtk-rs/checker/pull/16.

Fixes https://github.com/gtk-rs/gtk-rs/issues/60.

For information: I used this command to replace most license headers:

```bash
$ sed -i '/^\/\/ See the COPYRIGHT file at the top-level directory of this distribution.$/d' */src/*.rs && sed -i '/^\/\/ Copyright 20/d' */src/*.rs && sed -i 's;^// Licensed under the MIT license, see the LICENSE file or <https://opensource.org/licenses/MIT>;// Take a look at the license at the top of the repository in the LICENSE file.;' */src/*.rs && sed -i 's;// Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>;// Take a look at the license at the top of the repository in the LICENSE file.;' */src/*.rs
```


atk/src/attribute.rs
```diff
@@ -1,6 +1,4 @@
-// Copyright 2013-2018, The Gtk-rs Project Developers.
-// See the COPYRIGHT file at the top-level directory of this distribution.
-// Licensed under the MIT license, see the LICENSE file or <https://opensource.org/licenses/MIT>
+// Take a look at the license at the top of the repository in the LICENSE file.
```
sdroege, line 1: That should actually be at the top of the crate in addition too, so that it ends up in what gets uploaded to crates.io.

The text can stay the same here, but we should link the license file into each crate's root.

atk/src/attribute_set.rs
```diff
@@ -1,5 +1,3 @@
-// Copyright 2013-2018, The Gtk-rs Project Developers.
-// See the COPYRIGHT file at the top-level directory of this distribution.
-// Licensed under the MIT license, see the LICENSE file or <https://opensource.org/licenses/MIT>
+// Take a look at the license at the top of the repository in the LICENSE file.
```
sdroege, line 1: Also we need to fix up the COPYRIGHT and LICENSE file. Basically following https://github.com/rust-lang/rust/pull/43498

PR #219, opened 2020-12-15
glib: Add an `Object::new_subclass()` function

ids1024:
This is on my list of annoying boilerplate when subclassing with glib-rs.

I'm not sure this is the best way to avoid it; but I can't think of a better API. I though of making `new()` generic, but apparently default type parameters can't be used on functions.


sdroege:
Funny, I thought about adding exactly the same this morning during breakfast :)

glib/src/object.rs
```diff
@@ -1095,6 +1095,14 @@ impl Object {
         unsafe { Object::new_internal(type_, &params) }
     }
 
+    pub fn new_subclass<T: IsA<Object>>(
```
sdroege, line 1098: I'm not entirely happen with the name. It's not only for "subclasses" but it's another way of creating any `GObject`.

Maybe this can become the new `new()` and the old `new()` could become `(new_)for_type()`? I expect the one with the `T: IsA<Object>` to be the most commonly used one.

ids1024: Isn't any type that `IsA<Object>` by definition a subclass of `Object`? But I'm not particularly enamoured with the name either.

> Maybe this can become the new new() and the old new() could become (new_)for_type()? I expect the one with the T: IsA<Object> to be the most commonly used one.

That could work.

ids1024: Then there's the existing `new_generic()` function. Not sure how `Value` is more "generic" than `&dyn ToValue`...

sdroege: > Isn't any type that IsA<Object> by definition a subclass of Object? But I'm not particularly enamoured with the name either.

Yes but the way you called it, it sounds like it's only for creating subclasses defined in Rust. `ObjectSubclass` impls :)

> Then there's the existing new_generic() function. Not sure how Value is more "generic" than &dyn ToValue...

If you have better suggestions... I don't like that one either :)

ids1024:
A bit annoying to change the uses and get CI passing, but I've updated it to call this version just `Object::new()`. Which requires a gir change.

The negative diffstat is nice, at least.

glib/src/object.rs
```diff
@@ -1069,7 +1069,17 @@ glib_object_wrapper!(@object
...
 impl Object {
-    pub fn new(type_: Type, properties: &[(&str, &dyn ToValue)]) -> Result<Object, BoolError> {
+    #[allow(clippy::new_ret_no_self)]
+    pub fn new<T: IsA<Object>>(properties: &[(&str, &dyn ToValue)]) -> Result<T, BoolError> {
+        Ok(Object::new_with_type(T::static_type(), properties)?
+            .downcast()
+            .unwrap())
+    }
+
+    pub fn new_with_type(
```
bilelmoussaoui, line 1079: This should probably be named `with_type` directly

gio/src/subclass/input_stream.rs
```diff
@@ -363,10 +363,7 @@ mod tests {
 
     #[test]
     fn test_simple_stream() {
-        let stream = glib::Object::new(SimpleInputStream::static_type(), &[])
-            .unwrap()
-            .downcast::<SimpleInputStream>()
-            .unwrap();
+        let stream: SimpleInputStream = glib::Object::new(&[]).unwrap();
```
sdroege, line 366:
```suggestion
        let stream = glib::Object::new::<SimpleInputStream>(&[]).unwrap();
```

And elsewhere

sdroege:
Looks good to me otherwise. I like the new naming

glib/src/object.rs
sdroege: Should maybe call the "generic" functions something with `value` :)

gio/src/subclass/input_stream.rs
ids1024: I personally prefer this style over the turbofish, but I can change it.

sdroege: I think that's what we're using everywhere else so for consistency at least...

sdroege (approved):

sdroege:
@ids1024 Can you update to the gir version that was just merged? Then this can go in

glib/src/object.rs
sdroege: @ids1024 What do you think? Let's do that as part of a different PR though.

ids1024: I have trouble judging it because I'm not sure exactly why `new_generic` would be used (it's not used in any test or example in this repo).

`Value` implements `ToValue`. Don't know if it's useful that it takes an owned `Value` rather than an `&` reference...

If it's useful, I guess `new()` and `with_type` could both have alternatives like that, but that's kind of annoying, hard to name, and I don't know exactly why it would be used.

sdroege: > Value implements ToValue. Don't know if it's useful that it takes an owned Value rather than an & reference...

That gives you a `Value` that contains a `Value`. It's not a no-op but wraps it one more time.

> I don't know exactly why it would be used.

If you have a `Value` you need that API. For example if you just want to proxy a value you got from `get_property()` to `set_property_generic()`.

Issue #168, opened 2020-11-22
Remove `glib_` prefix from macros


sdroege:
With the 2018 edition one can use `glib::foo!` instead of `glib::glib_foo!` then, which seems nicer.

PR #220, opened 2020-12-17
Rename glib_wrapper! and glib_object_subclass! to wrapper! and object_subclass!

ids1024:
I think it works quite nicely to use `glib::wrapper!` without having to `use` anything.

Also requires a gir change.


sdroege:
There's even a issue for this: https://github.com/gtk-rs/gtk-rs/issues/168 :)

Are there any macros left that are prefixed?

ids1024:
Grepping for `'macro_rules! glib'`, there's also `glib_boxed_wrapper!`, `glib_object_interface!`, `glib_bool_error!`, `glib_result_from_gboolean!`, `glib_shared_wrapper!`, `glib_weak_impl!`, `glib_object_wrapper!`.

I guess `glib_object_interface!` is part of the public non-hidden API and all(?) of the others aren't. But they could probably all be changed.

sdroege:
> `glib_object_interface`
> `glib_bool_error`
> `glib_result_from_gboolean`

These are public and supposed to be used. Maybe the others could be prefixed with a couple of underscores or so to make it clearer that they're not "public" (well they are, but you're not supposed to use them directly...)

sdroege:
Ok :) Do you want to prefix the other macros with `__` or keep it like this?

ids1024:
I think a common idiom for this sort of thing would be a `#[doc(hidden)]` module named something like `private`. So it could be `glib::private::boxed_wrapper!`.

But it should be fine to merge this without a change like that.

sdroege:
Yeah, fine for me to merge like this too. I just wasn't sure if you would like to do that change or not before getting it merged :)

sdroege (approved):

ids1024:
Updated gir, and all checks are now passing.

GuillaumeGomez (approved):
Thanks!

PR #223, opened 2020-12-19
docs: update per the new macro names

bilelmoussaoui:
follow up of #220


GuillaumeGomez (approved):

glib/src/object.rs
```diff
@@ -2554,7 +2554,7 @@ pub unsafe trait ParentClassIs: ObjectType {
 }
 
 /// Automatically implemented by `ObjectSubclass` variants of
-/// [`wrapper!`][crate::glib_wrapper]
+/// [`wrapper!`][crate::wrapper]
```
bilelmoussaoui, line 2557: Hmm how should I handle such complaint? maybe I should use `crate::macro::wrapper`?

GuillaumeGomez: Add a `!` at the end. :)

PR #228, opened 2020-12-20
Consistently re-export subclassing preludes of dependency crates

sdroege:
@bilelmoussaoui Same change is also needed for gtk4 I guess

### DIFF
--- a/gtk/src/subclass/mod.rs
+++ b/gtk/src/subclass/mod.rs
@@ -33,6 +33,11 @@ pub mod widget;
 pub mod window;
 
 pub mod prelude {
+    #[doc(hidden)]
+    pub use gio::subclass::prelude::*;
+    #[doc(hidden)]
+    pub use glib::subclass::prelude::*;
+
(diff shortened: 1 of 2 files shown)


gtk/src/subclass/mod.rs
```diff
@@ -33,6 +33,11 @@ pub mod widget;
 pub mod window;
 
 pub mod prelude {
+    #[doc(hidden)]
+    pub use gio::subclass::prelude::*;
+    #[doc(hidden)]
+    pub use glib::subclass::prelude::*;
```
GuillaumeGomez, line 39: What does it allow to do btw?

sdroege: Not having to import all the subclass preludes every time. Same as what was done for the normal preludes.

GuillaumeGomez: I never used subclass so I really can't judge here. I'll trust you here. :)

bilelmoussaoui (approved):

PR #230, opened 2020-12-20
update bitflags to 1.2 to fix macro namespace issues

bvinc:
Update bitflags from 1.0 to 1.2.

I was working with a project that was pulling in glib from github, and it kept failing because an bitflags-internal macro `__bitflags` was not in scope.  I'm not exactly sure why it's not always triggered, but it's something to do with 2018 edition and macro scopes.  It was fixed in bitflags PR 165.

It appears 1.2 has the same minimum rust version.  I don't see any reason not to update it.

### DIFF
--- a/atk/Cargo.toml
+++ b/atk/Cargo.toml
@@ -37,7 +37,7 @@ git = "https://github.com/gtk-rs/lgpl-docs"
 
 [dependencies]
 libc = "0.2"
-bitflags = "1.0"
+bitflags = "1.2"
 ffi = { package = "atk-sys", path = "sys" }
 glib = { path = "../glib" }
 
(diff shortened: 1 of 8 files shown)


GuillaumeGomez (approved):

sdroege:
I'm not sure what the point of this is. `bitflags = "1.0"` would pull in bitflags 1.2 or newer if available and not prevented by another dependency.

Is the problem that bitflags < 1.2 does not work at all with gtk-rs anymore?

bvinc:
You're right.  I learned something new about semver after 1.0.  I'll close this.

sdroege:
It also seems that building glib with bitflags 1.0.4 at least works fine still.

PR #231, opened 2020-12-21
Minor type / value improvements

sdroege:


glib/src/types.rs
```diff
@@ -201,13 +201,121 @@ macro_rules! builtin {
     };
 }
 
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct ILong(pub libc::c_long);
```
GuillaumeGomez, line 205: Maybe a stupid question but why not writing a macro for those implementations so you don't repeat code?

sdroege: Didn't seem worth it for this. There are only ever 2 types like this.

GuillaumeGomez: If there are only two, not really worth it, indeed.

GuillaumeGomez (approved):

Issue #195, opened 2020-12-07
glib::Date API is unsound and/or asserts


sdroege:
For example
- All the constructors assert if invalid arguments are provided
- Add/subtract asserts if overflows happen
- Various functions assert if an invalid date is passed in (as can be constructed by setting e.g. to February 31`)
- `clear()` overwrites memory with zeroes without any checks that the memory is actually accessible

PR #232, opened 2020-12-22
Add checks in Date methods

GuillaumeGomez:
Part of #195.

### DIFF
--- a/glib/src/date.rs
+++ b/glib/src/date.rs
@@ -1,11 +1,11 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
 use crate::translate::*;
+use crate::BoolError;
 use crate::DateDay;
 use crate::DateMonth;
 use crate::DateWeekday;
 use crate::DateYear;
-use crate::Time;
 use std::cmp;
 use std::fmt;
 use std::hash;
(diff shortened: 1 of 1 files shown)


glib/src/date.rs
```diff
@@ -130,73 +152,125 @@ impl Date {
...
+                ffi::g_date_subtract_days(self.to_glib_none_mut().0, n_days);
+            }
+            Ok(())
         }
     }
 
-    pub fn subtract_months(&mut self, n_months: u32) {
+    pub fn subtract_months(&mut self, n_months: u32) -> Result<(), BoolError> {
+        // The checks for this function are just a mess in the C code, allowing intermediate
+        // unknonwn state. So for now, nothing can be done...
```
sdroege, line 259:
```suggestion
        // unknown state. So for now, nothing can be done...
```

`subtract_days()` should be much worse then, but you don't have such a comment there :)

sdroege:
I guess this is better than nothing...

glib/src/date.rs
GuillaumeGomez: Hum no, `subtract_days` is easy to check in comparison: https://gitlab.gnome.org/GNOME/glib/-/blob/master/glib/gdate.c#L1709

sdroege: Ah because it goes through julian days...

sdroege (approved):

Issue #73, opened 2020-06-02
No GTask wrappers?


shutton:
I notice that there's no wrapper code for GTask.  I was just wondering if that was a conscious decision, or simply work that hadn't been done yet, since it seems like a rather glaring omission.

GTask is (or could be) quite helpful in exporting Rust-based implementations to existing C/GObject-based environments that are already designed around async/finish APIs.

PR #245, opened 2021-01-04
Add bindings for GTask

mbiggio:
This addresses https://github.com/gtk-rs/gtk-rs/issues/73


sdroege:
> This branch has conflicts that must be resolved

In case you missed that :)

gio/src/auto/data_input_stream.rs
```diff
@@ -122,8 +124,11 @@ pub trait DataInputStreamExt: 'static {
     #[doc(alias = "g_data_input_stream_read_int64")]
     fn read_int64<P: IsA<Cancellable>>(&self, cancellable: Option<&P>) -> Result<i64, glib::Error>;
 
-    //#[doc(alias = "g_data_input_stream_read_line_finish_utf8")]
-    //fn read_line_finish_utf8(&self, result: /*Ignored*/&AsyncResult) -> Result<(Option<glib::GString>, usize), glib::Error>;
+    #[doc(alias = "g_data_input_stream_read_line_finish_utf8")]
+    fn read_line_finish_utf8<P: IsA<AsyncResult>>(
```
sdroege, line 128: Needs a change to gir so that all the `finish()` functions are not generated. They're already handled as part of the `async()` functions

gio/src/auto/task.rs
```diff
@@ -0,0 +1,297 @@
...
+
+    #[cfg(any(feature = "v2_60", feature = "dox"))]
+    #[cfg_attr(feature = "dox", doc(cfg(feature = "v2_60")))]
+    #[doc(alias = "g_task_get_name")]
+    pub fn get_name(&self) -> Option<glib::GString> {
+        unsafe { from_glib_none(ffi::g_task_get_name(self.to_glib_none().0)) }
+    }
+
+    #[doc(alias = "g_task_get_priority")]
+    pub fn get_priority(&self) -> i32 {
```
sdroege, line 76:
```suggestion
    pub fn get_priority(&self) -> glib::source::Priority {
```

gio/src/auto/task.rs
```diff
@@ -0,0 +1,297 @@
...
+    #[cfg_attr(feature = "dox", doc(cfg(feature = "v2_60")))]
+    #[doc(alias = "g_task_set_name")]
+    pub fn set_name(&self, name: Option<&str>) {
+        unsafe {
+            ffi::g_task_set_name(self.to_glib_none().0, name.to_glib_none().0);
+        }
+    }
+
+    #[doc(alias = "g_task_set_priority")]
+    pub fn set_priority(&self, priority: i32) {
```
sdroege, line 216:
```suggestion
    pub fn set_priority(&self, priority: glib::source::Priority) {
```

gio/src/task.rs
```diff
@@ -0,0 +1,121 @@
+use crate::AsyncResult;
+use crate::Cancellable;
+use crate::Task;
+use glib::object::IsA;
+use glib::translate::*;
+use std::boxed::Box as Box_;
+
+impl Task {
+    pub fn new<P: IsA<Cancellable>, Q: FnOnce(&AsyncResult, Option<&glib::Object>) + 'static>(
```
sdroege, line 9: This looks like it should also get a futures variant that can be awaited on

sdroege (commented):
Can you also put a meaningful example for this into the examples repo, that shows the API works correctly and what it can be used for?

gio/src/auto/data_input_stream.rs
mbiggio: Sure, I will add rules for those functions in the Gir

sdroege: I meant in https://github.com/gtk-rs/gir , not in the `Gir.toml` :)

gio/src/task.rs
mbiggio: Well, the idea is more to instantiate a `Task` object and pass it to async functions/futures. When the future will complete, it will call `Task::return_int` or similar and this will trigger the callback associated to the `Task`. Inside the callback typically there will be a call to `Task::propagate_int()` to get the result of the task.
All this is just to be able to implement "old-style" APIs using the async/finish machinery (which requires GTasks and GAsyncResults) fully in Rust.

Is this what you meant / does this address your point? :)

sdroege: Can you provide an example of how all that would work in the `examples` subdirectory? :)

mbiggio: sure, I'm adding the example, I will soon update the PR with it

examples/src/bin/gio_task.rs
```diff
@@ -0,0 +1,188 @@
...
+
+pub mod imp {
+    use super::*;
+
+    // FileSize is a simple object that will just contain the read file size.
+    // Initially the size field will be initialized to 0.
+    pub type FileSize = <FileSizePrivate as ObjectSubclass>::Instance;
+    pub type FileSizeClass = <FileSizePrivate as ObjectSubclass>::Class;
+    pub struct FileSizePrivate {
+        pub size: std::cell::RefCell<i64>,
```
sdroege, line 25:
```suggestion
        pub size: std::cell::RefCell<Option<u64>>,
```

Maybe?

examples/src/bin/gio_task.rs
```diff
@@ -0,0 +1,188 @@
+//! # GTask Sample
+//!
+//! This sample demonstrates how to implement a GIO async/finish API based on GTask (see
+//! https://developer.gnome.org/gio/stable/GTask.html) in Rust using the GTask generated bindings.
+//! This can be useful, for example, when porting to Rust some existing C code exposing such an API.
+
+use glib::subclass;
+use glib::subclass::prelude::*;
+use glib::translate::ToGlibPtr;
+use gtk::glib::translate::FromGlibPtrNone;
```
sdroege, line 10:
```suggestion
use gtk::glib::translate::*;
```

examples/src/bin/gio_task.rs
```diff
@@ -0,0 +1,188 @@
+//! # GTask Sample
+//!
+//! This sample demonstrates how to implement a GIO async/finish API based on GTask (see
+//! https://developer.gnome.org/gio/stable/GTask.html) in Rust using the GTask generated bindings.
+//! This can be useful, for example, when porting to Rust some existing C code exposing such an API.
+
+use glib::subclass;
+use glib::subclass::prelude::*;
```
sdroege, line 8:
```suggestion
use gtk::glib::subclass;
use gtk::glib::subclass::prelude::*;
```

examples/src/bin/gio_task.rs
sdroege: And then also update the comment above ("will be initialized to 0")

examples/src/bin/gio_task.rs
```diff
@@ -0,0 +1,188 @@
...
+    impl ObjectImpl for FileSizePrivate {}
+
+    /// # Safety
+    ///
+    /// This is the ffi method to asynchronously get the file size. It accepts a callback of type
+    /// GAsyncReadyCallback, that will be invoked when the async operation completes. Typically,
+    /// this callback will invoke the get_file_size_finish method (implemented below) to get the
+    /// Task result and perform some operation with it.
+    #[no_mangle]
+    pub unsafe extern "C" fn get_file_size_async(
```
sdroege, line 52: I would move these into an `ffi` module and have them only do FFI: all actual functionality should be done via safe (public) Rust API that is outside the `imp` module.

examples/src/bin/gio_task.rs
```diff
@@ -0,0 +1,188 @@
...
+                .upcast_ref::<gio::AsyncResult>()
+                .get_source_object()
+                .unwrap();
+
+            let source_object = FileSizePrivate::from_instance(
+                &source_object.downcast_ref::<super::FileSize>().unwrap(),
+            );
+
+            source_object.size.replace(size);
+            task.return_int(size.try_into().unwrap());
```
sdroege, line 89: Maybe return as `glib::Value` so you can have a proper `i64` or `u64` instead of having to work with `isize` and failing on 32 bit platforms if the file is bigger than 2GB

gio/src/auto/async_result.rs
```diff
@@ -0,0 +1,71 @@
...
+impl<O: IsA<AsyncResult>> AsyncResultExt for O {
+    fn get_source_object(&self) -> Option<glib::Object> {
+        unsafe {
+            from_glib_full(ffi::g_async_result_get_source_object(
+                self.as_ref().to_glib_none().0,
+            ))
+        }
+    }
+
+    //fn get_user_data(&self) -> /*Unimplemented*/Option<Fundamental: Pointer> {
```
sdroege, line 43: How would one do this in bindings usually? A subclass of `GAsyncResult`?

gio/Gir.toml
```diff
@@ -408,6 +408,11 @@ generate_builder = true
     #return vec of u8
     manual = true
 
+    [[object.function]]
+    name = "read_line_finish_utf8"
+    #finish functions not needed
+    manual = true
```
sdroege, line 414: This needs to be done in https://github.com/gtk-rs/gir/ directly and automatically

gio/src/async_result.rs
```diff
@@ -0,0 +1 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
```
sdroege, line 1: This file is basically empty

gio/src/auto/task.rs
```diff
@@ -0,0 +1,243 @@
...
+        unsafe { from_glib(ffi::g_task_get_return_on_cancel(self.to_glib_none().0)) }
+    }
+
+    //#[doc(alias = "g_task_get_source_tag")]
+    //pub fn get_source_tag(&self) -> /*Unimplemented*/Option<Fundamental: Pointer> {
+    //    unsafe { TODO: call ffi:g_task_get_source_tag() }
+    //}
+
+    //#[doc(alias = "g_task_get_task_data")]
+    //pub fn get_task_data(&self) -> /*Unimplemented*/Option<Fundamental: Pointer> {
```
sdroege, line 86: Same question for this: a `GTask` subclass would be used here in case of bindings to pass around custom data?

gio/Gir.toml
```diff
@@ -997,6 +1002,34 @@ status = "generate"
...
+    #wrong return value
+    ignore = true
+    [[object.function]]
+    name = "return_value"
+    #wrong mutability
+    ignore = true
+    [[object.function]]
+    name = "propagate_value"
+    #wrong mutability
+    ignore = true
```
sdroege, line 1031: `propagate_value()` and `return_value` seem rather important for proper usage from bindings so just implement them manually please :)

gio/Gir.toml
```diff
@@ -997,6 +1002,34 @@ status = "generate"
...
+    name = "get_priority"
+    #wrong priority type
+    ignore = true
+    [[object.function]]
+    name = "set_priority"
+    #wrong priority type
+    ignore = true
+    [[object.function]]
+    name = "propagate_boolean"
+    #wrong return value
```
sdroege, line 1022: This should probably also be manually implemented

examples/src/bin/gio_task.rs
```diff
@@ -0,0 +1,188 @@
...
+use std::convert::TryInto;
+
+pub mod imp {
+    use super::*;
+
+    // FileSize is a simple object that will just contain the read file size.
+    // Initially the size field will be initialized to 0.
+    pub type FileSize = <FileSizePrivate as ObjectSubclass>::Instance;
+    pub type FileSizeClass = <FileSizePrivate as ObjectSubclass>::Class;
+    pub struct FileSizePrivate {
```
sdroege, line 24:
```suggestion
    pub struct FileSize {
```

We have namespaces in Rust :)

gio/Gir.toml
pbor: Do you have pointers on where to start? the gir program is not small :)

sdroege: I'd search for `_async` and `_finish` as strings. That should lead you to the code where the gio-style async functions are handled.

Somewhere in `src/analysis` you should be able to mark the `finish()` functions as ignored then.

gio/Gir.toml
mbiggio: This is already manually implemented in `gio/src/task.rs` (actually ... because it was the easiest to do :)) I will implement `propagate_value` and `return_value` as well

gio/src/async_result.rs
mbiggio: will remove

gio/src/auto/async_result.rs
mbiggio: In Rust bindings, associating some data to the Task callback can be done with a closure with the `user_data` moved into its scope, without the need to explicitly pass a `user_data` pointer to the Task constructor. I will add this to the `gio_task.rs` example

gio/src/auto/task.rs
mbiggio: In this case probably an option would be to subclass `Task` indeed

gio/Gir.toml
mbiggio: I left this modification to the `Gir.toml` for now in this new commit to check if it fixes the build, but I did a parallel PR to gir: https://github.com/gtk-rs/gir/pull/1031

I will remove this once the gir sub-repo is updated

gio/Gir.toml
mbiggio: I added the `value` ones and removed the `propagate_int()` and `return_int` and the like. I did that because looking at [task.c](https://github.com/frida/glib/blob/master/gio/gtask.c#L589), the result is implemented as a `union`, and nothing would prevent calling `return` with one type and after `propagate` with another type (for example `return_int` and `propagate_object`), and this would in general cause undefined behavior. Sticking to `Values` would probably cause some overhead but I think is safer. I'm open to suggestions of course :)

sdroege: Makes sense

examples/src/bin/gio_task.rs
```diff
@@ -0,0 +1,192 @@
...
+    pub type FileSize = <imp::FileSize as ObjectSubclass>::Instance;
+
+    /// # Safety
+    ///
+    /// This is the ffi method to asynchronously get the file size. It accepts a callback of type
+    /// GAsyncReadyCallback, that will be invoked when the async operation completes. Typically,
+    /// this callback will invoke the get_file_size_finish method (implemented below) to get the
+    /// Task result and perform some operation with it.
+    #[no_mangle]
+    pub unsafe extern "C" fn get_file_size_async(
```
sdroege, line 54: Maybe worth adding some namespacing on the symbol name here lik in C. `my_file_size_get_file_size_async()` or so?

examples/src/bin/gio_task.rs
```diff
@@ -0,0 +1,192 @@
...
+        this: *mut FileSize,
+        cancellable: *mut gio::ffi::GCancellable,
+        callback: gio::ffi::GAsyncReadyCallback,
+        user_data: glib::ffi::gpointer,
+    ) {
+        let cancellable = gio::Cancellable::from_glib_none(cancellable);
```
sdroege, line 60:
```suggestion
        let cancellable = gio::Cancellable::from_glib_borrow(cancellable);
```

examples/src/bin/gio_task.rs
```diff
@@ -0,0 +1,192 @@
...
+        let closure = move |result: &gio::AsyncResult, source_object: Option<&glib::Object>| {
+            let result: *mut gio::ffi::GAsyncResult = result.to_glib_none().0;
+            let source_object: *mut glib::object::GObject = source_object.to_glib_none().0;
+            callback.unwrap()(source_object, result, user_data)
+        };
+
+        let source_object = &super::FileSize::from_glib_none(this);
```
sdroege, line 67:
```suggestion
        let source_object = &super::FileSize::from_glib_borrow(this);
```

examples/src/bin/gio_task.rs
```diff
@@ -0,0 +1,192 @@
...
+        let task = gio::Task::new(
+            Some(&source_object.clone().upcast::<glib::Object>()),
```
sdroege, line 69:
```suggestion
            Some(&source_object.upcast_ref::<glib::Object>()),
```

examples/src/bin/gio_task.rs
```diff
@@ -0,0 +1,192 @@
...
+            task.return_value(&size.to_value());
+        });
+    }
+
+    /// # Safety
+    ///
+    /// This method will typically be invoked in the callback passed to get_file_size_async in order
+    /// to propagate the Task result.
+    #[no_mangle]
+    pub unsafe extern "C" fn get_file_size_finish(
```
sdroege, line 100: symbol name as above

examples/src/bin/gio_task.rs
```diff
@@ -0,0 +1,192 @@
...
+        _this: *mut FileSize,
+        result: *mut gio::ffi::GAsyncResult,
+        _error: *mut *mut glib::ffi::GError,
+    ) -> i64 {
+        gio::AsyncResult::from_glib_none(result)
```
sdroege, line 105:
```suggestion
        gio::AsyncResult::from_glib_borrow(result)
```

examples/src/bin/gio_task.rs
```diff
@@ -0,0 +1,192 @@
...
+            .unwrap()
+            .unwrap()
+    }
+
+    /// # Safety
+    ///
+    /// Simple getter
+    #[no_mangle]
+    pub unsafe extern "C" fn get_retrieved_size(this: *mut FileSize) -> i64 {
+        let simple_object = super::FileSize::from_glib_none(this);
```
sdroege, line 120:
```suggestion
        let simple_object = super::FileSize::from_glib_borrow(this);
```

examples/src/bin/gio_task.rs
```diff
@@ -0,0 +1,192 @@
...
+            .get::<i64>()
+            .unwrap()
+            .unwrap()
+    }
+
+    /// # Safety
+    ///
+    /// Simple getter
+    #[no_mangle]
+    pub unsafe extern "C" fn get_retrieved_size(this: *mut FileSize) -> i64 {
```
sdroege, line 119: symbol name as above

examples/src/bin/gio_task.rs
```diff
@@ -0,0 +1,192 @@
...
+}
+
+glib::wrapper! {
+    pub struct FileSize(ObjectSubclass<imp::FileSize>);
+}
+
+impl FileSize {
+    pub fn new() -> Self {
+        glib::Object::new(&[]).expect("Failed to create FileSize")
+    }
```
sdroege, line 135: I think for consistency the proper API from the `extern "C"` functions above should be available here too. With the same scheme as in the bindings (but not necessarily calling the FFI functions).

gio/src/auto/task.rs
```diff
@@ -0,0 +1,205 @@
...
+    pub struct Task(Object<ffi::GTask, ffi::GTaskClass>) @implements AsyncResult;
+
+    match fn {
+        get_type => || ffi::g_task_get_type(),
+    }
+}
+
+impl Task {
+    //#[doc(alias = "g_task_new")]
+    //pub fn new<P: IsA<Cancellable>, Q: FnOnce(Result<(), glib::Error>) + 'static>(source_object: Option<&glib::Object>, cancellable: Option<&P>, callback: Q) -> Task {
```
sdroege, line 36: Maybe mark as manual in `Gir.toml`

gio/src/test_util.rs
```diff
@@ -23,3 +23,22 @@ pub fn run_async<T: Send + 'static, Q: FnOnce(Sender<T>, MainLoop) + Send + 'sta
 
     rx.recv().unwrap()
 }
+
+#[allow(dead_code)]
+pub fn run_async_local<T: 'static, Q: FnOnce(Sender<T>, MainLoop) + Send + 'static>(start: Q) -> T {
```
sdroege, line 28: What's with this? :)

glib/src/value.rs
```diff
@@ -394,7 +394,7 @@ impl<'a> ToGlibContainerFromSlice<'a, *mut gobject_ffi::GValue> for &'a Value {
         }
 
         unsafe {
-            let res = ffi::g_malloc(mem::size_of::<gobject_ffi::GValue>() * t.len())
+            let res = ffi::g_malloc0(mem::size_of::<gobject_ffi::GValue>() * t.len())
```
sdroege, line 397: Oops. Thanks!

gio/src/test_util.rs
mbiggio: I did that because using `invoke` i would need `Value` to implement `Send`, but it does not since the raw pointer to `GValue` is not `Send`:
```
error[E0277]: `*mut c_void` cannot be sent between threads safely
   --> gio/src/task.rs:228:15
    |
228 |         match run_async(|tx, l| {
    |               ^^^^^^^^^ `*mut c_void` cannot be sent between threads safely
    | 
   ::: gio/src/test_util.rs:7:21
    |
7   | pub fn run_async<T: Send + 'static, Q: FnOnce(Sender<T>, MainLoop) + Send + 'static>(
    |                     ---- required by this bound in `run_async`
    |
    = help: within `std::result::Result<Value, glib::Error>`, the trait `Send` is not implemented for `*mut c_void`
    = note: required because it appears within the type `GValue_data`
    = note: required because it appears within the type `[GValue_data; 2]`
    = note: required because it appears within the type `GValue`
    = note: required because it appears within the type `Value`
    = note: required because it appears within the type `std::result::Result<Value, glib::Error>`

```

sdroege: Makes sense

examples/src/bin/gio_task.rs
```diff
@@ -0,0 +1,263 @@
...
+    /// # Safety
+    ///
+    /// Simple getter
+    #[no_mangle]
+    pub unsafe extern "C" fn my_file_size_get_retrieved_size(this: *mut FileSize) -> i64 {
+        let simple_object = super::FileSize::from_glib_borrow(this);
+        let simple_object =
+            imp::FileSize::from_instance(&simple_object.downcast_ref::<super::FileSize>().unwrap());
+        let x = *simple_object.size.borrow();
+        x.unwrap()
```
sdroege, line 124:
```suggestion
        x.unwrap_or(-1)
```

Maybe?

examples/src/bin/gio_task.rs
```diff
@@ -0,0 +1,263 @@
...
+glib::wrapper! {
+    pub struct FileSize(ObjectSubclass<imp::FileSize>);
+}
+
+impl FileSize {
+    pub fn new() -> Self {
+        glib::Object::new(&[]).expect("Failed to create FileSize")
+    }
+
+    fn retrieved_size(&self) -> i64 {
```
sdroege, line 137:
```suggestion
    fn retrieved_size(&self) -> Option<i64> {
```

Maybe?

examples/src/bin/gio_task.rs
```diff
@@ -0,0 +1,263 @@
...
+    fn default() -> Self {
+        FileSize::new()
+    }
+}
+
+// This function mimicks what the C code using the exported async/finish API would do. It first defines a
+// callback of type GAsyncResult that internally calls get_file_size_finish to retrieve the Task
+// result, and simply prints it out. Then it invokes the my_file_size_get_file_size_async method, passing the
+// callback to it as parameter.
+async fn run_unsafe() {
```
sdroege, line 198: You're not `await`ing anywhere, so this can as well be a non-async function. It will return immediately and then asynchronously call `c_callback` at some point.

examples/src/bin/gio_task.rs
```diff
@@ -0,0 +1,263 @@
...
+            cancellable.to_glib_none().0,
+            Some(c_callback),
+            std::ptr::null_mut(),
+        );
+    }
+}
+
+// This function is the "safe" counterpart of "run_unsafe", using only safe rust
+// bindings to accomplish the same task
+async fn run_safe() {
```
sdroege, line 239: Same here but here you could relatively easily make it actually async by passing a oneshot channel into the callback and awaiting on that

examples/src/bin/gio_task.rs
```diff
@@ -0,0 +1,263 @@
...
+}
+
+fn main() {
+    let c = glib::MainContext::default();
+    let l = glib::MainLoop::new(Some(&c), false);
+
+    c.push_thread_default();
+    c.spawn_local(run_unsafe());
+    c.spawn_local(run_safe());
+    l.run();
```
sdroege, line 261: Would be nice to make it so that the mainloop quits once both callbacks were called

sdroege (approved):

mbiggio:
Thanks @sdroege for your time on this, much appreciated :)

PR #246, opened 2021-01-04
object: Add a From<&mut $ffi_type>

cgwalters:
The use case for this is a bit obscure.  In rpm-ostree
we're combining glib-rs with https://cxx.rs - the latter supports
passing "external binding types" across its own FFI, such as types
created by `bindgen`.  It works equally well for glib-rs generated
types.

However, rather than passing raw pointers it provides `Pin<&mut $ffitype>`.
Add a `From<>` implementation that makes it convenient to go
from a reference of the FFI type to the wrapper.  This will allow
us to drop some custom wrapper glue defined per GObject type
in rpm-ostree.

### DIFF
--- a/glib/src/object.rs
+++ b/glib/src/object.rs
@@ -658,6 +658,20 @@ macro_rules! glib_object_wrapper {
             type GlibType = *mut $ffi_name;
         }
 
+        // This method isn't something most "pure Rust" users would need, so hide
+        // it from the docs.  One motivational use is combining gtk-rs with cxx-rs;
+        // the latter provides `Pin<&mut $ffitype>` rather than raw pointers.
+        #[doc(hidden)]
+        impl From<&mut $ffi_name> for $name {
+            fn from(r: &mut $ffi_name) -> $name {
+                // Safety: We are incrementing the refcount, so converting the
+                // reference into a raw pointer is OK.
+                unsafe {
+                    $name($crate::translate::from_glib_none(r as *mut $ffi_name as *mut $crate::object::GObject))
+                }
+            }
+        }
+
         #[doc(hidden)]
         unsafe impl $crate::object::ObjectType for $name {
             type GlibType = $ffi_name;


cgwalters:
xref thread in https://github.com/coreos/rpm-ostree/pull/2416#discussion_r551185413

cgwalters:
Example code:

```
use gio::prelude::*;

fn main() {
    let p = unsafe {
        let s = std::ffi::CString::new(".").expect("cstr");
        gio_sys::g_file_new_for_path(s.as_ptr())
    };
    let r = unsafe { &mut *p };
    let f: gio::File = r.into();
    println!("{}", f.get_uri());
}
```

This could probably go in a test...somewhere?  I guess in gio, since there's no actual concrete GObject types we can instantiate in the gobject create?
This

sdroege:
> This could probably go in a test...somewhere? I guess in gio, since there's no actual concrete GObject types we can instantiate in the gobject crate?

Sounds like a good plan

> However, rather than passing raw pointers it provides `Pin<&mut $ffitype>`.
> Add a `From<>` implementation that makes it convenient to go
> from a reference of the FFI type to the wrapper.

That seems a bit strange as it doesn't provide any ownership transfer information but ok...

Should probably also be added to `src/boxed.rs` and `src/shared.rs` in the same way for consistency :)

cgwalters:
> That seems a bit strange as it doesn't provide any ownership transfer information but ok...

Right, I went with bumping the refcount (`from_glib_none`) because all of our use cases I suspect aren't going to hit any performance issues from doing so (vs borrowing aka `from_glib_borrow`).  I can't think of any use cases for taking ownership (`from_glib_full`).  Basically this "reference to ffi/sys type" seems like as noted a rather obscure special case.

That said, since this actually is a reference we could use Rust lifetimes to borrow...I think that'd involve a new `FromGlibReference` trait implemented by the wrapper macro on the ffi type or so?

sdroege:
> That said, since this actually is a reference we could use Rust lifetimes to borrow...I think that'd involve a new `FromGlibReference` trait implemented by the wrapper macro on the ffi type or so?

That sounds like more complications for little gain and a very special corner case :)

cgwalters:
Thinking about this more I'm looking more closely at [Borrowed](http://gtk-rs.org/docs/glib/translate/struct.Borrowed.html).  It feels this is much like [standard Pin type](https://doc.rust-lang.org/std/pin/struct.Pin.html) right?  If we did that, then alignment with what cxx-rs is doing would be much greater because `from_glib_borrow` would natively take exactly what it's outputting, right?

sdroege:
Indeed, that might work. Can you give it a try to see if that fits API-wise? :)

sdroege:
@cgwalters What's the status here?

sdroege:
@cgwalters Ping? :) The next release is going to be in a month or so. If you want this included we'd have to decide something in the next weeks.

cgwalters:
It turns out we needed more here, see https://github.com/coreos/rpm-ostree/pull/2574

I'm going to just land the code in rpm-ostree for now and gain experience with it, and we can consider landing these conversions in gtk-rs if it works out later.

sdroege:
Thanks, makes sense! If there's anything else that would be useful to have here to make your life easier, just create an issue or PR :)

PR #247, opened 2021-01-04
Doc checks

GuillaumeGomez:


sdroege:
Can you improve the description of the PR and also the commit messages? I don't know what the point of these changes is :)

sdroege (changes requested):
Can you improve the description of the PR and also the commit messages? I don't know what the point of these changes is :)

bilelmoussaoui:
What I don't like about this change is that we end up with more than 60 checks without having all the -sys enabled nor a windows/macos tests. I think that if we were to split the CI into a "matrix" for the configurations, it has to be per crate and it would run everything needed: build, test, clippy, rustfmt & gtk-rs checker

.github/workflows/CI.yml
```diff
@@ -17,6 +18,18 @@ jobs:
           - beta
           - nightly
           - "1.48.0"
+        triple:
```
pbor, line 21: Not a triple anymore :)
I would use something more generic as "projects" or "crates"

.github/workflows/CI.yml
```diff
@@ -17,6 +18,18 @@ jobs:
...
+          - { name: "cairo", features: "png,pdf,svg,ps,use_glib,v1_16,freetype,script,xcb,xlib,win32-surface", nightly: "--features 'png,pdf,svg,ps,use_glib,v1_16,freetype,script,xcb,xlib,win32-surface'", test_sys: true }
+          - { name: "gdk", features: "v3_24", nightly: "--features v3_24", test_sys: true }
+          - { name: "gdk-pixbuf", features: "v2_40", nightly: "--all-features", test_sys: true }
+          - { name: "gdkx11", features: "v3_24", nightly: "--all-features", test_sys: true }
+          - { name: "gio", features: "v2_66", nightly: "--all-features", test_sys: false }
+          - { name: "glib", features: "v2_66", nightly: "--all-features", test_sys: true }
+          - { name: "graphene", features: "v1_10", nightly: "", test_sys: false }
+          - { name: "gtk", features: "v3_24_9", nightly: "--all-features", test_sys: true }
+          - { name: "pango", features: "v1_46", nightly: "--all-features", test_sys: false }
+          - { name: "pangocairo", features: "", nightly: "--all-features", test_sys: true }
```
pbor, line 32: having features as the empty string failed for me because cargo does not like --features with no param... I went with having "args" (including the `--features` part) in the table, this also lets you handle the examples

.github/workflows/CI.yml
```diff
@@ -25,115 +38,64 @@ jobs:
         with:
           toolchain: ${{ matrix.rust }}
           override: true
-      # atk
       # nightly
-      - name: "atk: tests nightly"
-        run: xvfb-run --auto-servernum cargo test --manifest-path atk/Cargo.toml --all-features
+      - name: tests nightly
+        run: xvfb-run --auto-servernum cargo test --manifest-path ${{ matrix.triple.name }}/Cargo.toml ${{ matrix.triple.nightly }}
```
pbor, line 43: Can we run xvfb once as a `service` instead of wrapping each command?

.github/workflows/CI.yml
GuillaumeGomez: First, making it work, then cleanup! :laughing: But yes, you're absolutely right.

GuillaumeGomez:
So CI passed finally. How would you update it now @bilelmoussaoui ?

.github/workflows/CI.yml
bilelmoussaoui: I think there are actions that does that for you and make sure to shut down the service once the command has stopped running

.github/workflows/CI.yml
```diff
@@ -25,115 +38,67 @@ jobs:
...
+        run: cargo build --manifest-path ${{ matrix.conf.name }}/Cargo.toml --features "dox,embed-lgpl-docs"
+        if: matrix.rust == 'nightly' && matrix.conf.name != 'glib'
+      - name: doc removal
+        run: cargo build --manifest-path ${{ matrix.conf.name }}/Cargo.toml --features "purge-lgpl-docs"
+        if: matrix.rust == 'nightly' && matrix.conf.name != 'glib'
+      - name: glib doc check
+        run: cargo build --manifest-path ${{ matrix.conf.name }}/Cargo.toml --features "dox"
+        if: matrix.rust == 'nightly' && matrix.conf.name == 'glib'
+      - name: check diff
+        run: git diff -R --exit-code
```
bilelmoussaoui, line 73: I think this one could be moved the to other job along with gtk-rs checker?

.github/workflows/CI.yml
```diff
@@ -17,6 +18,18 @@ jobs:
           - beta
           - nightly
           - "1.48.0"
+        conf:
```
pbor, line 21: The checker job calls this `crate`... I am fine with either, but I would be consistent (I will use the same for windows)

.github/workflows/CI.yml
```diff
@@ -25,115 +38,67 @@ jobs:
...
+      - name: glib doc check
+        run: cargo build --manifest-path ${{ matrix.conf.name }}/Cargo.toml --features "dox"
+        if: matrix.rust == 'nightly' && matrix.conf.name == 'glib'
+      - name: check diff
+        run: git diff -R --exit-code
         if: matrix.rust == 'nightly'
+
+      - uses: bcomnes/cleanup-xvfb@v1
+
+  build-others:
```
pbor, line 78: Why do we need to split `build-others` to its own job? with a couple of tweaks it can be part of the main build

.github/workflows/CI.yml
```diff
@@ -25,115 +38,67 @@ jobs:
...
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: ${{ matrix.rust }}
+          override: true
+      # glib (macros)
       - name: "glib: tests glib macros nightly"
         working-directory: glib
         run: python3 clone_tests/clone_compilation_errors.py
         if: matrix.rust == 'nightly'
```
pbor, line 101: I may be misreading the patch, but it seems to me that `python3 clone_tests/clone_compilation_errors.py` could be run unconditionally instead of duplicating it for nightly

pbor (commented):
The clippy job should also use a matrix

.github/workflows/CI.yml
GuillaumeGomez: Good idea.

.github/workflows/CI.yml
GuillaumeGomez: You're right, I think it's an artifact. Removing the condition and the duplica.

GuillaumeGomez:
It finally worked! \o/

GuillaumeGomez:
Gonna merge it as is and we'll use this as new base for further developments.

PR #252, opened 2021-01-07
Reduce external dependencies for futures-related things

### LINKED ISSUE #251: Consider replacing futures dependency with sub-crates
glib currently depends on `futures`, pulling in all of [`futures-channel`, `future-core`, `futures-executor`, `futures-io`, `futures-sink`, `futures-task` and `futures-tokio`] as transitive dependencies. Unless you actually use functionality from all of those crates, it would be good to only depend on the specific sub-crates whose functionality is being used.

jplatte:
Closes #251.

Sorry for the formatting change in `gio/Cargo.toml`, I can undo that if you want.

This touches some unsafe code, but I'm fairly certain it is no less sound than before 😅

### DIFF
--- a/gio/Cargo.toml
+++ b/gio/Cargo.toml
@@ -47,10 +47,9 @@ git = "https://github.com/gtk-rs/lgpl-docs"
 libc = "0.2"
 bitflags = "1.0"
 once_cell = "1.0"
-futures = "0.3"
 futures-core = "0.3"
 futures-channel = "0.3"
-futures-util = "0.3"
+futures-util = { version = "0.3", features = ["io"] }
 futures-io = "0.3"
 ffi = { package = "gio-sys", path = "sys" }
 glib = { path = "../glib" }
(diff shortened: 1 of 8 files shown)


GuillaumeGomez (approved):
Less dependencies makes me happy.

GuillaumeGomez:
The regen check failure is weird though.

sdroege:
> The regen check failure is weird though.

Well it's because of the ` Remove unused imports ` commit

graphene/sys/tests/abi.rs
```diff
@@ -2,10 +2,8 @@
 // from gir-files (https://github.com/gtk-rs/gir-files)
 // DO NOT EDIT
 
-use graphene_sys::*;
 use std::env;
 use std::error::Error;
-use std::mem::{align_of, size_of};
```
sdroege, line 8: This breaks the CI

graphene/sys/tests/abi.rs
```diff
@@ -2,10 +2,8 @@
 // from gir-files (https://github.com/gtk-rs/gir-files)
 // DO NOT EDIT
 
-use graphene_sys::*;
```
bilelmoussaoui, line 5: Those changes are not intended I guess?

sdroege (commented):
Looks good to me otherwise

graphene/sys/tests/abi.rs
jplatte: That is kind of weird, IMHO there should at least be an `#[allow(unused_imports)]` then.

jplatte:
Removed the problematic commit.

sdroege (approved):

PR #254, opened 2021-01-09
Regenerate with latest gir/gir-files, remove use_boxed_functions

MarijnS95:
Regenerate after https://github.com/gtk-rs/gir-files/pull/74, https://github.com/gtk-rs/gir/pull/1032 and https://github.com/gtk-rs/gir/pull/1033.

---

#### Update submodules and regenerate

`gir-files` only had whitespace cleanups and fresh imports of gir files, none of which affect code generated here.
`gir` however deprecated and removed `use_boxed_functions`, hence the replacement of `g_boxed_copy/free` with `g_unix_mount_point_copy/free`. See the previous commit for more details.

---

#### gio: Remove deprecated use_boxed_functions override

`Gio.UnixMountPoint` likely still uses `use_boxed_functions` because `g_unix_mount_point_copy` is only available on 2.54. However, the `version` statement on the next line already ensures this type is only available on that version or higher.

If anyone plans to support a lower version later on this can easily be coded up in gir to replace the panic on version mismatch with some `cfg_if` or `#[cfg()]` logic, just like what was done in [gstreamer-rs].

[gstreamer-rs]: https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/677/diffs?commit_id=3c610e12e5a72f3d25949d84e5a4756b7545e34a

---

CC @sdroege


sdroege (approved):

sdroege:
> `Gio.UnixMountPoint` likely still uses `use_boxed_functions` because `g_unix_mount_point_copy` is only available on 2.54. However, the `version` statement on the next line already ensures this type is only available on that version or higher.

Yeah, the reason is that the `get_type()` function was added in 2.54 as was the `copy()` function. The `free()` function existed since 2.16, but we need either `get_type()` or both `copy()` and `free()` for being able to bind a type.

MarijnS95:
Yeah, we can't implement this without `get_type` so the paragraph about wishing to support it on a lower version (in PR description and commit) is moot.

Luckily you added `version = "2.54"` for that function in sys as the `GLIB_AVAILABLE_IN_2_54` macro did not seem to make it to `Gio-20.gir`, the only mention is `glib:get-type="g_unix_mount_point_get_type"` which is why I did not notice :grimacing:

Issue #54, opened 2020-11-09
ABI tests are failing


ignatenkobrain:
```rust
Mock Version: 1.3.4
Mock Version: 1.3.4
ENTER ['do'](['bash', '--login', '-c', '/usr/bin/rpmbuild -bs --target x86_64 --nodeps /builddir/build/SPECS/rust-glib-sys.spec'], chrootPath='/var/lib/mock/f30-build-13706180-992465/root'env={'TERM': 'vt100', 'SHELL': '/bin/bash', 'HOME': '/builddir', 'HOSTNAME': 'mock', 'PATH': '/usr/bin:/bin:/usr/sbin:/sbin', 'PROMPT_COMMAND': 'printf "\\033]0;<mock-chroot>\\007"', 'PS1': '<mock-chroot> \\s-\\v\\$ ', 'LANG': 'en_US.UTF-8'}shell=Falselogger=<mockbuild.trace_decorator.getLog object at 0x7f650d7f4fd0>timeout=172800uid=1000gid=425user='mockbuild'nspawn_args=[]printOutput=False)
Executing command: ['bash', '--login', '-c', '/usr/bin/rpmbuild -bs --target x86_64 --nodeps /builddir/build/SPECS/rust-glib-sys.spec'] with env {'TERM': 'vt100', 'SHELL': '/bin/bash', 'HOME': '/builddir', 'HOSTNAME': 'mock', 'PATH': '/usr/bin:/bin:/usr/sbin:/sbin', 'PROMPT_COMMAND': 'printf "\\033]0;<mock-chroot>\\007"', 'PS1': '<mock-chroot> \\s-\\v\\$ ', 'LANG': 'en_US.UTF-8'} and shell False
Building target platforms: x86_64
Building for target x86_64
Wrote: /builddir/build/SRPMS/rust-glib-sys-0.7.0-1.fc30.src.rpm
Child return code was: 0
[...]

PR #256, opened 2021-01-11
CI: add CI for Windows

pbor:
Add a build-win job, which builds the create with MSVC.
The job builds the gtk dependencies using gvsbuild and caches them
so that they are not rebuilt on each run.
For now I just added the stable toolchain and x86_64 arch.
Many steps are currently failing (eg ABI checks), so I left them
commented out.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -1,0 +1,129 @@
+on:
+  push:
+    branches: [master]
+  pull_request:
+
+name: Windows
+
+jobs:
+
+  gvsbuild:
+    name: build GTK binaries with gvsbuild
+    runs-on: windows-latest
+
+    env:
+      # git revision of gvsbuild we use for to build GTK and the other dependencies
+      gvsbuildref: 69cdac34c36d74eeae37006bb0fcedc64ee518dd
+
+      # bump this number if you want to force a rebuild of gvsbuild with the same revision
+      gvsbuildupdate: 1
+
+    outputs:
+      cachekey: ${{ steps.output.outputs.cachekey }}
+
+    steps:
+      # this is needed for the cache restore to work
+      - name: (GTK binaries) create dir
+        run: mkdir C:\gtk-build\gtk\x64\release
+
+      - name: (GTK binaries) get from cache
+        uses: actions/cache@v2
+        id: cache
+        with:
+          path: C:\gtk-build\gtk\x64\release\**
+          key: gvsbuild-${{ env.gvsbuildupdate }}-${{ env.gvsbuildref }}
+
+      - name: (GTK binaries) checkout gvsbuild
+        if: steps.cache.outputs.cache-hit != 'true'
+        uses: actions/checkout@v2
+        with:
+          repository: wingtk/gvsbuild
+          ref: ${{ env.gvsbuildref }}
+          path: gvsbuild
+
+      # Temporarily move the preinstalled git, it causes errors related to cygwin.
+      - name: (GTK binaries) move git binary
+        if: steps.cache.outputs.cache-hit != 'true'
+        run: move "C:\Program Files\Git\usr\bin" "C:\Program Files\Git\usr\notbin"
+        shell: cmd
+
+      - name: (GTK binaries) run gvsbuild
+        if: steps.cache.outputs.cache-hit != 'true'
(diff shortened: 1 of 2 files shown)


pbor:
Marked as draft to get some feedback on the approach

.github/workflows/CI.yml
```diff
@@ -215,6 +215,271 @@ jobs:
 
       - uses: bcomnes/cleanup-xvfb@v1
 
+  build-win:
```
bilelmoussaoui, line 218: I would move the Window CI into a separate workflow (as a separate file) to avoid making the "main" CI's configuration more complex

pbor: I agree the file is very long and needs to be split... I wonder however if that is the right split: I was thinking that maybe it is better to split functionally instead of per platform. For instance have a file that is for quick checks (fmt etc) and then a file with the full build matrix that runs if the first is successfull

.github/workflows/CI.yml
```diff
@@ -215,6 +215,271 @@ jobs:
...
+
+      # GDK
+
+      - name: gdk test
+        uses: actions-rs/cargo@v1
+        with:
+          command: test
+          args: --manifest-path gdk/Cargo.toml --features v3_24
+
+      # FAILS: abi check
```
bilelmoussaoui, line 324: Most of the abi tests are failing,  see #54. I would try to keep them in-sync if possible

.github/workflows/CI.yml
bilelmoussaoui: There's #247 for a little clean up of the build. The CI as is right now is just bare-bone to get us started post-migration to a single repository, any improvements are very much welcome :)

.github/workflows/CI.yml
sdroege: And if different sys crates fail on Windows than on Linux that should be documented and investigated shortly at least.

pbor:
Hold on with the review, I am improving this to use a matrix

GuillaumeGomez:
I'll try to finish #247 today so that you can use it in order to add windows CI.

pbor:
Updated the CR

Here is a successful run https://github.com/pbor/gtk-rs/runs/1687412308?check_suite_focus=true

pbor:
(the CI on this PR failed because it failed to download some of the deps to build... Once the cache is built this should not be a problem)

pbor:
Rebased on top of master after, https://github.com/gtk-rs/gtk-rs/pull/247 was merged.

I think this is now ready.

.github/workflows/CI.yml
GuillaumeGomez: The argument still stand. :)

pbor: I am not sure what you are asking me to do to be honest... when we started the CI file was much longer and had a lot of jobs and steps, but right now even with my patch it is 270 which is not bad. I looked into various other rust repositories and the standard practice seems to be to use different workflows when they have different triggers.

bilelmoussaoui:
Do you mind adding a comment to https://github.com/gtk-rs/gtk-rs/issues/54 about the various -sys ones that you had to disable that are enabled in the other CI? Otherwise looks good to me

bilelmoussaoui (approved):

pbor:
> Do you mind adding a comment to #54 about the various -sys ones that you had to disable that are enabled in the other CI? Otherwise looks good to me

Done. I also opened https://github.com/gtk-rs/gtk-rs/issues/280 to track the `test_paths` failure

.github/workflows/windows.yml
```diff
@@ -0,0 +1,129 @@
...
+name: Windows
+
+jobs:
+
+  gvsbuild:
+    name: build GTK binaries with gvsbuild
+    runs-on: windows-latest
+
+    env:
+      # git revision of gvsbuild we use for the GTK dependendency
```
GuillaumeGomez, line 15: "dependendency"? :p

GuillaumeGomez (approved):

GuillaumeGomez:
Just a small typo then it's good to go. Thanks!

pbor:
> Just a small typo then it's good to go. Thanks!

fixed!

GuillaumeGomez:
Thanks!

PR #257, opened 2021-01-12
add ActionMap/ActionGroup subclassing support

### LINKED ISSUE #190: gio ActionMap/ActionGroup subclassing support
It's pretty useful to be able to subclass gio::ActionMap when creating a custom widget with it's own actions group instead of having to add an actions field into your `imp::WidgetStruct`

bilelmoussaoui:
Fixes #190

### DIFF
--- a/gio/src/subclass/action_group.rs
+++ b/gio/src/subclass/action_group.rs
@@ -1,0 +1,466 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use crate::ActionGroup;
+use glib::subclass::prelude::*;
+use glib::translate::*;
+use glib::{Cast, GString, ObjectExt, Quark, Variant, VariantType};
+use once_cell::sync::Lazy;
+use std::ptr;
+
+pub trait ActionGroupImpl: ObjectImpl {
+    fn action_added(&self, action_group: &Self::Type, action_name: &str) {
+        unsafe {
+            let type_ = ffi::g_action_group_get_type();
+            let iface = glib::gobject_ffi::g_type_default_interface_ref(type_)
+                as *mut ffi::GActionGroupInterface;
+            assert!(!iface.is_null());
+
+            if let Some(f) = (*iface).action_added.as_ref() {
+                f(
+                    action_group
+                        .unsafe_cast_ref::<ActionGroup>()
+                        .to_glib_none()
+                        .0,
+                    action_name.to_glib_none().0,
+                );
+            }
+
+            glib::gobject_ffi::g_type_default_interface_unref(iface as glib::ffi::gpointer);
+        }
+    }
+
+    fn action_enabled_changed(&self, action_group: &Self::Type, action_name: &str, enabled: bool) {
+        unsafe {
+            let type_ = ffi::g_action_group_get_type();
+            let iface = glib::gobject_ffi::g_type_default_interface_ref(type_)
+                as *mut ffi::GActionGroupInterface;
+            assert!(!iface.is_null());
+
+            if let Some(f) = (*iface).action_enabled_changed.as_ref() {
+                f(
+                    action_group
+                        .unsafe_cast_ref::<ActionGroup>()
+                        .to_glib_none()
+                        .0,
+                    action_name.to_glib_none().0,
+                    enabled.to_glib(),
+                );
+            }
+            glib::gobject_ffi::g_type_default_interface_unref(iface as glib::ffi::gpointer);
+        }
+    }
+
+    fn action_removed(&self, action_group: &Self::Type, action_name: &str) {
+        unsafe {
+            let type_ = ffi::g_action_group_get_type();
+            let iface = glib::gobject_ffi::g_type_default_interface_ref(type_)
+                as *mut ffi::GActionGroupInterface;
+            assert!(!iface.is_null());
+
+            if let Some(f) = (*iface).action_removed.as_ref() {
+                f(
+                    action_group
+                        .unsafe_cast_ref::<ActionGroup>()
+                        .to_glib_none()
+                        .0,
+                    action_name.to_glib_none().0,
+                );
+            }
+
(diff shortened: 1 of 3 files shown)


gio/src/subclass/action_group.rs
```diff
@@ -0,0 +1,371 @@
...
+
+use crate::ActionGroup;
+use glib::subclass::prelude::*;
+use glib::translate::*;
+use glib::{Cast, GString, ObjectExt, Quark, Variant, VariantType};
+use once_cell::sync::Lazy;
+use std::ptr;
+
+pub trait ActionGroupImpl: ObjectImpl {
+    fn action_added(&self, action_group: &Self::Type, action_name: &str);
```
sdroege, line 11: There are default implementations for many of these

sdroege: And this specifically is a signal default handler, so can stay `NULL` anyway

gio/src/subclass/action_group.rs
```diff
@@ -0,0 +1,371 @@
...
+        action_group: &Self::Type,
+        action_name: &str,
+    ) -> Option<VariantType>;
+    fn has_action(&self, action_group: &Self::Type, action_name: &str) -> bool;
+    fn list_actions(&self, action_group: &Self::Type) -> Vec<String>;
+    fn query_action(
+        &self,
+        action_group: &Self::Type,
+        action_name: &str,
+    ) -> Option<(bool, VariantType, VariantType, Variant, Variant)>;
```
sdroege, line 45: This has missing annotations in C. The 4 last parts of the tuple are all `nullable`.

gio/src/subclass/action_group.rs
```diff
@@ -0,0 +1,371 @@
...
+        }
+        if !state_hint.is_null() {
+            ptr::write(state_hint, {
+                let ret = rs_state_hint.to_glib_full();
+                wrap.set_qdata(*ACTION_GROUP_QUERY_ACTION_STATE_HINT_QUARK, ret);
+                ret
+            });
+        }
+        if !state.is_null() {
+            ptr::write(state, {
```
sdroege, line 360: Why `ptr::write()` here btw? You could also just do `*state = ...` :)

gio/src/subclass/action_map.rs
```diff
@@ -0,0 +1,79 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use crate::{Action, ActionMap};
+use glib::subclass::prelude::*;
+use glib::translate::*;
+use glib::{Cast, GString, IsA, ObjectExt, Quark};
+use once_cell::sync::Lazy;
+
+pub trait ActionMapImpl: ObjectImpl {
```
sdroege, line 9: This file looks all ok

sdroege (commented):
actiongroup needs some changes but otherwise ok

sdroege:
See also https://github.com/gtk-rs/gobject-subclass/tree/master/gio-subclass/src/auto for some earlier version

gio/src/subclass/action_map.rs
```diff
@@ -0,0 +1,79 @@
...
+    ) {
+        let action_map_iface = &mut *(iface as *mut ffi::GActionMapInterface);
+
+        action_map_iface.lookup_action = Some(action_map_lookup_action::<T>);
+        action_map_iface.add_action = Some(action_map_add_action::<T>);
+        action_map_iface.remove_action = Some(action_map_remove_action::<T>);
+    }
+}
+
+static ACTION_MAP_LOOKUP_ACTION_QUARK: Lazy<Quark> =
```
bilelmoussaoui, line 31: I'm wondering if the quark here has to be per action ?

sdroege: Well, no. But you should store a hash table of action_name->results here I guess.

gio/src/subclass/action_group.rs
sdroege: Now it would simply panic instead of doing nothing. There is no default implementation of it, but because it's a signal handler nothing would happen if it stays `NULL`.

gio/src/subclass/action_group.rs
sdroege: You should probably set them all to `NULL` before returning to C if they were `None` in Rust

gio/src/subclass/action_group.rs
bilelmoussaoui: I guess I have properly fixed it now :P

sdroege (approved):

ids1024:
It could be useful to have an example of this.

https://developer.gnome.org/gio/stable/GActionGroup.html

> Implementations of GActionGroup should provide implementations for the virtual functions g_action_group_list_actions() and g_action_group_query_action(). The other virtual functions should not be implemented - their "wrappers" are actually implemented with calls to g_action_group_query_action().

I'm somewhat confused about how this relates to the `ActionGroupImpl` trait here. This suggests only implementing two virtual functions, while `ActionGroupImpl` seems to have 9 methods without default implementations.

bilelmoussaoui:
> It could be useful to have an example of this.
> 
> https://developer.gnome.org/gio/stable/GActionGroup.html
> 
> > Implementations of GActionGroup should provide implementations for the virtual functions g_action_group_list_actions() and g_action_group_query_action(). The other virtual functions should not be implemented - their "wrappers" are actually implemented with calls to g_action_group_query_action().
> 
> I'm somewhat confused about how this relates to the `ActionGroupImpl` trait here. This suggests only implementing two virtual functions, while `ActionGroupImpl` seems to have 9 methods without default implementations.

Oups, indeed, it seems to have defaults for all the methods except `query_action` & `list_actions` per https://github.com/GNOME/glib/blob/master/gio/gactiongroup.c#L239-L245

bilelmoussaoui:
@ids1024 do you want to have a look at it? Otherwise I will try to send a patch this weekend

Issue #259, opened 2021-01-12
Provide wrapper around `glib-compile-resources`


ids1024:
It can be useful to embed resources using `glib-compile-resources` in `build.rs`, then using `include_bytes!` to include it in the executable. (Equivalent to using `glib-compile-resources --generate-source` in C).

This requires a certain amount of boilerplate, so it could be helpful for gtk-rs to provide a wrapper for it, unless there's already something I'm unaware of.

bilelmoussaoui:
As a bunch of people already asked on IRC on how to do that, having a helper along with an example would be nice.

ids1024:
It should produce a `cargo:rerun-if-changed=` line for every file listed in the `.gresource.xml` file. Though that's a bit awkward since it would require an xml parser.

bilelmoussaoui:
Well, if you really want that kind of integration a build system like meson on top of cargo is more suitable I guess.

ids1024:
Nevermind depending on an xml parser; `glib-compile-resources --generate-dependencies` outputs a newline separated list of dependencies.

PR #261, opened 2021-01-13
gio: Add `glib-compile-resources` wrapper, and macro to include

ids1024:
Updates examples to use this.

Implementation of https://github.com/gtk-rs/gtk-rs/issues/259.

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -29,6 +29,9 @@ optional = true
 version = "0.6"
 optional = true
 
+[build-dependencies.gio]
+path = "../gio"
+
 [[bin]]
 name = "accessibility"
 
--- a/examples/src/bin/gtktest.rs
+++ b/examples/src/bin/gtktest.rs
@@ -1,7 +1,7 @@
 #![cfg_attr(not(feature = "gtk_3_10"), allow(unused_variables, unused_mut))]
 
 use gtk::prelude::*;
-use gtk::{gdk, glib};
+use gtk::{gdk, gio, glib};
 use gtk::{
     AboutDialog, AppChooserDialog, ApplicationWindow, Builder, Button, Dialog, Entry,
     FileChooserAction, FileChooserDialog, FontChooserDialog, RecentChooserDialog, ResponseType,
@@ -187,6 +187,8 @@ fn build_ui(application: &gtk::Application) {
 }
 
 fn main() {
+    gio::resources_register_include!("compiled.gresource").unwrap();
+
     let application = gtk::Application::new(
         Some("com.github.gtk-rs.examples.gtktest"),
         Default::default(),
--- a/gio/src/resource.rs
+++ b/gio/src/resource.rs
@@ -1,8 +1,11 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
-use crate::Resource;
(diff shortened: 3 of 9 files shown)


gio/src/resource.rs
```diff
@@ -28,3 +30,54 @@ impl Resource {
         }
     }
 }
+
+/// Call from build script to run `glib-compile-resources` to generate compiled gresources to embed
+/// in binary with [resources_register_include]. `target` is relative to `OUT_DIR`.
```
sdroege, line 35: Add a code example here

sdroege (commented):
Looks good otherwise

examples/src/bin/gtktest.rs
```diff
@@ -187,6 +187,8 @@ fn build_ui(application: &gtk::Application) {
 }
 
 fn main() {
+    gio::resources_register_include!("compiled.gresource").unwrap();
```
pbor, line 190: I am a bit rusty (pun intended) in gtk, but iirc when using gtkapplication in C, registering the resources happens automatically... why can't we piggy back on the same mechanism?

sdroege: The way how that works is via shared library constructors, which are not available on all platforms (see GLib docs) and require special code.

This https://crates.io/crates/ctor crate could be used for doing that but it will probably have some impact on portability.

examples/Cargo.toml
```diff
@@ -29,6 +29,9 @@ optional = true
 version = "0.6"
 optional = true
 
+[build-dependencies.gtk]
+path = "../gtk"
```
GuillaumeGomez, line 33: Shouldn't it be gio then? It seems a bit overkill to import the whole `gtk` crate if you only use `gio`.

gio/src/resource.rs
```diff
@@ -28,3 +30,54 @@ impl Resource {
...
+pub fn compile_resources(sourcedir: &str, gresource: &str, target: &str) {
```
GuillaumeGomez, line 36: Shouldn't `sourcedir` be called `source_dir` and be a `AsRef<Path>`?

examples/src/bin/gtktest.rs
ids1024: Ah, I was wondering if there's a crate that supports it.

Not sure what platforms that won't work on. But I think to integrate it the Rust code using the library might still need to add a `mod` declaration to include the generated code (unless the build script compiles a separate static library that is somehow linked in...).

sdroege: Should we wait for you testing this or ... ?

sdroege: Or maybe usage of that crate could simply be documented as part of the macro?

sdroege (approved):

examples/src/bin/gtktest.rs
ids1024: I'm not really sure the `ctor` crate is better here than just putting it at the start of `main()`, which shouldn't be any more code and is less magical.

bilelmoussaoui (approved):

examples/src/bin/gtktest.rs
sdroege: Ok!

PR #263, opened 2021-01-15
gio: use default methods when possible for ActionGroup

bilelmoussaoui:
cc @ids1024


ids1024:
Now this only requires implementing two methods, as expected. Though it would be good to have an example and perhaps a unit test (or more).

sdroege (approved):
Let's get this in and add a test/example later? Or do you want to prepare one now?

bilelmoussaoui:
> Let's get this in and add a test/example later? Or do you want to prepare one now?

I do want to add one, but not today :) I would prefer to have this merged as it's right now

PR #266, opened 2021-01-17
Add proc macro for binding methods to action maps

### LINKED ISSUE #240: Provide a proc macro to register actions
One very cool things about GTK is the usage of GAction. You can either easily add an action to a widget that implements the necessary interfaces to be an GActionGroup  (`gio::ActionMap` & `gio::ActionGroup`)  or you have to add a SimpleActionGroup to your widget yourself.  While the creation part of the actions is very verbose already with all the casting from GAction to GSimpleAction, it's currently not easy nor nice to write.

I have written in the past few macros that did nothing but wrap the procedure of creating an action, adding to the action group along with all the casting stuff and it looks currently more or less like this

```rust
fn setup_gactions(&self) {
        // Quit
        action!(
            self,
            "quit",
            clone!(@weak self as app => move |_, _| {
                app.quit();
            })
        );

        // About
        action!(
            self,
            "about",
            clone!(@weak self as app => move |_, _| app.display_about_dialog())
        );

        let win = self.get_main_window();
        action!(
            win,
            "fullscreen",
            clone!(@weak self as app => move |_, _| {
                app.get_main_window().toggle_fullscreen();
            })
        );
    }
```

Note that more complex apps can have a ton of actions, so having to define them "twice" is not great.

I suggest we provide a proc macro based on the future work in https://github.com/gtk-rs/gtk-rs/issues/128 to automatically create an action and use the annotated function as the activate callback. 

So that the example above would look more or less like this
```rust
#[gaction(name="quit")]
fn on_quit(&self, action: gio::SimpleAction, target: Option<glib::Value>) {
    self.quit();
}
#[gaction]
fn about(&self, action: gio::SimpleAction, target: Option<glib::Value>) {
    let dialog = gtk::AboutDialog::new();
    ...
    dialog.show();
}

#[gaction]
fn fullscreen(&self, action: gio::SimpleAction, target: Option<glib::Value>) {
    self. app.get_main_window().toggle_fullscreen();
}
```

This will require having some kind of `register_actions` methods to call later on to register all those actions in a specific action group (or something else? not sure how that could work)

andy128k:
Closes #240


gio-macros/src/lib.rs
```diff
@@ -0,0 +1,14 @@
...
+
+extern crate proc_macro;
+
+mod action_map;
+
+use proc_macro::TokenStream;
+use syn::{parse_macro_input, ItemImpl};
+
+#[proc_macro_attribute]
+pub fn actions(_attr: TokenStream, item: TokenStream) -> TokenStream {
```
sdroege, line 11: This could use some documentation with code example(s) :)

andy128k: Yea, sure. I just wanted to share what is done at the moment. Also some tests would be great too.

bilelmoussaoui: It looks overall good to me, I just can't see how you set the ActionMap, so an example would help figure out how the API is supposed to work :)

gio-macros/src/lib.rs
```diff
@@ -0,0 +1,63 @@
...
+///         let app = Self(gio::Application::new(
+///             Some("com.example.MyApplication"),
+///             gio::ApplicationFlags::FLAGS_NONE,
+///         ));
+///         app.register_actions(&app.0);
+///         app
+///     }
+/// }
+///
+/// #[gio::actions]
```
bilelmoussaoui, line 37: What I was actually thinking of is to rather uses this with subclasses, and have the gio::actions attribute implement ActionMap/ActionGroup for that subclass automatically and then you can add the actions directly to self as it would be an action map.

andy128k: You still may use this macro with subclasses. Just invoke `self.register_actions(&self)` in a constructor method.

bilelmoussaoui: Sure, I guess we can provide another derive macro like the CompositeTemplate for such use case I guess.

andy128k: @bilelmoussaoui Frankly, I was under impression that Gtk is moving [away from inheritance](https://blog.gtk.org/2020/04/23/custom-widgets-in-gtk-4-introduction/).

> The general direction of our API changes has been to emphasize delegation over subclassing.

>  many widgets are now final classes

bilelmoussaoui: > @bilelmoussaoui Frankly, I was under impression that Gtk is moving [away from inheritance](https://blog.gtk.org/2020/04/23/custom-widgets-in-gtk-4-introduction/).
> 
> > The general direction of our API changes has been to emphasize delegation over subclassing.
> 
> > many widgets are now final classes

You're mis-understanding what the announcement meant I guess. in gtk3 creating a custom widget was a painful process which was simplified in gtk4, and gtk developers expect you to create custom widgets (a subclass of gtk::Widget) where you compose multiple pre-defined widgets.

bilelmoussaoui: Without subclassing you can't use composite templates, properties, signals and all what gtk offers :)

alatiera: > i Frankly, I was under impression that Gtk is moving [away from inheritance](https://blog.gtk.org/2020/04/23/custom-widgets-in-gtk-4-introduction/).

It's more of "everything now will be inheriting from `GtkWidget` directly rather than 4 different classes deep"

gio-macros/src/lib.rs
```diff
@@ -0,0 +1,63 @@
...
+/// impl MyApplication {
+///     fn action1(&self) {
```
sdroege, line 39:
```suggestion
///     #[action]
///     fn action1(&self) {
```

Would probably be better than taking any and every function and turning it into an action

gio-macros/src/lib.rs
```diff
@@ -0,0 +1,63 @@
...
+///         // handle "action1"
+///     }
+///
+///     // Explicitely specify action name
+///     #[action(name = "action_second")]
+///     fn action2(&self) {
+///         // handle "action_second"
+///     }
+///
+///     // Action with a parameter
```
sdroege, line 49: Does it also handle return values?

gio-macros/src/lib.rs
```diff
@@ -0,0 +1,63 @@
...
+///     #[action(parameter_type = "s")]
+///     fn action3(&self, param: String) {
```
sdroege, line 51: What if the parameter type in the attribute and the function parameter don't match? Can we infer one from the other to not require both to be present? You can get the GVariant type from a Rust type in any case at runtime.

gio-macros/src/lib.rs
sdroege: Seems good to me too apart from the comments above but an example would be nice :)

gio-macros/tests/actions.rs
```diff
@@ -0,0 +1,79 @@
+use gio::prelude::*;
+use std::cell::RefCell;
+use std::rc::Rc;
+
+#[derive(glib::Downgrade)]
+pub struct MyActionGroup {
```
sdroege, line 6: How does this interfer with an actual `GActionGroup` implementation? @bilelmoussaoui

gio-macros/src/lib.rs
andy128k: They *are* actions. All methods inside this `impl` item are actions. It is an `impl` item who is annotated and produces an `ActionMap`. E.g. it is possible to have more than one `ActionMap` implemented by a single type.

gio-macros/src/lib.rs
sdroege: There are stateful actions though. `Action::change_state()` / `Action::get_state()`

gio-macros/src/lib.rs
sdroege: I guess that's why I'm asking for a proper example that makes use of most of the features of this :)

gio-macros/src/lib.rs
andy128k: It types doesn't match it will be a runtime error (`g_critical`). I am not sure, it is possible to derive the type without bloating the code (tuples of arbitrary length could be a PITA).

gio-macros/src/lib.rs
andy128k: This macro doesn't cover/implement stateful actions.

gio-macros/src/lib.rs
andy128k: There are two :) One in a doc-test and another in tests.

I also mentioned one in #240. [There is a regular `impl` block with private methods and one more `impl` block with actions](https://github.com/andy128k/find_duplicates/blob/master/find_duplicates_gui/src/main_window.rs#L377).

gio-macros/tests/actions.rs
andy128k: It doesn't. The macro just generates one method (`register_actions`) which allows you to register all actions in an arbitrary `ActionMap`.

`GSimpleActionGroup` is used here because it is the only class in `gio` which implements both `GActionMap` and `GActionGroup` and can be tested end-to-end.

gio-macros/src/lib.rs
sdroege: The traits in `glib::variant` could be used for this. They already handle exactly these things.

gio-macros/src/lib.rs
sdroege: Please document that then :)

gio-macros/src/lib.rs
sdroege: I meant a meaningful example how this would be used in practice as part of the examples or examples4 subdirectory.

Or maybe the docs and docs example just has to be extended to cover these questions too.

gio-macros/src/lib.rs
andy128k: Oh, I overlooked that `StaticVarianType` exists. So, it seems I have something to do this evening :)

gio-macros/src/lib.rs
andy128k: Sure. I will create an example with some realistic actions.

gio-macros/tests/actions.rs
```diff
@@ -0,0 +1,79 @@
...
+            group: gio::SimpleActionGroup::new(),
+            log: Default::default(),
+        };
+        app.register_actions(&app.group);
+        app
+    }
+}
+
+#[gio::actions]
+impl MyActionGroup {
```
bilelmoussaoui, line 23: Is this really needed? I would rather have everything on the same Impl and decorate all the actions with `#[action...]`

andy128k: I am afraid it is not possible. In this case proc-macro will not be able to generate a single method which gathers all actions into an ActionMap.

bilelmoussaoui: You don't need to gather the actions if you assume that self is a `glib::IsA<gio::ActionMap>` then, you can directly add the actions without requiring an extra method. Of course that would only work for widgets/object that implement ActionMap. Which means gtk::Application & gtk::ApplicationWindow for now.  We can then have a proc macro that would automatically implement ActionMap on top of a gobject subclass which would look much cleaner API wise I guess.

sdroege: > Of course that would only work for widgets/object that implement ActionMap.

Or something that implements `Deref<Target=ActionMap>` :)

andy128k: 1. Requiring that `Self` implements `ActionMap` makes this macro useless for my personal projects where I do not use inheritance but newtypes to create abstractions.
2. Even if `Self: ActionMap` is assumed, it doesn't help. Code which generates an action should be invoked from somewhere. Attribute proc-macro attached to `impl` item allows to append a single method (`register_actions` in this case) and then this method can be called somewhere. Attribute proc-macro attached to each action handler individually will not be able to do so. It can generate another function and that's it. There is no way to inject *invocation* of these functions (Unless we assume additional requirement. `GtkWidget::created-event` would fit here if existed :laughing: )

gio-macros/src/lib.rs
andy128k: This is done. And `parameter_type` parameter is removed.

gio-macros/tests/actions.rs
bilelmoussaoui: >     1. Requiring that `Self` implements `ActionMap` makes this macro useless for my personal projects where I do not use inheritance but newtypes to create abstractions.

Sure thing, the reason why everyone went that way when gtk-rs apps started showing up is that subclassing wasn't possible (required a ton of boilerplate). It still requires a bit of boilerplate and such macro would help in that direction :)  

>     2. Even if `Self: ActionMap` is assumed, it doesn't help. Code which generates an action should be invoked from somewhere. Attribute proc-macro attached to `impl` item allows to append a single method (`register_actions` in this case) and then this method can be called somewhere. Attribute proc-macro attached to each action handler individually will not be able to do so. It can generate another function and that's it. There is no way to inject _invocation_ of these functions (Unless we assume additional requirement. `GtkWidget::created-event` would fit here if existed laughing )

I don't see why you would need to register them, if the #[action] is called on a function already on self, you have access to self, so you can just do self.add_action directly instead of storing them somewhere, no?

andy128k: @bilelmoussaoui Who should call `self.add_action`? Or, where should macro put that `self.add_action`?

bilelmoussaoui: Ah indeed. We can have a trait (see the composite template macro) with a register_actions method that the user would have to call when the object is `constructed`. Until we have some kind of way to automatically hook those as the user could be using a combo of different macros (composite template, the actions one, the gobject properties one, something else in the future?)

andy128k: Proc-macros are **stateless** and they cannot generate **one** `register_action` method (or one implementation of some trait) while being invoked **multiple** times (once per annotation).

bilelmoussaoui: Here's a rough pseudo code idea of what I had in mind
```rust
mod imp {

    #[derive(CompositeTemplate, Actions)]
    #[actions(group="widget")]
    pub struct Widget {

    }

    impl ObjectSubclass for Widget {
        ....

        fn type_init(type_: subclass::InitializedType<Self>) {
            type_.add_interface::<gio::ActionMap>();
            type_.add_interface::<gio::ActionGrouo>();
        }
    }

    impl ObjectImpl for Widget {
        fn constructed(&self, obj: &Self::Type) {
            obj.register_actions();
        }
    }

    # both those traits would be automatically implemented when using #[actions(group....)]
    impl ActionMap for Widget {}
    impl ActionGroup for Widget {}
}

glib::wrapper! {
    pub struct Widget(ObjectSubclass<imp::Widget>) @derive gtk::Widget, @implements gio::ActionMap, gio::ActionGroup;
}


impl widget {

    #[action]
    fn name(&self, action: &gio::Action, target: Option<&glib::Value>) {

    }
}
```

andy128k:
```rust
    #[action]
    fn name(&self, action: &gio::Action, target: Option<&glib::Value>) {

    }
```

What a macro should emit here? Notice, that it is **inside** of an `impl` block and such block can contain functions or constants only. Nothing else.

bilelmoussaoui: Does it have to register anything actually? In my idea it was mostly used to identify the action functions so that you can register them later

andy128k: Of course, it should emit something. At least the function itself :laughing: with **stripped out** attribute (otherwise it won't compile).

bilelmoussaoui: Ah yeah, that's what I meant, as in not do anything else other than what the function contains ^^

andy128k:
Support of stateful actions is added.

andy128k:
I've done what I planned to do with this macro (at least for the first iteration). :sweat_smile: 

There is one minor issue which is easy to resolve, but I'd like to have a feedback/advice before. Here is an excerpt from Gio documentation.
>  For stateful actions where the state type is equal to the parameter type, the default is to forward them directly to “change-state”. This should allow almost all users of GSimpleAction to connect only one handler or the other.

With current implementation of the macro this will not work.

```rust
#[action(name = "my-action", change-state)]
fn my_action_change_state(&self, state: T) -> Option<T> { ... }
```

Such declaration of a `change-state` handler will generate: `gio::SimpleAction::new_stateful(None, T::default().to_variant())`. So, there is no parameter type in the action's definition. Type of a parameter is taken from a handler of `activate` signal. So, to make it work, `activate` handler is needed.

```rust
#[action(name = "my-action")]
fn my_action_activate(&self, _state: T, parameter: T) -> Option<T> {
    Some(parameter)
}

#[action(name = "my-action", change-state)]
fn my_action_change_state(&self, state: T) -> Option<T> { ... }
```

Which is just an ugly boilerplate and cancels the original idea to have only one handler.

So, a possible solution is to add one more annotation to a `change-state` handler. Something like this:

```rust
#[action(name = "my-action", change-state, use-same-parameter-type-as-a-state-type)]
fn my_action_change_state(&self, state: T) -> Option<T> { ... }
```

But I cannot come up with a good name for it (`use-same-parameter-type-as-a-state-type` is nothing but a joke :smile:).

sdroege:
@BrainBlasted does this look good to you API-wise now? I'm not doing much with actionmaps, so don't really know :)

andy128k:
Instead of introducing an opt in flag, I added an opt out one. So, an action parameter type is derived from the state type by default. To opt out a flag `no_parameter` should be specified.

PR #269, opened 2021-01-18
gtk: specify template as an attribute of CompositeTemplate

pbor:
The source can be a file, a resource, or a string.

WIP: because I only quickly tested the file case using the example file

Also, it would be great to automagically hook into `class_init` instead of requiring to call `template_init`, but I have no idea on how to do that

### DIFF
--- a/gtk3-macros/src/composite_template_derive.rs
+++ b/gtk3-macros/src/composite_template_derive.rs
@@ -10,6 +10,21 @@ use std::string::ToString;
 use crate::attribute_parser::*;
 use crate::util::*;
 
+fn gen_set_template(source: TemplateSource) -> TokenStream {
+    match source {
+        TemplateSource::File(file) => quote! {
+            let t = include_bytes!(#file);
+            klass.set_template(t);
+        },
+        TemplateSource::Resource(resource) => quote! {
+            klass.set_template_from_resource(&#resource);
+        },
+        TemplateSource::String(template) => quote! {
+            klass.set_template(&#template);
+        },
+    }
+}
+
 fn gen_template_child_bindings(fields: &syn::Fields) -> TokenStream {
     let crate_ident = crate_ident_new();
     let attributed_fields = match parse_fields(&fields) {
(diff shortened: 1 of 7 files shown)


gtk3-macros/src/composite_template_derive.rs
```diff
@@ -45,6 +93,16 @@ pub fn impl_composite_template(input: &syn::DeriveInput) -> TokenStream {
     let name = &input.ident;
     let crate_ident = crate_ident_new();
 
+    let source = match parse_template_source(&input) {
+        Ok(v) => v,
+        Err(e) => abort_call_site!(
+            "{}: derive(CompositeTemplate) requires #[template(ui = \"/resource/path/name.ui\")]",
```
pbor, line 99: error needs fixing

sdroege (approved):
Seems fine to me, and it has lots of useful docs now :)

sdroege:
CI is unhappy though:

```
error[E0599]: no method named `init_template` found for reference `&InitializingObject<ExApplicationWindow>` in the current scope
   --> examples/src/bin/composite_template.rs:56:17
    |
56  |             obj.init_template();
    |                 ^^^^^^^^^^^^^ method not found in `&InitializingObject<ExApplicationWindow>`
    | 
   ::: /__w/gtk-rs/gtk-rs/glib/src/subclass/types.rs:383:1
    |
383 | pub struct InitializingObject<T: ObjectType>(Borrowed<T>);
    | ----------------------------------------------------------
    | |
    | doesn't satisfy `_: IsA<gtk::Widget>`
    | doesn't satisfy `_: gtk::WidgetExt`
    |
    = note: the method `init_template` exists but the following trait bounds were not satisfied:
            `InitializingObject<ExApplicationWindow>: IsA<gtk::Widget>`
            which is required by `InitializingObject<ExApplicationWindow>: gtk::WidgetExt`
            `&InitializingObject<ExApplicationWindow>: IsA<gtk::Widget>`
            which is required by `&InitializingObject<ExApplicationWindow>: gtk::WidgetExt`

```

bilelmoussaoui (approved):

ids1024:
> Also, it would be great to automagically hook into class_init instead of requiring to call template_init, but I have no idea on how to do that

I think this may be possible through some abuse of specialization... but that won't be stable any time soon.

bilelmoussaoui:
@sdroege @GuillaumeGomez can we get this one merged so that both gtk3 & gtk4 composite templates support are synced

sdroege:
Ah I thought I merged it already. Sorry!

PR #270, opened 2021-01-21
Subclassing refactoring

sdroege:
Supersedes https://github.com/gtk-rs/gtk-rs/pull/31 for now, I'll update that one to only include the `Value` changes once this one is merged.

TODO
  * [x] Implement a builder and a `fn signals() -> &'static [Signal]` method to `ObjectSubclass`

### DIFF
--- a/glib/src/subclass/types.rs
+++ b/glib/src/subclass/types.rs
@@ -338,8 +338,8 @@ pub trait ObjectSubclass: Sized + 'static {
     ///
     /// This is called after `type_init` and before the first instance
     /// of the subclass is created. Subclasses can use this to do class-
-    /// specific initialization, e.g. for installing properties or signals
-    /// on the class or calling class methods.
+    /// specific initialization, e.g. for registering signals on the class
+    /// or calling class methods.
     ///
     /// Optional
     fn class_init(_klass: &mut Self::Class) {}
--- a/glib/src/subclass/types.rs
+++ b/glib/src/subclass/types.rs
@@ -4,8 +4,7 @@
 
 use crate::object::{Cast, ObjectSubclassIs, ObjectType};
 use crate::translate::*;
-use crate::{Closure, Object, SignalFlags, StaticType, Type, Value};
-use std::fmt;
+use crate::{Closure, Object, StaticType, Type, Value};
 use std::marker;
 use std::mem;
 use std::ptr;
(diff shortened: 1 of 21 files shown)


GuillaumeGomez (approved):

bilelmoussaoui (approved):

sdroege:
@bilelmoussaoui @GuillaumeGomez Can you review the newest commit too? That's doing the whole thing for signals now. I think that's a much nicer API, and it also gives a way for calling signals from subclasses more nicely.

sdroege:
From my side this PR is ready, but before merging this I'd like to update all of GStreamer to not have it fail to build for too long. I'll probably spend Saturday or Sunday with that, it will take some time.

sdroege:
https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/613 for the GStreamer side of things. Nice cleanup.

Now I just need to port over all the 50 or so subclasses in gst-plugins-rs :sob:

glib/src/subclass/types.rs
```diff
@@ -128,6 +107,103 @@ pub unsafe trait IsImplementable<T: ObjectSubclass>: StaticType {
...
+                    interface_finalize: None,
+                    interface_data: ptr::null_mut(),
+                },
+            )
+        );
+        $types
+    });
+);
+
+interface_list_trait!(A, B, C, D, E, F, G, H, I, J, K, L, M, N, O, P, Q, R, S);
```
GuillaumeGomez, line 205: I don't understand at all what this is about... What's the point of having this implemented randomly?

sdroege: What do you mean with randomly? This is implementing the `InterfaceList` trait on tuples of size 2 to 19

GuillaumeGomez: Exactly that. Do we really need to go up to 19 for example? Isn't it possible to make implementations on a trait that the tuples could implement?

sdroege: Well we could go up to any other number you would prefer. Implementing it not separately for each tuple size would require https://github.com/rust-lang/rfcs/issues/376 and until then a macro seemed nicer than duplicating this code a dozen times.

GuillaumeGomez: Oh damn, I was sure it was merged since the time... Then there is no issue.

GuillaumeGomez (approved):

PR #271, opened 2021-01-21
glib: Add wrapper for `g_type_ensure`

ids1024:
Presumably this is needed for in Rust, like C, at least in certain cases. (Perhaps with types defined in another language, if not in Rust?)

### DIFF
--- a/glib/src/types.rs
+++ b/glib/src/types.rs
@@ -142,6 +142,13 @@ impl Type {
             }
         }
     }
+
+    #[doc(alias = "g_type_ensure")]
+    pub fn ensure(&self) {
+        unsafe {
+            gobject_ffi::g_type_ensure(self.to_glib());
+        }
+    }
 }
 
 impl fmt::Debug for Type {


ids1024:
Maybe there's no need for this in Rust if such functions in Rust are not falsely asserted to be side effect free, and the C optimization in question (mentioned in the [docs](https://developer.gnome.org/gobject/stable/gobject-Type-Information.html#g-type-ensure)) can't occur across ffi boundaries...

I just know this is a function I've used in C that lacked bindings. And it's at least somewhat descriptive rather than the apparent no-op of just calling something like `static_type()` and ignoring the return value.

sdroege (changes requested):
This function is completely unnecessary in Rust for the reasons you gave above. Having a helper function that looks closer to that might make sense though for people who don't understand these details and just try to port C code. Something like

```rust
pub trait StaticType {
    fn ensure_type() {
        Self::static_type();
    }
}
```

Calling into C makes no sense for that.

PR #274, opened 2021-01-28
Implement Eq, PartialEq & Hash for WeakRef

### LINKED ISSUE #45: Eq / Hash / etc. for glib::object::WeakRef
Would it be a bad idea for weak references to derive Eq / Hash / PartialEq / etc.as the GTK widgets do? Currently in a situation where it would be ideal if I could store a collection of weak references in a HashSet.

A6GibKm:
Fixes #45.

### DIFF
--- a/glib/src/object.rs
+++ b/glib/src/object.rs
@@ -2216,6 +2216,14 @@ glib_wrapper! {
 #[derive(Debug)]
 pub struct WeakRef<T: ObjectType>(Pin<Box<gobject_ffi::GWeakRef>>, PhantomData<*mut T>);
 
+impl<T: ObjectType> PartialEq for WeakRef<T> {
+    fn eq(&self, other: &Self) -> bool {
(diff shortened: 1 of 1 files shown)


sdroege:
From the original issue from @mmstick
> Currently in a situation where it would be ideal if I could store a collection of weak references in a HashSet.

Doing this would cause chaos easily. Once the objects are all gone, all weak references would be equal (the pointer is `NULL`) and the invariants of the hashset would be broken (all items in it are different).

I think there's something to think through a bit more here.

A6GibKm:
Would it be better by just dropping the Hash impl from the PR?

sdroege:
`BTreeSet` with `PartialOrd`/`Ord` have the same problem. And I don't know if `PartialEq`/`Eq` are allowed to behave like this. Needs to be checked.

GuillaumeGomez:

Issue #74, opened 2020-08-15
All Context functions should return a Result instead of calling expect()


federicomenaquintero:
Things have gotten to a point in librsvg where most of the incoming bugs that make the library panic are due to cairo-rs calling `expect()` on an `Err` inside the `Context` methods.  (That is, rather than panicking in librsvg itself, something in the drawing calls causes the `cr` to enter an error state.)

For example, all of the `line_to` / `move_to` / etc. can set the `cr` to an out-of-memory state if they can't allocate the new command.  In theory, they could also set an error state if they detected coordinates outside Cairo's supported range (I intend to look into that soon - Cairo crashes really easily if given coordinates outside ±2^23, and it doesn't do range checking for coordinates).

I'd like to propose that either all the Context functions return a Result, or that we remove the expect()s and make `Context.status()` public again.

One thing librsvg would like to do, for example, is to feed a path to cairo with the `move_to`, `line_to`, etc. and just check at the end if there was an error.  It would also be fine to catch a result with `?` and return as early as possible if there is an error in one of the individual commands.

Cairo contexts and surfaces have the same "lenient" error checking - they can go into an error state and ignore further API calls, but this is just to help C programs avoid having `if (cairo_something(...) != OK)` everywhere.  We have better language facilities in Rust :smiley: 

Summary: 

* I think just returning a Result instead of calling expect() would let the calling program make the correct choice of what to do upon errors.

* I think we still want a public way to get the Context's status.  One thing librsvg does is to get a `*mut cairo_t` from the C world, then wraps it in a `Context` and checks to see if the context is in error - it will g_warning() and return if so, to avoid panicking inside librsvg/cairo-rs.

pbor:
Isn't calling `status()` on every call too much?

It seems to me like we should check when calling `fill/stroke/paint/etc`, but not when building the path to draw...

bilelmoussaoui:
Would be nice to know what's the plan here as I'm very interested in seeing librsvg getting an update to the latest glib stuff so that it's usable with gtk4

federicomenaquintero:
You know, @pbor has a good point - from debugging some things in librsvg recently, it seems indeed useful to only check when something "commits" to a surface, but not on every individual drawing call.

(Right now, I'm debugging a panic where `cr.restore()`  finds out that the previous drawing caused the cr to be in error.  For librsvg, **that** point seems like a good place to figure out that the drawing "transaction" failed.)

federicomenaquintero:
[...]

PR #279, opened 2021-01-30
Cairo: make status() public

pbor:
This starts to address some of the less ambiguous parts of https://github.com/gtk-rs/gtk-rs/issues/74

### DIFF
--- a/cairo/src/context.rs
+++ b/cairo/src/context.rs
@@ -163,14 +163,14 @@ impl Context {
         unsafe { Self::from_raw_full(ffi::cairo_create(target.to_raw_none())) }
     }
 
-    pub fn save(&self) {
+    pub fn save(&self) -> Result<(), Error> {
         unsafe { ffi::cairo_save(self.0.as_ptr()) }
-        self.status().expect("Failed to save")
+        self.status()
     }
 
-    pub fn restore(&self) {
+    pub fn restore(&self) -> Result<(), Error> {
         unsafe { ffi::cairo_restore(self.0.as_ptr()) }
-        self.status().expect("Failed to restore")
+        self.status()
     }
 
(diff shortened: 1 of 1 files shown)


cairo/src/context.rs
```diff
@@ -815,9 +815,9 @@ mod tests {
     }
 
     #[test]
-    fn clip_rectange() {
+    fn clip_rectangle() {
```
GuillaumeGomez, line 818: Nice typo fix! :)

cairo/src/context.rs
```diff
@@ -815,9 +815,9 @@ mod tests {
...
         let ctx = create_ctx();
-        let rect = ctx.copy_clip_rectangle_list();
+        let rect = ctx.copy_clip_rectangle_list()?;
```
GuillaumeGomez, line 820: Since the function doesn't return a `Result`, I think it'll fail. Better use a `expect` here I think.

GuillaumeGomez (approved):

GuillaumeGomez:
Apart from the fmt failure, looks good to me!

cairo/src/context.rs
pbor: Indeed that did not even compile, I forgot to amend

pbor:
> Apart from the fmt failure, looks good to me!

Fixed. thanks for the quick review!

bilelmoussaoui (approved):

GuillaumeGomez:
Please remove the update on the `gir` submodule too. :)

cairo/src/context.rs
pbor: > Please remove the update on the `gir` submodule too. :)

I wonder how many silly mistakes I can still make in a row :(
Hopefully ok now

PR #284, opened 2021-02-01
Mark 'key' parameter nullable for KeyFile's get_comment()

### LINKED ISSUE #282: KeyFile get_comment() doesn't allow None for key argument like set_comment() does
Seems like it should allow both key and group_name to be None just like set_comment.

For g_key_file_get_comment() the docs say: "If both key and group_name are NULL, then comment will be read from above the first group in the file."

epakai:
fixes #282

### DIFF
--- a/glib/Gir.toml
+++ b/glib/Gir.toml
@@ -523,6 +523,11 @@ status = "generate"
     name = "get_locale_string_list"
     # can return an error but still a value to be freed
     manual = true
+    [[object.function]]
+    name = "get_comment"
+        [[object.function.parameter]]
+        name = "key"
+        nullable = true
 
     # Fixed in GLib 2.66.2
     [[object.function]]


GuillaumeGomez:
You need to regen the code too, otherwise nothing changes. To do so, run `python3 generator.py`.

epakai:
Running the generator seemed to change many things in other modules. Perhaps due to my local environment. I'm not sure about creating a clean pull request from this.

sdroege:
I've created a new PR with that here: https://github.com/gtk-rs/gtk-rs/pull/286

OOC, what local changes did you get when running `generator.py`? That should be independent of the local setup, so if you get changes there that seems like something we should investigate :)

epakai:
Sorry, I reverted those changes. I ran it again on two machines and just saw minimal changes. It probably has nothing to do with my environment.

I remember generator.py prompting me about updating modules and I said yes. There were changes to the commit ids in lines like:
`Generated by gir (https://github.com/gtk-rs/gir @ 21213ed)`
and I remember an extra use line for OsString that seemed out of place for the change I made.

sdroege:
Ok :) Let's keep that for next time it happens then

PR #288, opened 2021-02-03
gio: Add DBusActionGroup

sophie-h:


GuillaumeGomez:
What is it used for?

sophie-h:
> What is it used for?

I think it's the only way to call a GAction exported from an other app.

GuillaumeGomez:
Weird that it's not being used anywhere else... Well, looks good to me otherwise.

GuillaumeGomez (approved):

PR #290, opened 2021-02-04
Keep glib::clone spans

### LINKED ISSUE #287: clone! macro makes error messages cryptic
Hello,

when using the `clone!` macro rustc doesn't report the line where the errror occured correctly and instead just points to the entire macro invocation. This forces users of the macro to guess where the error occured, which isn't trivial when the function `clone!` encapsulates is longer than a few lines.

E.g. when I forget to put a semicolon somewhere the error message looks like this:

```
error: expected one of `.`, `;`, `?`, or an operator, found `self_`
   --> src/widgets/sync_list_box.rs:133:40
    |
133 |                   .connect_row_activated(clone!(@weak obj => move |list_box, row| {
    |  ________________________________________^
134 | |                     let self_ = imp::HealthSyncListBox::from_instance(&obj);
135 | |                     if (row == &self_.google_fit_start_sync_row.get()) {
136 | |                         self_.google_fit_stack.set_visible(true);
...   |
180 | |                     }
181 | |                 }));
    | |__________________^ expected one of `.`, `;`, `?`, or an operator
    |
    = note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)

```

So I have to check some 50 lines to know where I forgot the semicolon.

See https://github.com/Cogitri/clone-reproducer for a reproducer:

```
error[E0599]: no method named `qiut` found for struct `gtk4::Application` in the current scope
   --> src/bin/entry_completion.rs:101:9
    |
101 | /         glib::clone!(@weak application => move |_action, _parameter| {
102 | |             application.qiut();
103 | |         }),
    | |__________^ method not found in `gtk4::Application`
    |
    = note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)
```

In this case the error message seems OK because it's only a one line function, but ideally it'd point to only line 102 and not 101/103.

GuillaumeGomez:
Fixes #287

### DIFF
--- a/glib-macros/src/clone.rs
+++ b/glib-macros/src/clone.rs
@@ -2,7 +2,7 @@
 
 use crate::utils::crate_ident_new;
 use proc_macro::token_stream::IntoIter as ProcIter;
-use proc_macro::{Delimiter, Group, TokenStream, TokenTree};
+use proc_macro::{Delimiter, Group, Ident, Punct, Spacing, Span, TokenStream, TokenTree};
 use std::iter::Peekable;
 
 #[derive(Clone, Copy, Debug)]
@@ -452,13 +452,13 @@ fn check_before_closure(parts: &mut Peekable<ProcIter>) {
     }
 }
 
-fn get_closure(parts: &mut Peekable<ProcIter>) -> String {
-    let mut ret = String::new();
+fn get_closure(parts: &mut Peekable<ProcIter>) -> Vec<TokenTree> {
+    let mut ret = Vec::new();
 
     loop {
         match parts.next() {
             Some(TokenTree::Punct(p)) if p.to_string() == "|" => break,
-            Some(x) => ret.push_str(&x.to_string()),
+            Some(x) => ret.push(x),
             None => panic!("Unexpected end 3"),
         }
     }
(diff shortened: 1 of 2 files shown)


glib-macros/src/clone.rs
```diff
@@ -539,22 +539,66 @@ pub(crate) fn clone_inner(item: TokenStream) -> TokenStream {
     let return_kind = parse_return_kind(&mut parts);
     check_before_closure(&mut parts);
     let closure = get_closure(&mut parts);
-    let body = tokens_to_string(parts);
-
-    let x = format!(
-        "{{\n{}\nmove |{}| {{\n{}\nlet ____ret = {{ {} }};\n____ret\n}}\n}}",
```
sdroege, line 545: How did we ever think that was a good idea?

GuillaumeGomez: Well, it makes the reading easier?

GuillaumeGomez:
@pbor: Any idea what's wrong with Windows CI here?

sdroege:
```
 error: returning the result of a `let` binding from a block
   --> glib/tests/clone.rs:478:9
    |
477 |         let y = 19;
    |         ----------- unnecessary `let` binding
478 |         y
    |         ^
```

@GuillaumeGomez

GuillaumeGomez:
Ah right, forgot to remove this one.

glib/tests/clone.rs
```diff
@@ -400,11 +400,15 @@ fn test_clone_macro_double_rename() {
     closure(0i8); // to prevent compiler error for unknown `x` type.
     let _ = clone!(@weak v, @weak w as _x => @default-return true, move || false);
 
-    let closure = clone!(@strong v as _x, @strong w => @default-return true, move |_| { false });
+    let closure = clone!(@strong v as _x, @strong w => @default-return true, move |_| {
+        false
```
sdroege, line 404: Ah does rustfmt understand the body of the `clone!` macro now because we don't go through a string?

GuillaumeGomez: No idea, it was complaining about the unneeded parens and I wanted to them.

sdroege (approved):

PR #292, opened 2021-02-06
glib: rename TypeData::interface_data to class_data

bilelmoussaoui:
this also converts it to a HashMap instead of a Vec

### DIFF
--- a/glib/src/subclass/types.rs
+++ b/glib/src/subclass/types.rs
@@ -5,10 +5,10 @@
 use crate::object::{Cast, ObjectSubclassIs, ObjectType};
 use crate::translate::*;
 use crate::{Closure, Object, StaticType, Type, Value};
-use std::collections::HashMap;
 use std::marker;
 use std::mem;
 use std::ptr;
+use std::{any::Any, collections::HashMap};
 
 /// A newly registered `glib::Type` that is currently still being initialized.
 ///
@@ -212,7 +212,7 @@ pub struct TypeData {
     #[doc(hidden)]
     pub parent_class: ffi::gpointer,
     #[doc(hidden)]
-    pub class_data: *const HashMap<ffi::GType, ffi::gpointer>,
+    pub class_data: Option<ptr::NonNull<HashMap<ffi::GType, Box<dyn Any + Send + Sync>>>>,
     #[doc(hidden)]
     pub private_offset: isize,
 }
@@ -237,30 +237,49 @@ impl TypeData {
     /// Returns a pointer to the class implementation specific data.
     ///
     /// This is used for class implementations to store additional data.
-    pub fn get_class_data(&self, type_: ffi::GType) -> ffi::gpointer {
+    pub fn get_class_data<T: Any + Send + Sync + 'static>(&self, type_: ffi::GType) -> Option<&T> {
         unsafe {
-            if self.class_data.is_null() {
-                return ptr::null_mut();
+            match self.class_data {
+                None => None,
+                Some(data) => data.as_ref().get(&type_).and_then(|ptr| ptr.downcast_ref()),
             }
-
-            *(*self.class_data).get(&type_).unwrap_or(&ptr::null_mut())
         }
     }
 
-    pub unsafe fn set(&mut self, type_: ffi::GType, data: ffi::gpointer) {
-        if self.class_data.is_null() {
-            self.class_data = Box::into_raw(Box::new(HashMap::new()));
+    /// Gets a mutable reference of the class implementation specific data
+    ///
+    /// # Safety
+    ///
+    /// This can only be executed before initializing the type data
+    pub unsafe fn get_class_data_mut<T: Any + Send + Sync>(
+        &mut self,
+        type_: ffi::GType,
+    ) -> Option<&mut T> {
+        match self.class_data.as_mut() {
+            None => None,
+            Some(map) => map.as_mut().get_mut(&type_).and_then(|v| v.downcast_mut()),
         }
-        if self.class_data.as_ref().unwrap().get(&type_).is_some() {
-            panic!("The class_data already contains a key for {}", type_);
-        }
-        (*(self.class_data as *mut HashMap<ffi::GType, ffi::gpointer>)).insert(type_, data);
     }
 
-    pub unsafe fn get_mut(&mut self, type_: ffi::GType) -> ffi::gpointer {
-        *(*(self.class_data as *mut HashMap<ffi::GType, ffi::gpointer>))
-            .get_mut(&type_)
-            .unwrap_or(&mut ptr::null_mut())
+    /// Sets class specific implementation data
+    ///
+    /// # Safety
+    ///
+    /// Note the set_class_data can only used before type initialization
+    ///
+    /// # Panics
+    ///
+    /// If the class_data contains a data for the specified `type_`
+    pub unsafe fn set_class_data<T: Any + Send + Sync>(&mut self, type_: ffi::GType, data: T) {
+        if self.class_data.is_none() {
+            self.class_data = ptr::NonNull::new(Box::into_raw(Box::new(HashMap::new())));
+        }
+        if let Some(class_data) = self.class_data.as_mut() {
+            if class_data.as_ref().get(&type_).is_some() {
+                panic!("The class_data already contains a key for {}", type_);
+            }
+            class_data.as_mut().insert(type_, Box::new(data));
+        }
     }
 
     /// Returns the offset of the private struct in bytes relative to the
@@ -280,7 +299,7 @@ macro_rules! object_subclass {
             static mut DATA: $crate::subclass::TypeData = $crate::subclass::TypeData {
                 type_: $crate::Type::Invalid,
                 parent_class: std::ptr::null_mut(),
-                class_data: std::ptr::null_mut(),
+                class_data: None,
                 private_offset: 0,
             };
 
--- a/glib/src/subclass/types.rs
+++ b/glib/src/subclass/types.rs
@@ -243,14 +243,24 @@ impl TypeData {
                 return ptr::null_mut();
             }
 
-            for (t, p) in &(*self.class_data) {
-                if t == &type_ {
-                    return *p;
-                }
-            }
-
-            ptr::null_mut()
+            *(*self.class_data).get(&type_).unwrap_or(&ptr::null_mut())
         }
+    }
+
+    pub unsafe fn set(&mut self, type_: ffi::GType, data: ffi::gpointer) {
+        if self.class_data.is_null() {
+            self.class_data = Box::into_raw(Box::new(HashMap::new()));
+        }
+        if self.class_data.as_ref().unwrap().get(&type_).is_some() {
+            panic!("The class_data already contains a key for {}", type_);
+        }
+        (*(self.class_data as *mut HashMap<ffi::GType, ffi::gpointer>)).insert(type_, data);
+    }
+
+    pub unsafe fn get_mut(&mut self, type_: ffi::GType) -> ffi::gpointer {
+        *(*(self.class_data as *mut HashMap<ffi::GType, ffi::gpointer>))
+            .get_mut(&type_)
+            .unwrap_or(&mut ptr::null_mut())
     }
 
     /// Returns the offset of the private struct in bytes relative to the
--- a/glib/src/subclass/types.rs
+++ b/glib/src/subclass/types.rs
@@ -5,6 +5,7 @@
 use crate::object::{Cast, ObjectSubclassIs, ObjectType};
 use crate::translate::*;
 use crate::{Closure, Object, StaticType, Type, Value};
+use std::collections::HashMap;
 use std::marker;
 use std::mem;
 use std::ptr;
@@ -211,7 +212,7 @@ pub struct TypeData {
     #[doc(hidden)]
     pub parent_class: ffi::gpointer,
     #[doc(hidden)]
-    pub interface_data: *const Vec<(ffi::GType, ffi::gpointer)>,
+    pub class_data: *const HashMap<ffi::GType, ffi::gpointer>,
     #[doc(hidden)]
     pub private_offset: isize,
 }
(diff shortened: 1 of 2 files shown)


glib/src/subclass/types.rs
```diff
@@ -233,18 +234,18 @@ impl TypeData {
...
         unsafe {
-            if self.interface_data.is_null() {
+            if self.class_data.is_null() {
                 return ptr::null_mut();
             }
 
-            for &(t, p) in &(*self.interface_data) {
-                if t == type_ {
-                    return p;
+            for (t, p) in &(*self.class_data) {
```
sdroege, line 246: How about

```suggestion
            (*self.class_data).get(&type_).unwrap_or(ptr::null_mut())
```

bilelmoussaoui: Ah yes, that's definitely better than what I did.

bilelmoussaoui: Although, I wonder if we should have a helper to add something to class_data the same way we have a getter?

sdroege: I think that would be very useful, yes. A setter that allocates the hashmap if not existing, and that then sets the value (and panics if a value was already set!). That setter must be unsafe and must only be called during type initialization.

You could also add a function like the `get()` above that is unsafe and `get_mut()`.

Aaaaand. You can also make the whole API slightly more safe and less ugly at no meaningful cost at runtime after initialization by having it store `Box<dyn Any + Send + Sync>` instead of working with raw pointers for the hashmap values.

bilelmoussaoui: > I think that would be very useful, yes. A setter that allocates the hashmap if not existing, and that then sets the value (and panics if a value was already set!). That setter must be unsafe and must only be called during type initialization.

Done, I'm not sure how can I ensure it's only called at type initialization ? Or that should be just in the safety section of the function documentation ?

> 
> You could also add a function like the `get()` above that is unsafe and `get_mut()`.

What would be the difference between get() & get_class_data? make the former return a specific type instead of a pointer?

sdroege: > Done, I'm not sure how can I ensure it's only called at type initialization ? Or that should be just in the safety section of the function documentation ?

Yes in the docs, and that's why it's unsafe :)

> What would be the difference between get() & get_class_data? make the former return a specific type instead of a pointer?

I was just lazy to type. I meant `fn get_class_data() -> &dyn Any + Send + Sync` and `unsafe fn get_class_data_mut() -> &mut dyn Any + Send + Sync`

glib/src/subclass/types.rs
```diff
@@ -211,7 +212,7 @@ pub struct TypeData {
     #[doc(hidden)]
     pub parent_class: ffi::gpointer,
     #[doc(hidden)]
-    pub interface_data: *const Vec<(ffi::GType, ffi::gpointer)>,
+    pub class_data: *const HashMap<ffi::GType, ffi::gpointer>,
```
sdroege, line 215:
```suggestion
    pub class_data: *const HashMap<ffi::GType, Box<dyn Any + Send + Sync>,
```

glib/src/subclass/types.rs
```diff
@@ -233,23 +234,33 @@ impl TypeData {
         self.parent_class
     }
 
-    /// Returns a pointer to the interface implementation specific data.
+    /// Returns a pointer to the class implementation specific data.
     ///
-    /// This is used for interface implementations to store additional data.
-    pub fn get_interface_data(&self, type_: ffi::GType) -> ffi::gpointer {
+    /// This is used for class implementations to store additional data.
+    pub fn get_class_data(&self, type_: ffi::GType) -> ffi::gpointer {
```
sdroege, line 240:
```suggestion
    pub fn get_class_data(&self, type_: ffi::GType) -> Option<&dyn Any + Send + Sync> {
```

glib/src/subclass/types.rs
```diff
@@ -233,23 +234,33 @@ impl TypeData {
...
-                if t == type_ {
-                    return p;
-                }
-            }
+            *(*self.class_data).get(&type_).unwrap_or(&ptr::null_mut())
+        }
+    }
 
-            ptr::null_mut()
+    pub unsafe fn set(&mut self, type_: ffi::GType, data: ffi::gpointer) {
```
sdroege, line 250: Docs (especially for unsafety) plus

```suggestion
    pub unsafe fn set_class_data<T: Any + Send + Sync + 'static>(&mut self, type_: ffi::GType, data: T) {
```

glib/src/subclass/types.rs
sdroege: Or even

```suggestion
    pub fn get_class_data<T: Any + Send + Sync + 'static>(&self, type_: ffi::GType) -> Option<&T> {
```

And have it also panic on type mismatch (because really must not happen!)

glib/src/subclass/types.rs
sdroege: Could also make it an `Option<ptr::NonNull<_>>` for cleanness

glib/src/subclass/types.rs
```diff
@@ -233,23 +234,33 @@ impl TypeData {
...
+        if self.class_data.is_null() {
+            self.class_data = Box::into_raw(Box::new(HashMap::new()));
         }
+        if self.class_data.as_ref().unwrap().get(&type_).is_some() {
+            panic!("The class_data already contains a key for {}", type_);
+        }
+        (*(self.class_data as *mut HashMap<ffi::GType, ffi::gpointer>)).insert(type_, data);
+    }
+
+    pub unsafe fn get_mut(&mut self, type_: ffi::GType) -> ffi::gpointer {
```
sdroege, line 260:
```suggestion
    pub unsafe fn get_class_data_mut(&mut self, type_: ffi::GType) -> ffi::gpointer {
```

And all as above for the normal getter

sdroege (changes requested):
Some more changes, let's make it nice while we're at it :)

glib/src/subclass/types.rs
bilelmoussaoui: Thanks

bilelmoussaoui:
> Some more changes, let's make it nice while we're at it :)

Thanks for the review! I will send a PR to update the use case you have in gstreamer and see how that will look with the new API

sdroege:
> Thanks for the review! I will send a PR to update the use case you have in gstreamer and see how that will look with the new API

It would still be a `Vec<*const c_char>` or `Box<[*const c_char]>` btw, I don't think you can nicely get rid of that raw pointer there without additional overhead.

glib/src/subclass/types.rs
bilelmoussaoui: > Could also make it an `Option<ptr::NonNull<_>>` for cleanness

It doesn't seem to like `ptr::NonNull<_>`, especially the `_` part : `the type placeholder `_` is not allowed within types on item signatures`

glib/src/subclass/types.rs
bilelmoussaoui: > And have it also panic on type mismatch

Not sure how can I do something like this

bilelmoussaoui:
> It would still be a `Vec<*const c_char>` or `Box<[*const c_char]>` btw, I don't think you can nicely get rid of that raw pointer there without additional overhead.

Do we need the `Sync` constraint everywhere? because otherwise we can't get the gstreamer stuff to compile with the changes on this merge request

```
  --> gstreamer/src/subclass/uri_handler.rs:52:10
   |
52 |         .get_class_data::<*const *const libc::c_char>(URIHandler::static_type().to_glib())
   |          ^^^^^^^^^^^^^^ `*const *const i8` cannot be shared between threads safely
   |
```

glib/src/subclass/types.rs
sdroege: Yes I was too lazy to type it out again. I meant `Option<ptr::NonNull<HashMap<ffi::GType, Box<dyn Any + Send + Sync>>>`

glib/src/subclass/types.rs
sdroege: https://doc.rust-lang.org/nightly/std/any/trait.Any.html#method.is-2 for checking the type, https://doc.rust-lang.org/nightly/std/any/trait.Any.html#method.downcast_ref-2 and the mutable variant for getting the value out

sdroege:
> Do we need the `Sync` constraint everywhere? because otherwise we can't get the gstreamer stuff to compile with the changes on this merge request

Unfortunately yes. The trick here would be to make a transparent wrapper type around the raw pointer and implement the two traits on that.

glib/src/subclass/types.rs
```diff
@@ -233,23 +234,48 @@ impl TypeData {
...
+            if self.class_data.is_null() {
+                return None;
             }
 
-            for &(t, p) in &(*self.interface_data) {
-                if t == type_ {
-                    return p;
-                }
-            }
+            (*(self.class_data as *const HashMap<ffi::GType, T>)).get(&type_)
```
sdroege, line 246: That won't work. You need to `get()` it with the `Box<_>` type from above and then call `downcast_ref()` on it to get the `&T` out of it.

sdroege: Also everywhere else. You need to handle the hashmap with its original type and make use of the `Any` API

glib/src/subclass/types.rs
```diff
@@ -233,23 +234,60 @@ impl TypeData {
...
+    /// This is used for class implementations to store additional data.
+    pub fn get_class_data<T: Any + Send + Sync + 'static>(&self, type_: ffi::GType) -> Option<&T> {
         unsafe {
-            if self.interface_data.is_null() {
-                return ptr::null_mut();
+            if self.class_data.is_none() {
+                return None;
             }
+            let map = self.class_data.unwrap();
+            map.as_ref().get(&type_).map(|v| v.downcast_ref()).flatten()
```
bilelmoussaoui, line 246: This doesn't compile due to 
```

   --> glib/src/subclass/types.rs:246:13
    |
246 |             map.as_ref().get(&type_).map(|v| v.downcast_ref()).flatten()
    |             ---^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |             |
    |             returns a value referencing data owned by the current function
```
not sure how to handle this?

sdroege: Use a `match` instead of `as_ref()` and `map()` and `flatten()`. Also your `map(f).flatten()` is basically `and_then(f)` :)

glib/src/subclass/types.rs
```diff
@@ -233,23 +234,60 @@ impl TypeData {
...
-    pub fn get_interface_data(&self, type_: ffi::GType) -> ffi::gpointer {
+    /// This is used for class implementations to store additional data.
+    pub fn get_class_data<T: Any + Send + Sync + 'static>(&self, type_: ffi::GType) -> Option<&T> {
         unsafe {
-            if self.interface_data.is_null() {
-                return ptr::null_mut();
+            if self.class_data.is_none() {
+                return None;
             }
+            let map = self.class_data.unwrap();
```
sdroege, line 245: You can get around the `unwrap()` here by either matching above, or doing

```rust
let class_data = self.class_data?;
```

glib/src/subclass/types.rs
```diff
@@ -233,23 +234,60 @@ impl TypeData {
...
+        type_: ffi::GType,
+    ) -> Option<&mut T> {
+        self.class_data
+            .as_mut()
+            .map(|hashmap| {
+                hashmap
+                    .as_mut()
+                    .get_mut(&type_)
+                    .map(|v| v.downcast_mut())
+                    .flatten()
```
sdroege, line 266:
```suggestion
                    .and_then(|v| v.downcast_mut())
```

glib/src/subclass/types.rs
```diff
@@ -233,23 +234,60 @@ impl TypeData {
...
+    /// If the class_data contains a data for the specified `type_`
+    pub unsafe fn set_class_data<T: Any + Send + Sync>(&mut self, type_: ffi::GType, data: T) {
+        if self.class_data.is_none() {
+            self.class_data = ptr::NonNull::new(Box::into_raw(Box::new(HashMap::new())));
+        }
+        if self.class_data.unwrap().as_ref().get(&type_).is_some() {
+            panic!("The class_data already contains a key for {}", type_);
         }
+        self.class_data
+            .unwrap()
```
sdroege, line 288: Unnecessary unwrap here and 3 lines above

glib/src/subclass/types.rs
```diff
@@ -233,23 +234,60 @@ impl TypeData {
...
+        self.class_data
+            .as_mut()
+            .map(|hashmap| {
+                hashmap
+                    .as_mut()
+                    .get_mut(&type_)
+                    .map(|v| v.downcast_mut())
+                    .flatten()
+            })
+            .flatten()
```
sdroege, line 268: Also this `flatten()` can be made an `and_then()` (and you can remove one level of nesting). But the whole thing would probably easier with match and/or `?`

glib/src/subclass/types.rs
bilelmoussaoui: The as_ref here comes from NonNull and not Option

bilelmoussaoui: Thanks for the tip on and_then, wasn't aware of that :)

glib/src/subclass/types.rs
bilelmoussaoui: Am I supposed to map/and_then here as well instead of unwrap?

sdroege: That or you could match or use `?`

sdroege: Just do it consistently in all 3 functions :)

bilelmoussaoui: Updated to make it a bit nicer, We can't make use of `?` and match doesn't make much sense I guess as we check few lines above that there's something in self.class_data first before attempting to read it's content

glib/src/subclass/types.rs
bilelmoussaoui: This still doesn't compile btw :P

sdroege (approved):
Good from my side now. @bilelmoussaoui ?

bilelmoussaoui:
Looks good to me as well

PR #294, opened 2021-02-08
gio add Icon & LoadableIcon subclassing support

bilelmoussaoui:
can be useful for serializing your custom icon & sending it over dbus :-) I'm not super sure about the whole load_async & load_finish ones though.

### DIFF
--- a/gio/src/subclass/icon.rs
+++ b/gio/src/subclass/icon.rs
@@ -1,0 +1,101 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use crate::Icon;
+use glib::subclass::prelude::*;
+use glib::translate::*;
+use glib::{Cast, IsA};
+
+pub trait IconImpl: ObjectImpl {
+    fn hash(&self, icon: &Self::Type) -> u32;
+    fn equal(&self, icon: &Self::Type, other_icon: &Icon) -> bool;
+    fn to_tokens(&self, icon: &Self::Type) -> Option<(Vec<String>, i32)>;
+    fn from_tokens(tokens: Vec<String>, version: i32) -> Result<Self::Type, glib::Error>;
+    fn serialize(&self, icon: &Self::Type) -> Option<glib::Variant>;
+}
+
+unsafe impl<T: IconImpl> IsImplementable<T> for Icon
+where
+    <T as ObjectSubclass>::Type: IsA<glib::Object>,
+{
+    unsafe extern "C" fn interface_init(
+        iface: glib::ffi::gpointer,
+        _iface_data: glib::ffi::gpointer,
+    ) {
+        let icon_iface = &mut *(iface as *mut ffi::GIconIface);
+
+        icon_iface.hash = Some(icon_hash::<T>);
+        icon_iface.equal = Some(icon_equal::<T>);
+        icon_iface.to_tokens = Some(icon_to_tokens::<T>);
+        icon_iface.from_tokens = Some(icon_from_tokens::<T>);
+        icon_iface.serialize = Some(icon_serialize::<T>);
+    }
+}
+
+unsafe extern "C" fn icon_hash<T: IconImpl>(icon: *mut ffi::GIcon) -> u32 {
+    let instance = &*(icon as *mut T::Instance);
+    let imp = instance.get_impl();
+    let wrap = from_glib_borrow::<_, Icon>(icon);
+
+    imp.hash(wrap.unsafe_cast_ref())
+}
+
+unsafe extern "C" fn icon_equal<T: IconImpl>(
+    icon1: *mut ffi::GIcon,
+    icon2: *mut ffi::GIcon,
+) -> glib::ffi::gboolean {
+    let instance = &*(icon1 as *mut T::Instance);
+    let imp = instance.get_impl();
+
+    imp.equal(
+        from_glib_borrow::<_, Icon>(icon1).unsafe_cast_ref(),
+        from_glib_borrow::<_, Icon>(icon2).unsafe_cast_ref(),
+    )
+    .to_glib()
+}
+
+unsafe extern "C" fn icon_to_tokens<T: IconImpl>(
+    icon: *mut ffi::GIcon,
+    tokensptr: *mut glib::ffi::GPtrArray,
+    version: *mut libc::c_int,
+) -> glib::ffi::gboolean {
+    let instance = &*(icon as *mut T::Instance);
+    let imp = instance.get_impl();
+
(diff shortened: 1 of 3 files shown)


gio/src/subclass/icon.rs
```diff
@@ -0,0 +1,101 @@
...
+    tokensptr: *mut glib::ffi::GPtrArray,
+    version: *mut libc::c_int,
+) -> glib::ffi::gboolean {
+    let instance = &*(icon as *mut T::Instance);
+    let imp = instance.get_impl();
+
+    let tokens_version = imp.to_tokens(from_glib_borrow::<_, Icon>(icon).unsafe_cast_ref());
+    if let Some((tokens, tversion)) = tokens_version {
+        *version = tversion;
+        *tokensptr = tokens.to_glib_full();
```
bilelmoussaoui, line 67: Not sure how can I convert my Vec<String> to a `gptrarray`. Are there some traits I should implement for `gptarray` to allow such conversions?

GuillaumeGomez: I'm not if we have that yet. Isn't something like this working:

```rust
let tokens = tokens.into_iter().map(|t| t.to_glib_full()).collect::<Vec<_>>();
tokens.to_glib_full()
```

?

sdroege: That should work in theory, there are lots of trait impls in `glib/src/translate.rs` for `GPtrArray`

sdroege: If that doesn't work, please paste the compiler error so we can find out which impl is missing :)

gio/src/subclass/icon.rs
```diff
@@ -0,0 +1,101 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use crate::Icon;
+use glib::subclass::prelude::*;
+use glib::translate::*;
+use glib::{Cast, IsA};
+
+pub trait IconImpl: ObjectImpl {
+    fn hash(&self, icon: &Self::Type) -> u32;
+    fn equal(&self, icon: &Self::Type, other_icon: &Icon) -> bool;
```
sdroege, line 10: Or require `Hash`, `PartialEq` and `Eq` to be implemented?

gio/src/subclass/icon.rs
```diff
@@ -0,0 +1,101 @@
...
+    fn to_tokens(&self, icon: &Self::Type) -> Option<(Vec<String>, i32)>;
```
sdroege, line 11: Is that `i32` an icon size? I think we have an enum for that?

sdroege: Ah no, it's a version Nevermind

gio/src/subclass/icon.rs
```diff
@@ -0,0 +1,101 @@
...
+    }
+}
+
+unsafe extern "C" fn icon_from_tokens<T: IconImpl>(
+    tokensptr: *mut *mut libc::c_char,
+    num_tokens: i32,
+    version: i32,
+    error: *mut *mut glib::ffi::GError,
+) -> *mut ffi::GIcon {
+    let tokens: Vec<String> = String::from_glib_none_num_as_vec(tokensptr, num_tokens as usize);
```
sdroege, line 81: You could also make it a `Vec<&str>` but probably doesn't matter much

gio/src/subclass/icon.rs
bilelmoussaoui: the i32 here is the "version" that should be passed when using from_tokens

gio/src/subclass/icon.rs
bilelmoussaoui: Much better indeed.

gio/src/subclass/icon.rs
```diff
@@ -0,0 +1,101 @@
...
+use crate::Icon;
+use glib::subclass::prelude::*;
+use glib::translate::*;
+use glib::{Cast, IsA};
+
+pub trait IconImpl: ObjectImpl {
+    fn hash(&self, icon: &Self::Type) -> u32;
+    fn equal(&self, icon: &Self::Type, other_icon: &Icon) -> bool;
+    fn to_tokens(&self, icon: &Self::Type) -> Option<(Vec<String>, i32)>;
+    fn from_tokens(tokens: Vec<String>, version: i32) -> Result<Self::Type, glib::Error>;
```
sdroege, line 12:
```suggestion
    fn from_tokens(tokens: &[String], version: i32) -> Result<Self::Type, glib::Error>;
```

(or even better `&[&str]` see below)

gio/src/subclass/loadable_icon.rs
```diff
@@ -0,0 +1,134 @@
...
+pub struct AsyncReadyCallback {
+    pub callback: ffi::GAsyncReadyCallback,
+    pub user_data: glib::ffi::gpointer,
+}
+
+pub trait LoadableIconImpl: IconImpl {
+    fn load(
+        &self,
+        icon: &Self::Type,
+        size: i32,
```
sdroege, line 19: There's IIRC an enum for the icon size

gio/src/subclass/loadable_icon.rs
```diff
@@ -0,0 +1,134 @@
...
+        cancellable: Option<&Cancellable>,
+    ) -> Result<(InputStream, Option<GString>), glib::Error>;
+    fn load_async(
+        &self,
+        icon: &Self::Type,
+        size: i32,
+        cancellable: Option<&Cancellable>,
+        callback: AsyncReadyCallback,
```
sdroege, line 27: Maybe this can all be done a bit nicer with the new `GTask` bindings? I don't think it's actually usable the way you currently do it here

gio/src/subclass/loadable_icon.rs
```diff
@@ -0,0 +1,134 @@
...
+        Ok((stream, some_type)) => {
+            if let Some(type_) = some_type {
+                *typeptr = type_.to_glib_none().0;
+            }
+
+            stream.to_glib_full()
+        }
+        Err(err) => {
+            *errorptr = err.to_glib_full() as *mut _;
+            *typeptr = std::ptr::null_mut();
```
sdroege, line 76: The pointers are `optional`, so they might be `NULL`. Please also check for the others in both files

gio/src/subclass/loadable_icon.rs
bilelmoussaoui: The enum is on gtk though :/

sdroege: Oh no. Maybe it should be moved?

GuillaumeGomez:

PR #302, opened 2021-02-15
SignalId: add missing methods

bilelmoussaoui:
Depends on https://github.com/gtk-rs/gir/pull/1060


glib/src/subclass/signal.rs
```diff
@@ -68,9 +68,103 @@ impl fmt::Debug for SignalInvocationHint {
...
+        }
+    }
+
+    pub fn id(&self) -> u32 {
+        self.1
+    }
+
+    pub fn type_(&self) -> Type {
+        self.0
+    }
```
bilelmoussaoui, line 147: Not very sure if it make sense to make those public, maybe just pub(crate) ?

GuillaumeGomez: Is this an issue to have them public? What problems could it cause?

glib/src/subclass/signal.rs
```diff
@@ -68,9 +68,103 @@ impl fmt::Debug for SignalInvocationHint {
...
+            .field("flags", &self.flags())
+            .field("return_type", &self.return_type())
+            .field("param_types", &self.param_types())
+            .finish()
+    }
+}
 /// Signal ID.
 #[derive(Debug, Clone, Copy)]
-pub struct SignalId(pub(super) Type, pub(super) u32);
+pub struct SignalId(Type, u32);
```
sdroege, line 121:
```suggestion
pub struct SignalId(Type, NonZeroU32);
```

Could be used

glib/src/subclass/signal.rs
sdroege: Seems fine to have that public but should probably get some docs

glib/src/subclass/signal.rs
```diff
@@ -68,9 +68,103 @@ impl fmt::Debug for SignalInvocationHint {
...
+
+impl SignalId {
+    unsafe fn new(type_: Type, id: u32) -> Self {
```
sdroege, line 124: `pub`?

sdroege, line 124: Also as the type is only for validation, maybe we shouldn't include it? OTOH checking the type in `emit()` wastes the performance advantage of using the signal id over the name so...

glib/src/subclass/signal.rs
```diff
@@ -68,9 +68,103 @@ impl fmt::Debug for SignalInvocationHint {
...
+        self.1
+    }
+
+    pub fn type_(&self) -> Type {
+        self.0
+    }
+
+    /// Queries more in-depth information about the current signal
+    #[doc(alias = "g_signal_query")]
+    pub fn query(&self) -> SignalQuery {
```
sdroege, line 151:
```suggestion
    pub fn query(&self) -> &SignalQuery {
```

glib/src/subclass/signal.rs
```diff
@@ -68,9 +68,103 @@ impl fmt::Debug for SignalInvocationHint {
...
+        unsafe {
+            let query_ptr = std::ptr::null_mut();
+            gobject_ffi::g_signal_query(self.id(), query_ptr);
+            SignalQuery(*query_ptr)
+        }
+    }
+
+    /// Find the signal name
+    #[doc(alias = "g_signal_name")]
+    pub fn name(&self) -> Option<crate::GString> {
```
sdroege, line 161:
```suggestion
    pub fn name(&self) -> Option<&'static str> {
```

glib/src/subclass/signal.rs
```diff
@@ -68,9 +68,103 @@ impl fmt::Debug for SignalInvocationHint {
     }
 }
 
+/// In-depth information of a specific signal
+#[repr(transparent)]
+pub struct SignalQuery(gobject_ffi::GSignalQuery);
+
+impl SignalQuery {
+    pub fn signal_name(&self) -> crate::GString {
```
sdroege, line 76:
```suggestion
    pub fn signal_name<'a>(&self) -> &'a str {
```

(also below instead of `'static`)

glib/src/subclass/signal.rs
sdroege: Ah nevermind, this doesn't work of course :)

glib/src/subclass/signal.rs
```diff
@@ -68,9 +68,103 @@ impl fmt::Debug for SignalInvocationHint {
...
+    pub fn type_(&self) -> Type {
+        self.0
+    }
+
+    /// Queries more in-depth information about the current signal
+    #[doc(alias = "g_signal_query")]
+    pub fn query(&self) -> SignalQuery {
+        unsafe {
+            let query_ptr = std::ptr::null_mut();
+            gobject_ffi::g_signal_query(self.id(), query_ptr);
```
sdroege, line 154: `assert_ne!(query.signal_id, 0);`

glib/src/subclass/signal.rs
```diff
@@ -68,9 +68,103 @@ impl fmt::Debug for SignalInvocationHint {
...
+
+    pub fn type_(&self) -> Type {
+        self.0
+    }
+
+    /// Queries more in-depth information about the current signal
+    #[doc(alias = "g_signal_query")]
+    pub fn query(&self) -> SignalQuery {
+        unsafe {
+            let query_ptr = std::ptr::null_mut();
```
sdroege, line 153:
```suggestion
            let mut query_ptr = mem::MaybeUninit::uninit();
```

And `.as_mut_ptr()` and `.assume_init()` below

glib/src/subclass/signal.rs
bilelmoussaoui: I don't know much about the signals internals, so I will just go ahead and remove the type checking from emit

sdroege: That should be OK, it just causes a critical warning then if you emit the signal on the wrong type

sdroege: Basically, it's less bad than disconnecting a signal handler from the wrong object :)

glib/src/subclass/signal.rs
bilelmoussaoui: Cool, wasn't aware of such type :)

glib/src/subclass/signal.rs
bilelmoussaoui: Alright, i guess that's fine to be handled by the api users :)

glib/src/subclass/signal.rs
bilelmoussaoui: I do wonder if the signal name required, because g_signal_name is output is annotated as nullable, but no idea about this one :)

bilelmoussaoui: never mind, it should never be nullable as the SignalId is now a wrapper around NonZeroU32.

glib/src/subclass/signal.rs
bilelmoussaoui: I'm not so sure how can I convert a GString into a &'a str "cannot return value referencing local variable name returns a value referencing data owned by the current function"

glib/src/subclass/signal.rs
```diff
@@ -68,9 +69,117 @@ impl fmt::Debug for SignalInvocationHint {
...
+            let signal_id = gobject_ffi::g_signal_lookup(name.to_glib_none().0, type_.to_glib());
+            if signal_id == 0 {
+                None
+            } else {
+                Some(Self::new(NonZeroU32::new_unchecked(signal_id)))
+            }
+        }
+    }
+
+    pub fn id(&self) -> NonZeroU32 {
```
bilelmoussaoui, line 157: As the type_ function is removed, I wonder if we can just impl Deref for u32/NonZeroU32? Not sure if it's really useful outside of unsafe ffi code.

glib/src/subclass/signal.rs
```diff
@@ -68,9 +69,117 @@ impl fmt::Debug for SignalInvocationHint {
...
 #[derive(Debug, Clone, Copy)]
-pub struct SignalId(pub(super) Type, pub(super) u32);
+pub struct SignalId(NonZeroU32);
+
+impl SignalId {
+    /// Create a new Signal Identifier.
+    ///
+    /// # Safety
+    ///
+    /// The caller has to ensure it's a valid signal identifier.
```
bilelmoussaoui, line 139: I guess that's the reason it's unsafe? :-)

glib/src/subclass/signal.rs
```diff
@@ -68,9 +69,117 @@ impl fmt::Debug for SignalInvocationHint {
...
+        unsafe { from_glib(self.0.return_type) }
+    }
+
+    /// The number of parameters the user callback takes.
+    pub fn n_params(&self) -> u32 {
+        self.0.n_params
+    }
+
+    /// The parameters for the user callback.
+    pub fn param_types(&self) -> Vec<Type> {
```
bilelmoussaoui, line 111: The SignalBuilder code uses arg_types & ret_types, I wonder if we should use the same names here? :)

glib/src/subclass/signal.rs
sdroege: Yes this only ever returns `NULL` if the signal was invalid, and we catch that earlier :)

glib/src/subclass/signal.rs
sdroege: You wouldn't use `GString` but `std::ffi::CStr`:

```rust
std::ffi::CStr::from_ptr(ptr).to_str().unwrap()
```

glib/src/subclass/signal.rs
sdroege: I wouldn't, but I would implement `FromGlib` and `ToGlib` instead

glib/src/subclass/signal.rs
sdroege: Yes

glib/src/subclass/signal.rs
sdroege: Yes let's make it consistent. Unfortunately it's also not consistent in the C API though.

I would go with param/parameter instead of arg/argument, apart from that select something you like :)

sdroege: Might also make sense to use a `SmallVec<[Type; 16]>` or something like that here. We depend on smallvec anyway and this will basically never be more than 10 or so :)

glib/src/subclass/signal.rs
```diff
@@ -86,7 +195,7 @@ enum SignalRegistration {
     },
     Registered {
         type_: Type,
-        signal_id: u32,
+        signal_id: NonZeroU32,
```
sdroege, line 198: Why not use `SignalId` here?

glib/src/subclass/object.rs
```diff
@@ -241,7 +241,6 @@ impl<T: ObjectImpl> ObjectImplExt for T {
             let instance = self.get_instance();
 
             let signal_id = signal.signal_id();
-            assert!(type_.is_a(&signal_id.0));
```
sdroege, line 244: You can move this whole function to `ObjectExt` now and make it work directly on `SignalId`

glib/src/subclass/signal.rs
bilelmoussaoui: Much better indeed.

glib/src/subclass/object.rs
bilelmoussaoui: There's already an emit function ObjectExt that works with signal name instead of a signal id. I wonder how to refactor such api in a nice way. For that I have two questions:
- Do we want a way to emit a signal by name or just always require a signal id?
- Do we want the `emit_with_details` to be moved as well ? I guess yes

glib/src/subclass/signal.rs
bilelmoussaoui: For my curiosity, why not use GString here?

glib/src/subclass/object.rs
sdroege: > Do we want a way to emit a signal by name or just always require a signal id? Do we want to provide two variants that work with different things: signal name/signal id

I would make the one with the signal id the main implementation, and simply have the one with the name query the signal id and call the other one. It internally does that already anyway.

> Do we want the `emit_with_details` to be moved as well ? I guess yes

Yes

glib/src/subclass/signal.rs
sdroege: Because that's another heap allocation, which is unnecessary for statically allocated strings

glib/src/subclass/signal.rs
```diff
@@ -68,9 +68,129 @@ impl fmt::Debug for SignalInvocationHint {
     }
 }
 
+/// In-depth information of a specific signal
+#[repr(transparent)]
```
sdroege, line 72: Doesn't have to be transparent

glib/src/subclass/signal.rs
```diff
@@ -68,9 +68,129 @@ impl fmt::Debug for SignalInvocationHint {
...
+            std::ffi::CStr::from_ptr(ptr).to_str().unwrap()
+        }
+    }
+
+    /// The ID of the signal.
+    pub fn signal_id(&self) -> SignalId {
+        unsafe { SignalId::from_glib(self.0.signal_id) }
+    }
+
+    /// The interface/instance this signal can be emitted for.
```
sdroege, line 89:
```suggestion
    /// The interface/class this signal can be emitted for.
```

Maybe? Or "instance type"?

glib/src/object.rs
```diff
@@ -1264,15 +1264,19 @@ pub trait ObjectExt: ObjectType {
     where
         N: Into<&'a str>,
         F: Fn(&[Value]) -> Option<Value>;
-    fn emit<'a, N: Into<&'a str>>(
+    /// Emit signal by signal id.
+    fn emit(&self, signal_id: SignalId, args: &[&dyn ToValue]) -> Result<Option<Value>, BoolError>;
```
bilelmoussaoui, line 1268: Gir seems to generate code using emit("signal-name"...). Do it make sense to do the change in gir's side? :)

sdroege: Where would gir get the signal id from?

glib/src/object.rs
```diff
@@ -1264,15 +1264,19 @@ pub trait ObjectExt: ObjectType {
...
+    fn emit_by_name<'a, N: Into<&'a str>>(
         &self,
         signal_name: N,
         args: &[&dyn ToValue],
     ) -> Result<Option<Value>, BoolError>;
-    fn emit_generic<'a, N: Into<&'a str>>(
```
sdroege, line 1272: Why do you remove this one? And we should probably instead of removing this, also add generic variants for the two new functions :)

Also if you have suggestions for the name, something else than generic would be nice. `_with_values()`?

sdroege: Note that there is also other "generic" API, for setting properties for example. That should be changed consistently.

glib/src/object.rs
```diff
@@ -1760,8 +1743,9 @@ impl<T: ObjectType> ObjectExt for T {
         }
 
         // This is actually G_SIGNAL_TYPE_STATIC_SCOPE
-        let return_type: Type =
-            from_glib(details.return_type & (!gobject_ffi::G_TYPE_FLAG_RESERVED_ID_BIT));
+        let return_type: Type = from_glib(
+            signal_query.return_type().to_glib() & (!gobject_ffi::G_TYPE_FLAG_RESERVED_ID_BIT),
```
sdroege, line 1747: This is actually a problem and should probably already be done as part of the `GSignalQuery` bindings. The "return type" should have that flag removed already or otherwise it's not actually a `GType`, but instead we should have another bool for `fn static_scope(&self) -> bool` that only matches that flag.

Aaaalso: we need to be able to set the flag in the signal builder :)

glib/src/object.rs
```diff
@@ -1875,18 +1855,20 @@ impl<T: ObjectType> ObjectExt for T {
...
-                validate_signal_arguments(type_, signal_name, &mut args[1..])?;
+            let signal_detail = validate_signal_arguments(type_, &signal_query, &mut args[1..])?;
 
             let mut return_value = Value::uninitialized();
-            if return_type != Type::Unit {
-                gobject_ffi::g_value_init(return_value.to_glib_none_mut().0, return_type.to_glib());
+            if signal_query.return_type() != Type::Unit {
+                gobject_ffi::g_value_init(
+                    return_value.to_glib_none_mut().0,
+                    signal_query.return_type().to_glib(),
```
sdroege, line 1864: E.g. here the flag should've also been removed

glib/src/object.rs
sdroege: Also the flag can be in all the parameter types of `GSignalQuery` and needs to be removed and handled the same way there...

glib/src/subclass/object.rs
```diff
@@ -848,7 +661,7 @@ mod test {
         .expect("Failed to connect on 'create-child-object'");
 
         let value = obj
-            .emit("create-child-object", &[])
+            .emit_by_name("create-child-object", &[])
```
sdroege, line 664: We could make use of `emit()` here in one place or another to increase test coverage a bit (but also keep calling `emit_by_name()` somewhere for the same reason)

glib/src/object.rs
sdroege: Or maybe for now we can simply unset that flag in `Type::from_glib()`, and handle the static scope stuff in a separate PR later.

glib/src/object.rs
bilelmoussaoui: change it as in replace the call with emit_by_name

glib/src/object.rs
bilelmoussaoui: > Why do you remove this one? And we should probably instead of removing this, also add generic variants for the two new functions :)

Ah, just noticed the difference between both emit & emit_generic. Maybe `emit_with_values`?

glib/src/object.rs
sdroege: Ah, yes that makes sense :)

glib/src/object.rs
sdroege: And `emit_by_name_with_values()` and `emit_with_details_and_values()` and `set_property_with_values()`, etc?

Apart from the details one this seems like an improvement.

glib/src/object.rs
bilelmoussaoui: Aha, interesting. I ended up always removing the extra bit in Type::from_glib but I don't know how to check if it's set in a GType for `static_scope`. For the signal builder part, the extra bit has to be set only in the return_type I assume? or in all the various types?

sdroege: > but I don't know how to check if it's set in a GType for static_scope

What do you mean with that? You can check if it's set by doing `(type_ & gobject_ffi::G_TYPE_FLAG_RESERVED_ID_BIT) != 0` for example.

> For the signal builder part, the extra bit has to be set only in the return_type I assume? or in all the various types?

Return and parameter types.

bilelmoussaoui: I have pushed something that should handle the static scope stuff :) not very confident about some parts of it ^^

glib/src/object.rs
```diff
@@ -1726,42 +1729,17 @@ impl<T: ObjectType> ObjectExt for T {
...
-        let mut details = mem::MaybeUninit::zeroed();
-        gobject_ffi::g_signal_query(signal_id, details.as_mut_ptr());
-        let details = details.assume_init();
-        if details.signal_id != signal_id {
-            return Err(bool_error!(
-                "Signal '{}' of type '{}' not found",
-                signal_name,
-                type_
-            ));
+        if !signal_query.static_scope() {
```
bilelmoussaoui, line 1738: Does it make sense to have such validation here? I'm not very sure.

sdroege: No, that shouldn't be checked there

glib/src/subclass/object.rs
```diff
@@ -806,8 +619,11 @@ mod test {
         })
         .expect("Failed to connect on 'create-string'");
 
+        let signal_id = SignalId::lookup("create-string", SimpleObject::static_type())
+            .expect("Signal crate-string not found");
```
sdroege, line 623: You can also get it directly from `SimpleObject::signals()[0]` :)

glib/src/subclass/signal.rs
```diff
@@ -6,17 +6,18 @@ use crate::SignalFlags;
...
 /// Builder for signals.
 #[allow(clippy::type_complexity)]
 pub struct SignalBuilder<'a> {
     name: &'a str,
     flags: SignalFlags,
-    arg_types: &'a [Type],
-    ret_type: Type,
+    param_types: &'a [Type],
+    return_type: Type,
+    static_scope: bool,
```
sdroege, line 20: Each parameter and the return type can be static scope individually

glib/src/object.rs
sdroege: Looks good except for the detail that it's a flag per return type / parameter, not one flag for all of them per signal. They can be marked individually as static scope.

sdroege: We can move that to a separate PR if you prefer

glib/src/subclass/signal.rs
bilelmoussaoui: Right, which means that the static_scope should be per param_type & return_type as well I guess?

glib/src/object.rs
bilelmoussaoui: > We can move that to a separate PR if you prefer

It should be fine :-) unless we can't figure out a nice way to set the static_scope per param/return type, for now I went with (Type, bool) which isn't super clear...

glib/src/subclass/signal.rs
bilelmoussaoui: I meant for the SignalQuery part

glib/src/subclass/signal.rs
```diff
@@ -361,8 +385,8 @@ impl Signal {
     /// Create a new builder for a signal.
     pub fn builder<'a>(
         name: &'a str,
-        param_types: &'a [Type],
-        return_type: Type,
+        param_types: &'a [(Type, bool)],
+        return_type: (Type, bool),
```
bilelmoussaoui, line 389: I'm not very happy about this, as the user can't tell it those booleans are for setting the param/return type as static scope. Any ideas?

glib/src/subclass/signal.rs
sdroege: Yes

glib/src/subclass/signal.rs
sdroege: Maybe there could be some kind of new type `SignalType(ffi::GType)` that is `From<Type> for SignalType` (and the inverse of course) and also has a `fn with_static_scope(t: Type) -> Self`, and we'd work with those here instead?

bilelmoussaoui: Ah, great. That can work and simplify the SignalQuery part as well. I like it.

glib/src/subclass/object.rs
bilelmoussaoui: I'm not sure about this one, ObjectImpl is implemented for the private struct, you can't call signals on the public one. see
```

error[E0599]: no function or associated item named `signals` found for struct `subclass::object::test::SimpleObject` in the current scope
   --> glib/src/subclass/object.rs:622:39
    |
622 |         let signal_id = SimpleObject::signals()[0];
    |                                       ^^^^^^^ function or associated item not found in `subclass::object::test::SimpleObject`
    | 
   ::: glib/src/object.rs:640:9
    |
640 |         pub struct $name($crate::object::ObjectRef);
    |         -------------------------------------------- function or associated item `signals` not found for this
    ```

File: glib/src/subclass/signal.rs [hunk 25]
```diff
@@ -225,6 +221,92 @@ impl PartialEq for SignalId {
...
+
+    pub fn static_scope(&self) -> bool {
+        (self.0 & gobject_ffi::G_TYPE_FLAG_RESERVED_ID_BIT) != 0
+    }
+}
+
+impl From<Type> for SignalType {
+    fn from(type_: Type) -> Self {
+        Self(type_.to_glib())
+    }
```
bilelmoussaoui, line 240: I'm not super sure about this, as the extra-bit for static scope is removed at `Type::from_glib` which means the `SignalType` might contain less information here. I wonder if it would make more sense to remove it only when doing `From<SignalType> for Type`

glib/src/subclass/object.rs
sdroege: Oh yeah, you need to get the other struct :)

glib/src/subclass/signal.rs
sdroege: Yes that could make sense, the you only need to be very careful to consistently use `SignalType` for all the signal related types.

glib/src/subclass/signal.rs
```diff
@@ -12,11 +12,11 @@ use std::{fmt, num::NonZeroU32};
 
 /// Builder for signals.
 #[allow(clippy::type_complexity)]
-pub struct SignalBuilder<'a> {
+pub struct SignalBuilder<'a, T: Into<SignalType> + Clone> {
```
sdroege, line 15: That makes not much sense. You will usually use `Type` for most things and a `SignalType` for one parameter or two, but you can't mix those two here and will get confusing errors.

I'd make this un-generic and use `SignalType` everywhere

glib/src/subclass/signal.rs
```diff
@@ -225,6 +221,92 @@ impl PartialEq for SignalId {
     }
 }
 
+#[derive(Copy, Clone, Hash)]
```
sdroege, line 224: And `PartialEq, Eq` etc maybe?

glib/src/subclass/signal.rs
```diff
@@ -225,6 +221,92 @@ impl PartialEq for SignalId {
...
+
+impl From<SignalType> for Type {
+    fn from(type_: SignalType) -> Self {
+        unsafe { from_glib(type_.0) }
+    }
+}
+
+impl PartialEq<Type> for SignalType {
+    fn eq(&self, other: &Type) -> bool {
+        let type_: Type = self.clone().into();
```
sdroege, line 251: It's `Copy`, you don't need to `clone()` :)

sdroege: Also below

glib/src/subclass/signal.rs
```diff
@@ -225,6 +221,92 @@ impl PartialEq for SignalId {
...
+        type_.eq(other)
+    }
+}
+
+impl std::fmt::Debug for SignalType {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        let type_: Type = self.clone().into();
+        f.write_str(&type_.name())
```
sdroege, line 259: Might want to add something for static scope here

glib/src/subclass/signal.rs
bilelmoussaoui: Yes, I wanted to avoid having to do .into() everywhere where I had `T::static_type()` before. I have made it un-generic now

glib/src/subclass/object.rs
bilelmoussaoui: Sorry, I feel stupid.

bilelmoussaoui:
I think this one is ready for another round of review :-)

glib/src/subclass/signal.rs
```diff
@@ -68,9 +70,254 @@ impl fmt::Debug for SignalInvocationHint {
...
+#[doc(hidden)]
+impl ToGlib for SignalId {
+    type GlibType = u32;
+
+    fn to_glib(&self) -> u32 {
+        self.0.into()
+    }
+}
+
+impl PartialEq for SignalId {
```
sdroege, line 222: Can't we derive that?

glib/src/subclass/signal.rs
```diff
@@ -68,9 +70,254 @@ impl fmt::Debug for SignalInvocationHint {
...
+
+#[derive(Copy, Clone, Hash)]
+pub struct SignalType(ffi::GType);
+
+impl SignalType {
+    pub fn with_static_scope(type_: Type) -> Self {
+        Self(type_.to_glib() | gobject_ffi::G_TYPE_FLAG_RESERVED_ID_BIT)
+    }
+
+    pub fn static_scope(&self) -> bool {
```
sdroege, line 236: Maybe for completeness also add a `fn type(&self) -> Type`?

Also why doesn't clippy complain that `self` should be taken by value here? Should be the case here and also for `Type` everywhere as they're `Copy` and small.

glib/src/subclass/signal.rs
```diff
@@ -247,7 +498,10 @@ impl Signal {
     pub fn signal_id(&self) -> SignalId {
         match &*self.registration.lock().unwrap() {
             SignalRegistration::Unregistered { .. } => panic!("Signal not registered yet"),
-            SignalRegistration::Registered { type_, signal_id } => SignalId(*type_, *signal_id),
+            SignalRegistration::Registered {
+                type_: _,
+                signal_id,
+            } => *signal_id,
```
sdroege, line 504: Why not `SignalRegistration::Registered { signal_id, .. } => *signal_id`?

sdroege (approved):
Looks good to me otherwise, thanks :)

glib/src/subclass/signal.rs
bilelmoussaoui: > Also why doesn't clippy complain that self should be taken by value here? Should be the case here and also for Type everywhere as they're Copy and small.

Maybe such linter doesn't exist because it doesn't complain for SignalId neither for example. Should I handle it on the same PR or it can be done in an another one?

sdroege: Well, it exists and I saw it triggered for such cases in the past :) That would be https://rust-lang.github.io/rust-clippy/master/index.html#trivially_copy_pass_by_ref

Can be done in a separate PR but then please create an issue so we don't forget.

glib/src/subclass/signal.rs
```diff
@@ -68,9 +70,252 @@ impl fmt::Debug for SignalInvocationHint {
...
+            .field("flags", &self.flags())
+            .field("return_type", &self.return_type())
+            .field("param_types", &self.param_types())
+            .finish()
+    }
+}
 /// Signal ID.
-#[derive(Debug, Clone, Copy)]
-pub struct SignalId(pub(super) Type, pub(super) u32);
+#[derive(Debug, Clone, Copy, PartialEq)]
```
sdroege, line 133:
```suggestion
#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
```

While we're at it

sdroege (approved):

glib/src/subclass/signal.rs
bilelmoussaoui: see https://github.com/gtk-rs/gtk-rs/issues/313

sdroege:
Waiting for @GuillaumeGomez now. Before merging this we need to merge the `gir` PR, then update this here to regenerate with that `gir` version and then it can be merged.

Also will need to spend some time updating the GStreamer bindings and plugins.

GuillaumeGomez (approved):

sdroege:
@bilelmoussaoui Then update this PR to latest gir and let's get it merged :)

glib/src/object.rs
```diff
@@ -1190,13 +1190,16 @@ pub trait ObjectExt: ObjectType {
         property_name: N,
         value: &V,
     ) -> Result<(), BoolError>;
-    fn set_property_generic<'a, N: Into<&'a str>>(
+    fn set_property_with_value<'a, N: Into<&'a str>>(
```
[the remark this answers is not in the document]
sdroege: Also we missed `Object::new_generic()` -> `with_values()`? Or ...?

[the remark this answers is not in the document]
sdroege: And `Closure::invoke_generic()` -> `invoke_with_values()`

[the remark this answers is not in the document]
bilelmoussaoui: > This is acutally not very nice. Nicer would be `set_property_from_value()` and `set_properties_from_value()` (or `from_values()` but that's double plural then...).

Ah. I missed a bunch it seems, I will open an other PR this evening :)

PR #304, opened 2021-02-16
glib: Correctly mark future returned by ThreadPool::push_future() as …

sdroege:
…Send+Sync+'static

----

It could've also been `Unpin` but because of the `async` block that's not possible anymore. Probably not much of a problem in practice. (CC @jplatte )

### DIFF
--- a/glib/src/thread_pool.rs
+++ b/glib/src/thread_pool.rs
@@ -76,7 +76,7 @@ impl ThreadPool {
     pub fn push_future<T: Send + 'static, F: FnOnce() -> T + Send + 'static>(
         &self,
         func: F,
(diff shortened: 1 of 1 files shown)


GuillaumeGomez (approved):

jplatte:
That's unfortunate. I wonder whether there are any plans to make futures that don't borrow across `.await`s `Unpin` (a cursory search returned nothing).

sdroege:
> That's unfortunate. I wonder whether there are any plans to make futures that don't borrow across `.await`s `Unpin` (a cursory search returned nothing).

Can you create an issue and link it here? I would've assumed that to be part of the plans :)

jplatte:
https://github.com/rust-lang/rust/issues/82187

Issue #305, opened 2021-02-16
Crash while calling gio::File::load_contents() on a gresource


bellegarde-c:
Calling gio::File::load_contents() on a resource:// URI segfaults.

Reproducer: https://github.com/bellegarde-c/gtk4-rs-gio-file-loads-content-crash

Crash is in: https://github.com/bellegarde-c/gtk4-rs-gio-file-loads-content-crash/blob/master/src/application.rs -> fn setup_css() after println!()

backtrace: 
[gdb.txt](https://github.com/gtk-rs/gtk4-rs/files/5990615/gdb.txt)

sdroege:
The `etag` out parameter of `g_file_load_contents()` needs to be nullable apparently, but isn't. Unclear if that's a bug in gio (maybe it must never be `NULL` and that's the problem here), or just an annotation problem (it's valid to be `NULL`).

Will check in more detail tomorrow.

sdroege:
See https://gitlab.gnome.org/GNOME/glib/-/merge_requests/1956 . The annotations were simply missing.

bellegarde-c:
Thanks, will use load_bytes() until the fix is in org.gnome.Sdk//master.

PR #307, opened 2021-02-17
gio: Mark ETag out parameter in various GFile functions as nullable

sdroege:
See https://gitlab.gnome.org/GNOME/glib/-/merge_requests/1956

Fixes https://github.com/gtk-rs/gtk-rs/issues/305


sdroege:
@GuillaumeGomez Any ideas why gir doesn't actually make use of this? :)

GuillaumeGomez:
Very likely because it wasn't implemented... Do you need the `name` field for a return parameter?

sdroege:
> Very likely because it wasn't implemented... Do you need the `name` field for a return parameter?

This is an out parameter. Ok, I'll investigate...

sdroege:
I think because the parameter analysis happens *after* creating the bounds...

sdroege:
`src/analysis/functions.rs:analyze_callbacks()` is the problem. It first calls `bounds.add_for_parameter()` and only at the bottom calls `function_parameters::analyze()`.

sdroege:
Or maybe not. The same code is called in many different places in slightly different ways...

bilelmoussaoui (approved):

GuillaumeGomez (approved):

PR #310, opened 2021-02-20
gio: File::get_child() is not nullable

sophie-h:


gio/Gir.toml
```diff
@@ -620,6 +620,10 @@ manual_traits = ["FileExtManual"]
         [[object.function.parameter]]
         name = "new_etag"
         nullable = true
+    [[object.function]]
+    name = "get_child"
+        [object.function.return]
+        nullable = false
```
sdroege, line 626: It's going to be so great once we can update to 2.68 and just trust the upstream annotations...

Issue #285, opened 2021-02-01
GTK TreeView: panic when using more than two columns


RomeuG:
Hi everyone.

I hope I am not missing anything but, when trying out your simple TreeView example, if I add one more column, the program just crashes.

First off, I am using these dependencies:
```
[dependencies.gtk]
version = "0.9.0"
features = ["v3_16"]

[dependencies.gio]
version = ""
features = ["v2_44"]
```

Here is a gist with the changed code: https://gist.github.com/RomeuG/b1c20a4e7bbd3c715fa1e962695bb661

The crash is this:

```
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `2`,
 right: `3`', /home/romeu/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/libstd/macros.rs:16:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

The backtrace offers more information on the line of code this is happening:

```
13: <O as gtk::list_store::GtkListStoreExtManual>::insert_with_values
             at /home/romeu/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/libstd/macros.rs:16
14: liststore::create_and_fill_model
             at src/main.rs:21
```

Correct me if I am wrong, but can't ListStore have more than 2 columns? Is it something I am forgetting?
Thanks!

sdroege:
```rust
        model.insert_with_values(None, &[0, 2], &[&(i as u32 + 1), &entry, &more_entries[i]]);
```

You're here inserting 3 values (the last parameter) into 2 columns (the second parameter, `&[0, 2]`). That doesn't work, you always have to provide the same number of elements for both.

Maybe that function should be changed to take a slice of tuples instead so this mistake is prevented by the API. @GuillaumeGomez @bilelmoussaoui, opinions?

RomeuG:
Oh, wow. I get it now, I should use something like `&[0, 1, 2]`.
I will be honest I was reading `&[0, 2]` as a range of values from 0 to 2 inclusive.

I don't really know what is best in this case, so I will let you discuss it.

Thanks!

bilelmoussaoui:
> Maybe that function should be changed to take a slice of tuples instead so this mistake is prevented by the API. @GuillaumeGomez @bilelmoussaoui, opinions?

So that you would write the code above like
 ```rust
[...]

PR #312, opened 2021-02-20
gtk: Take (column, value) tuples instead of separate slices for the t…

sdroege:
…ree/list store setter/insertion function

This automatically ensures that there are as many columns as there are
values and makes the API closer to the C API, which takes varargs that
are column, value, column, value, etc.

Fixes https://github.com/gtk-rs/gtk-rs/issues/285

### DIFF
--- a/gtk/src/list_store.rs
+++ b/gtk/src/list_store.rs
@@ -28,15 +28,14 @@ pub trait GtkListStoreExtManual: 'static {
     fn insert_with_values(
         &self,
         position: Option<u32>,
-        columns: &[u32],
-        values: &[&dyn ToValue],
+        columns_and_values: &[(u32, &dyn ToValue)],
     ) -> TreeIter;
 
     #[doc(alias = "gtk_list_store_reorder")]
     fn reorder(&self, new_order: &[u32]);
 
     #[doc(alias = "gtk_list_store_set")]
-    fn set(&self, iter: &TreeIter, columns: &[u32], values: &[&dyn ToValue]);
+    fn set(&self, iter: &TreeIter, columns_and_values: &[(u32, &dyn ToValue)]);
 
     #[doc(alias = "gtk_list_store_set_value")]
     fn set_value(&self, iter: &TreeIter, column: u32, value: &Value);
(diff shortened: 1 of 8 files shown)


gtk/src/list_store.rs
```diff
@@ -46,31 +45,39 @@ impl<O: IsA<ListStore>> GtkListStoreExtManual for O {
...
         unsafe {
             assert!(position.unwrap_or(0) <= i32::max_value() as u32);
-            assert_eq!(columns.len(), values.len());
             let n_columns = ffi::gtk_tree_model_get_n_columns(
                 self.as_ref().upcast_ref::<TreeModel>().to_glib_none().0,
             ) as u32;
-            assert!(columns.len() <= n_columns as usize);
-            for (&column, value) in columns.iter().zip(values.iter()) {
-                assert!(column < n_columns);
+            assert!(columns_and_values.len() <= n_columns as usize);
```
GuillaumeGomez, line 55: Please add a error message as second argument.

gtk/src/list_store.rs
```diff
@@ -46,31 +45,39 @@ impl<O: IsA<ListStore>> GtkListStoreExtManual for O {
...
+            for (column, value) in columns_and_values {
+                assert!(*column < n_columns);
```
GuillaumeGomez, line 57: Same.

gtk/src/list_store.rs
```diff
@@ -110,26 +117,35 @@ impl<O: IsA<ListStore>> GtkListStoreExtManual for O {
...
+    fn set(&self, iter: &TreeIter, columns_and_values: &[(u32, &dyn ToValue)]) {
         unsafe {
-            assert_eq!(columns.len(), values.len());
             let n_columns = ffi::gtk_tree_model_get_n_columns(
                 self.as_ref().upcast_ref::<TreeModel>().to_glib_none().0,
             ) as u32;
-            assert!(columns.len() <= n_columns as usize);
-            for (&column, value) in columns.iter().zip(values.iter()) {
-                assert!(column < n_columns);
+            assert!(columns_and_values.len() <= n_columns as usize);
```
GuillaumeGomez, line 125: Same.

gtk/src/list_store.rs
```diff
@@ -110,26 +117,35 @@ impl<O: IsA<ListStore>> GtkListStoreExtManual for O {
...
+            for (column, value) in columns_and_values {
+                assert!(*column < n_columns);
```
GuillaumeGomez, line 127: Same.

gtk/src/list_store.rs
```diff
@@ -110,26 +117,35 @@ impl<O: IsA<ListStore>> GtkListStoreExtManual for O {
         }
     }
 
-    fn set(&self, iter: &TreeIter, columns: &[u32], values: &[&dyn ToValue]) {
+    fn set(&self, iter: &TreeIter, columns_and_values: &[(u32, &dyn ToValue)]) {
```
GuillaumeGomez, line 120: We should add a doc alias for `gtk_list_store_set_valuesv` here I think.

gtk/src/tree_store.rs
```diff
@@ -47,31 +46,40 @@ impl<O: IsA<TreeStore>> TreeStoreExtManual for O {
...
     ) -> TreeIter {
         unsafe {
             assert!(position.unwrap_or(0) <= i32::max_value() as u32);
-            assert_eq!(columns.len(), values.len());
             let n_columns = ffi::gtk_tree_model_get_n_columns(
                 self.as_ref().upcast_ref::<TreeModel>().to_glib_none().0,
             ) as u32;
-            assert!(columns.len() <= n_columns as usize);
-            for (&column, value) in columns.iter().zip(values.iter()) {
+            assert!(columns_and_values.len() <= n_columns as usize);
```
GuillaumeGomez, line 56: Same.

gtk/src/tree_store.rs
```diff
@@ -47,31 +46,40 @@ impl<O: IsA<TreeStore>> TreeStoreExtManual for O {
...
+            for (column, value) in columns_and_values {
+                assert!(*column < n_columns);
```
GuillaumeGomez, line 58: Same.

gtk/src/tree_store.rs
```diff
@@ -112,26 +120,35 @@ impl<O: IsA<TreeStore>> TreeStoreExtManual for O {
...
+    fn set(&self, iter: &TreeIter, columns_and_values: &[(u32, &dyn ToValue)]) {
         unsafe {
-            assert_eq!(columns.len(), values.len());
             let n_columns = ffi::gtk_tree_model_get_n_columns(
                 self.as_ref().upcast_ref::<TreeModel>().to_glib_none().0,
             ) as u32;
-            assert!(columns.len() <= n_columns as usize);
-            for (&column, value) in columns.iter().zip(values.iter()) {
-                assert!(column < n_columns);
+            assert!(columns_and_values.len() <= n_columns as usize);
```
GuillaumeGomez, line 128: Same.

gtk/src/tree_store.rs
```diff
@@ -112,26 +120,35 @@ impl<O: IsA<TreeStore>> TreeStoreExtManual for O {
...
+            for (column, value) in columns_and_values {
+                assert!(*column < n_columns);
```
GuillaumeGomez, line 130: Same.

gtk/src/list_store.rs
sdroege: You mean the assertion in words? Or what error message?

GuillaumeGomez: Just the second argument (which will be the error message). Something like "too many columns are present, expected {}, found {}".

sdroege: Ah, that makes sense yes

GuillaumeGomez (approved):
I think we could improve the error message but it can be done in another PR.

GuillaumeGomez:
Thanks!

sdroege:
> I think we could improve the error message but it can be done in another PR.

If you have suggestions, create a PR :) I spent a bit of time coming up with more descriptive messages than what the assertion would create by default and this is the best I could come up with.

PR #315, opened 2021-02-22
glib: rename remaining _generic to _values

bilelmoussaoui:
See https://github.com/gtk-rs/gtk-rs/pull/302#discussion_r580474627

### DIFF
--- a/glib/src/object.rs
+++ b/glib/src/object.rs
@@ -1105,7 +1105,7 @@ impl Object {
         unsafe { Object::new_internal(type_, &params) }
     }
 
-    pub fn new_generic(type_: Type, properties: &[(&str, Value)]) -> Result<Object, BoolError> {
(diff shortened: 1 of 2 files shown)


glib/src/object.rs
```diff
@@ -1105,7 +1105,7 @@ impl Object {
         unsafe { Object::new_internal(type_, &params) }
     }
 
-    pub fn new_generic(type_: Type, properties: &[(&str, Value)]) -> Result<Object, BoolError> {
+    pub fn with_values(type_: Type, properties: &[(&str, Value)]) -> Result<Object, BoolError> {
```
bilelmoussaoui, line 1108: Maybe `with_type_and_values`? Like `with_details_and_values`. I don't like neither though.

sdroege: You need to provide the type in any case, so mentioning it in the function name seems not very useful. Like this should be fine.

PR #316, opened 2021-02-23
WIP: glib: add a static attribute macro

bilelmoussaoui:


glib-macros/src/static_attribute.rs
```diff
@@ -0,0 +1,43 @@
...
+
+pub fn impl_static(input: ItemFn) -> TokenStream {
+    let vis = &input.vis;
+    let block = &input.block;
+    let name = &input.sig.ident;
+    let output = &input.sig.output;
+    
+    // We need to get the inner type T out of &'static [T]
+    let output_type = match output {
+        syn::ReturnType::Type(_, ret_type) => match &**ret_type {
```
bilelmoussaoui, line 16: My question here, is whether we would write something like `fn signals() -> &'static [Signal]`, if that's the case, we would need a way to get `T` from the return type so that I can use it in `Vec<T>` below.

bilelmoussaoui: In other words, would the the macro be used only for those two use cases for now: properties/signals?

glib-macros/src/lib.rs
```diff
@@ -374,3 +375,32 @@ pub fn downgrade(input: TokenStream) -> TokenStream {
...
+/// pub fn static_func() -> &'static [String] {
+///     static items: Lazy<Vec<String>> = Lazy::new(|| {
+///         vec!["item".to_string(), "item2".to_string()]
+///     });
+///     items.as_ref()
+/// }
+/// ```
+/// With the macro
+/// ```rust
+/// #[static_]
```
sdroege, line 396: Maybe `lazy_static` would be a more descriptive name

sdroege: (note that this is basically what the `lazy_static!` macro did, just in attribute position now...)

glib-macros/src/static_attribute.rs
sdroege: I think you could implement it a bit more generically. The user would write `fn signals() -> &'static [Signal]`, you extract the type: `[Signal]`. It's a slice, so you use a `Vec<_>` internally.

Similarly if you the is some plain `T`, you would work with that directly. If it's a `str` you would work with `String`, if it's a `Path` you would work with `PathBuf`. And that should cover most interesting cases for now.

glib-macros/src/lib.rs
bilelmoussaoui: The name was just there to get something working first. Renamed it to `lazy_static`

glib-macros/src/static_attribute.rs
```diff
@@ -0,0 +1,43 @@
...
+                }
+            }
+            _ => abort_call_site!("#[lazy_static] expects a [T] return type"),
+        },
+        _ => abort_call_site!("#[lazy_static] expects a &'static lifetime"),
+    };
+
+    (quote! {
+        #vis fn #name() #output {
+            static STATIC: once_cell::sync::Lazy<Vec<#output_type>> = once_cell::sync::Lazy::new(|| {
```
ids1024, line 37: Presumably this should use a re-export so the crate using the macro doesn't have to have `once_cell` in `Cargo.toml`.

glib-macros/src/static_attribute.rs
bilelmoussaoui: > I think you could implement it a bit more generically. The user would write fn signals() -> &'static [Signal], you extract the type: [Signal]. It's a slice, so you use a Vec<_> internally.

Right, that's what it does now

> Similarly if you the is some plain T, you would work with that directly. If it's a str you would work with String, if it's a Path you would work with PathBuf. And that should cover most interesting cases for now.

After a bit of research,  I couldn't find any nice way to special case `Path`. I pushed something but it's very hackish. I was suggested to use traits for such situation  but I'm not sure whether it's worth it here.

sdroege: Well there's [`ToOwned`](https://doc.rust-lang.org/std/borrow/trait.ToOwned.html), maybe that can actually be used here. It covers the `Vec`, `String` and `PathBuf` cases at least, and has a blanket implementation for everything that is `Clone`.

GuillaumeGomez:

PR #317, opened 2021-02-24
glib: Derive Ord on Type

heftig:
Useful when Types need to be used in sets or as map keys. Hash is
already derived, just PartialOrd and Ord were missing.


sdroege (approved):

bilelmoussaoui (approved):

GuillaumeGomez:
It would have been better to have put the second commit in another PR. It has literally nothing to do with the first one. :-/

Issue #313, opened 2021-02-22
figure out why trivially_copy_pass_by_ref doesn't report some cases


bilelmoussaoui:
see https://github.com/gtk-rs/gtk-rs/issues#discussion_r580166190

I guess it's mostly caused by the linter not being able to figure out that for  `Wrapper(SmallSizeAndCopy)` and has to be manually checked for now

sdroege:
`Type` was updated accordingly, `SignalType` not and should check some others too

PR #318, opened 2021-02-24
glib: Improve the API of Type

heftig:
- Make `Type` leaner, replace enum with assoc consts.
- Pass `Type` by value, where possible
- Introduce `Type::is_valid` and use it
- Make `Type::name` return a `str`

### DIFF
--- a/glib/src/types.rs
+++ b/glib/src/types.rs
@@ -3,8 +3,8 @@
 //! Runtime type information.
 
 use crate::translate::{
-    from_glib, from_glib_none, FromGlib, FromGlibContainerAsVec, ToGlib, ToGlibContainerFromSlice,
-    ToGlibPtr, ToGlibPtrMut,
+    from_glib, FromGlib, FromGlibContainerAsVec, ToGlib, ToGlibContainerFromSlice, ToGlibPtr,
+    ToGlibPtrMut,
 };
 use crate::value::{FromValue, FromValueOptional, SetValue, Value};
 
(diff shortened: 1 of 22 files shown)


sdroege:
This kind of handles https://github.com/gtk-rs/gtk-rs/issues/313 (for `Type`) at least, which is good.

glib/src/types.rs
```diff
@@ -13,70 +13,89 @@ use std::mem;
...
     /// The fundamental type from which all objects are derived
-    Object,
-    /// A non-fundamental type identified by value of type `usize`
-    Other(usize),
-}
+    pub const OBJECT: Self = Self(gobject_ffi::G_TYPE_OBJECT);
 
-impl Type {
     #[doc(alias = "g_type_name")]
     pub fn name(&self) -> String {
```
sdroege, line 87: Can you make this

```rust
     pub fn name<'a>(&'a self) -> &'a str { 
```
in a separate commit here? The name is backed by a quark so it's by definition a static reference

glib/src/types.rs
```diff
@@ -100,7 +100,7 @@ impl Type {
     }
 
     #[doc(alias = "g_type_is_a")]
-    pub fn is_a(&self, other: &Type) -> bool {
+    pub fn is_a(&self, other: Self) -> bool {
```
sdroege, line 103: Should do the same for all the other functions here, I guess. Take `self` instead of `&self`, etc.

sdroege (commented):
Looks good to me otherwise, thanks! That should improve the code generated by the compiler for these things quite a bit and keep it simpler.

glib/src/types.rs
heftig: Leaving `is_valid` with `&self` made it possible to write `.filter(Type::is_valid)`. On the other hand, `map` would prefer `self`.

sdroege: The reasoning here is mostly https://rust-lang.github.io/rust-clippy/master/index.html#trivially_copy_pass_by_ref

heftig: Done.

glib/src/types.rs
heftig: Done.

sdroege (approved):

bilelmoussaoui (approved):

PR #321, opened 2021-02-25
glib-macros: Forward the visibility of the type into the code generat…

sdroege:
…ed by the gflags! macro

----

CC @gdesmott


bilelmoussaoui (approved):

GuillaumeGomez (approved):
Thanks!

PR #324, opened 2021-02-26
API changes due to signal-detail PR to gir

zec:
As part of PR gtk-rs/gir#1062, I was asked to make a PR to show the changes that would have on gtk-rs. This is that PR.


atk/src/auto/object.rs
```diff
@@ -814,19 +821,27 @@ impl<O: IsA<Object>> AtkObjectExt for O {
...
-                b"active-descendant-changed\0".as_ptr() as *const _,
+                signal_name.as_ptr() as *const _,
                 Some(transmute::<_, unsafe extern "C" fn()>(
                     active_descendant_changed_trampoline::<Self, F> as *const (),
                 )),
                 Box_::into_raw(f),
-            )
+            );
+            drop(detailed_signal_name);
+            handler_id
```
MarijnS95, line 838: As mentioned in the gir review you really don't need to drop `detailed_signal_name` here; this is the end of the function where it goes out of scope and is cleaned up by definition :)

atk/src/auto/object.rs
```diff
@@ -814,19 +821,27 @@ impl<O: IsA<Object>> AtkObjectExt for O {
         }
         unsafe {
             let f: Box_<F> = Box_::new(f);
-            connect_raw(
+            let detailed_signal_name =
+                detail.map(|name| format!("active-descendant-changed::{}\0", name));
+            let signal_name: &[u8] = detailed_signal_name
```
MarijnS95, line 826: I don't recall the explicit type being needed here, both `&b"..."[..]` and `.as_bytes()` return `&[u8]`.

zec: It's more documentation for anyone looking at the generated code than anything else.

GuillaumeGomez (approved):

atk/src/auto/object.rs
sdroege: It's especially useful because we need a blanket cast below because of `char*` in C having different signedness depending on the platform.

sdroege (approved):

atk/src/auto/object.rs
MarijnS95: Yup, bytes are always `u8` yet `c_char` has a different definition based on the platform. Only way to get that consistent is to use a `CStr` iirc, which is another unnecessary can of worms here.

sdroege (changes requested):
Please add a commit that updates the `gir` submodule and then regenerate everything on top of that. Thanks!

zec:
Will do.

PR #325, opened 2021-02-27
CI: build stable/unstable docs via gh-actions and publish to docs repo

jsparber:
The stable docs are generated for published releases and pushed to
`gtk-rs/gtk-rs.github.io` in the folder `docs/stable`.
The unstable docs are pushed to the same repo but to the folder
`docs/unstable`.

Note: this doesn't change the current docs, but could break `https://gtk-rs.org/docs` since the `docs` repo and the `gtk-rs/gtk-rs.github.io/docs` share the same url. Apparently the repo gets preference: https://stackoverflow.com/questions/25396383/does-a-folder-in-my-user-site-conflict-with-a-project-site

This requires the setup of a `deploy key` for the repo `gtk-rs/gtk-rs.github.io` which need to be exposed as `secrets.ACTIONS_DEPLOY_KEY`. It's pretty easy see here: https://github.com/peaceiris/actions-gh-pages#%EF%B8%8F-create-ssh-deploy-key

This is only the first step in solving the confusion about the docs locations and is the first step in resolving the issue https://github.com/gtk-rs/gtk-rs.github.io/issues/105.
This also allows us turn off the gh-pages for `gtk-rs` repo.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -2,6 +2,8 @@ on:
   push:
     branches: [master]
   pull_request:
+  release:
+    types: [published]
 
 name: docs
 
@@ -50,9 +52,19 @@ jobs:
         env: 
           RUSTDOCFLAGS: "--enable-index-page -Zunstable-options"
 
-      - name: deploy
+      - name: Deploy stable docs
+        if: ${{ github.event_name == 'release' }}
+        uses: peaceiris/actions-gh-pages@v3
+        with:
+          publish_dir: target/doc/
+          deploy_key: ${{ secrets.ACTIONS_DEPLOY_KEY }}
+          external_repository: gtk-rs/docs
+          destination_dir: stable/
+      - name: Deploy unstable docs
         if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/master' }}
         uses: peaceiris/actions-gh-pages@v3
         with:
-          github_token: ${{ secrets.GITHUB_TOKEN }}
-          publish_dir: ./target/doc/
+          publish_dir: target/doc/
+          deploy_key: ${{ secrets.ACTIONS_DEPLOY_KEY }}
+          external_repository: gtk-rs/docs
+          destination_dir: unstable/


.github/workflows/docs.yml
```diff
@@ -2,6 +2,8 @@ on:
   push:
     branches: [master]
   pull_request:
+  release:
```
bilelmoussaoui, line 5: We don't really tag new releases so far, we just push stuff to the `crate` branch iirc. Not sure if that will change in the next release

.github/workflows/docs.yml
```diff
@@ -50,9 +52,19 @@ jobs:
...
+          external_repository: gtk-rs/gtk-rs.github.io
+          destination_dir: docs/stable/
+      - name: Deploy unstable docs
         if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/master' }}
         uses: peaceiris/actions-gh-pages@v3
         with:
-          github_token: ${{ secrets.GITHUB_TOKEN }}
-          publish_dir: ./target/doc/
+          publish_dir: target/doc/
+          deploy_key: ${{ secrets.ACTIONS_DEPLOY_KEY }}
```
bilelmoussaoui, line 68: deploy_key can't be secrets.GITHUB_TOKEN?

.github/workflows/docs.yml
jsparber: `gtk-rs/gtk-rs` has no branch crate, or do i miss something.
I simple assumed that there where no releases since the move from `gtk-rs/gtk` repo.

.github/workflows/docs.yml
jsparber: We can't use GITHUB_TOKEN since the docs get pushed to a different repo and therefore it doesn't have the correct permissions. See: https://github.com/peaceiris/actions-gh-pages#%EF%B8%8F-set-ssh-private-key-deploy_key

bilelmoussaoui: Yeah except I don't think we need that extra repository for docs anymore, gh-pages branch should be more than enough iirc. But I would wait for @GuillaumeGomez's opinion on this one

.github/workflows/docs.yml
bilelmoussaoui: Yes, because we still haven't done a release with the new repository structure. No idea if that will change in the future. But I am definitely in favour of tagging new releases :-)

.github/workflows/docs.yml
jsparber: I wasn't talking about the `docs` repo. The docs are pushed to the webpage repo.

bilelmoussaoui: What would be different between using the `gh-pages` branch & a new repository here? iirc, it would work exactly the same way.

jsparber: it still leave the odd situation with `docs-src` and `docs` like we have right now: https://github.com/gtk-rs/gtk-rs.github.io/issues/105

jsparber: > What would be different between using the gh-pages branch & a new repository here?
There is no new repository, the docs are pushed directly to https://github.com/gtk-rs/gtk-rs.github.io

GuillaumeGomez: Like said in a few places: I want to keep the documentation repository on its own because it's a mess, whatever the way we use to handle it.

jsparber:
I can rework this so it pushes the docs to the `docs` repo instead of the webpage repo. 
This would then create:
- `https://gtk-rs.org/docs/stable`
- `https://gtk-rs.org/docs/unstable` for unstable builds

@GuillaumeGomez What do you think?

GuillaumeGomez:
I strongly approve this one! :)

GuillaumeGomez:
@sdroege Does this naming seems good to you?

sdroege:
Sounds good to me, yes. Can we set up a redirect for the old URLs so that there are not thousands of dead links suddenly? :)

jsparber:
> Sounds good to me, yes. Can we set up a redirect for the old URLs so that there are not thousands of dead links suddenly? :)

Not really, we may be able to add a javascript redirect, which definitely isn't great. We could also leave the stable docs as-is and add only `dev/unstable`, but i'm not a big fan of it.

GuillaumeGomez:
I'm closing it for now. Don't hesitate to re-open it once it's updated!

PR #326, opened 2021-02-27
Conversions between Value, PathBuf and OsString

nagisa:
I did an initial take on this. I can see a number of ways to improve on this PR, but all of the things seem fairly involved.

### DIFF
--- a/glib/src/translate.rs
+++ b/glib/src/translate.rs
@@ -674,6 +674,12 @@ impl<'a> ToGlibPtr<'a, *const c_char> for OsStr {
         let tmp = os_str_to_c(self);
         Stash(tmp.as_ptr(), tmp)
     }
+
+    #[inline]
+    fn to_glib_full(&self) -> *const c_char {
+        let stash = self.to_glib_none();
+        unsafe { ffi::g_strdup(stash.0) as *mut c_char }
+    }
 }
 
 impl<'a> ToGlibPtr<'a, *mut c_char> for OsStr {
@@ -683,6 +689,12 @@ impl<'a> ToGlibPtr<'a, *mut c_char> for OsStr {
     fn to_glib_none(&'a self) -> Stash<'a, *mut c_char, Self> {
         let tmp = os_str_to_c(self);
         Stash(tmp.as_ptr() as *mut c_char, tmp)
+    }
+
+    #[inline]
+    fn to_glib_full(&self) -> *mut c_char {
+        let stash = self.to_glib_none();
+        unsafe { ffi::g_strdup(stash.0) as *mut c_char }
     }
 }
 
@@ -694,6 +706,11 @@ impl<'a> ToGlibPtr<'a, *const c_char> for OsString {
         let tmp = os_str_to_c(self);
         Stash(tmp.as_ptr(), tmp)
     }
+
+    #[inline]
+    fn to_glib_full(&self) -> *const c_char {
+        OsStr::to_glib_full(&*self)
+    }
 }
 
 impl<'a> ToGlibPtr<'a, *mut c_char> for OsString {
@@ -703,6 +720,11 @@ impl<'a> ToGlibPtr<'a, *mut c_char> for OsString {
     fn to_glib_none(&'a self) -> Stash<'a, *mut c_char, Self> {
         let tmp = os_str_to_c(self);
         Stash(tmp.as_ptr() as *mut c_char, tmp)
+    }
+
+    #[inline]
+    fn to_glib_full(&self) -> *mut c_char {
+        OsStr::to_glib_full(&*self)
     }
 }
 
--- a/glib/src/translate.rs
+++ b/glib/src/translate.rs
@@ -604,6 +604,12 @@ impl<'a> ToGlibPtr<'a, *const c_char> for Path {
         let tmp = path_to_c(self);
         Stash(tmp.as_ptr(), tmp)
     }
+
+    #[inline]
+    fn to_glib_full(&self) -> *const c_char {
+        let stash = self.to_glib_none();
+        unsafe { ffi::g_strdup(stash.0) as *mut c_char }
+    }
 }
 
 impl<'a> ToGlibPtr<'a, *mut c_char> for Path {
@@ -613,6 +619,12 @@ impl<'a> ToGlibPtr<'a, *mut c_char> for Path {
     fn to_glib_none(&'a self) -> Stash<'a, *mut c_char, Self> {
         let tmp = path_to_c(self);
         Stash(tmp.as_ptr() as *mut c_char, tmp)
+    }
+
+    #[inline]
+    fn to_glib_full(&self) -> *mut c_char {
+        let stash = self.to_glib_none();
+        unsafe { ffi::g_strdup(stash.0) as *mut c_char }
     }
 }
 
@@ -624,6 +636,11 @@ impl<'a> ToGlibPtr<'a, *const c_char> for PathBuf {
         let tmp = path_to_c(self);
         Stash(tmp.as_ptr(), tmp)
     }
+
+    #[inline]
+    fn to_glib_full(&self) -> *const c_char {
+        Path::to_glib_full(&*self)
+    }
 }
 
 impl<'a> ToGlibPtr<'a, *mut c_char> for PathBuf {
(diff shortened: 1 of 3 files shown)


glib/src/translate.rs
```diff
@@ -604,6 +604,14 @@ impl<'a> ToGlibPtr<'a, *const c_char> for Path {
         let tmp = path_to_c(self);
         Stash(tmp.as_ptr(), tmp)
     }
+
+    #[inline]
+    fn to_glib_full(&self) -> *const c_char {
+        let stash = self.to_glib_none();
+        unsafe {
+            ffi::g_strdup(stash.0) as *mut c_char
```
nagisa, line 612: Since `to_glib_none` allocates anyway, it seems like it'd make more sense to implement `to_glib_none` in terms of `to_glib_full`, and not the other way around, potentially saving an allocation.

sdroege: Good catch, that's true (also for all the other string-y types). The main problem is that we then have to replicate the `NUL`-check of `CString`, or silently truncate at `NUL`. For now continuing with the current behaviour should be safest (and I think is also the correct behaviour here).

Do you want to add another commit on top of this to implement this?

glib/src/translate.rs
```diff
@@ -682,6 +731,11 @@ impl<'a> ToGlibPtr<'a, *mut c_char> for OsString {
         let tmp = os_str_to_c(self);
         Stash(tmp.as_ptr() as *mut c_char, tmp)
     }
+
+    #[inline]
+    fn to_glib_full(&self) -> *mut c_char {
+        OsStr::to_glib_full(&*self)
+    }
 }
```
sdroege, line 739: Ideally all the changes that add more FFI conversions for these types should be a separate commit

glib/src/types.rs
```diff
@@ -351,6 +351,10 @@ builtin!(f32, F32);
 builtin!(f64, F64);
 builtin!(str, STRING);
 builtin!(String, STRING);
+builtin!(std::ffi::OsString, STRING);
+builtin!(std::ffi::OsStr, STRING);
+builtin!(std::path::PathBuf, STRING);
+builtin!(std::path::Path, STRING);
```
sdroege, line 357: It's really sad that GLib has annotations for distinguishing `utf8` from `filename` strings but at the type system level (`GType`) it's all a string.

Fortunately it doesn't matter too much because a) who uses non-UTF8 locales on UNIXy systems nowadays and b) filenames are always UTF-8 in GLib on Windows.

glib/src/translate.rs
sdroege: Well there's also the Windows special case here: in GLib they're all UTF-8 but in Rust [they're](https://doc.rust-lang.org/std/ffi/struct.OsString.html)

> On Windows, strings are often arbitrary sequences of non-zero 16-bit values, interpreted as UTF-16 when it is valid to do so.

So that conversion also needs to be moved around a bit.

sdroege:
Looks generally good to me, thanks :) You'll also want to run `cargo fmt` once though.

> I can see a number of ways to improve on this PR, but all of the things seem fairly involved.

What other ways than the above do you see? Some of the duplication could be reduced by macros but that doesn't seem worth it for these few lines.

GuillaumeGomez (approved):
Seems good to me. I'm curious about the improvements you have in mind though!

nagisa:
So one thing is that I'd love for this conversion to also transparently work when the `Value` contains a `GString` type. Another expressed as the inline comment – I think it'd be great to reduce the amount of allocations we end up with during the conversion to the C string `Value`. Finally, panics (instead of `Result`s) when the string e.g. contains WTF-8 that can't easily become UTF-8 bother me a little, but changing that would be fairly hard AFAICT.

nagisa:
Ran cargo fmt and split up the changes into more self-contained commits.

sdroege:
> Writing the commit message for `impl SetValue/FromValue for OsStr{,ing}` I began wondering if we really want to impl `OsStr{,ing}` variants right now. In particular I suspect that its going to block us from making these represented as `GString` by default in the future.
> > These correspond to UTF-8 encoded C string Value type currently, as an
> > arbitrary choice. A better option would be a GString, which can contain
> > arbitrary byte strings, thus potentially supporting UTF-16 encoded data
> > of Windows as well.

In GLib on Windows, all these strings (paths, environment variables, ...) are actually UTF-8 so that's correct. On UNIXy systems it should already pass those through in whatever encoding they're in, both here and in GLib.

What problems do you see?

sdroege:
> What problems do you see?

@nagisa How should we proceed here? :)

nagisa:
I don't know, honestly.

From this PR it is very apparent to me that the string situation is a big mess and it is way over my head to resolve. I would love it if interior null bytes within a `OsString` did not explode in some way, but it is also pretty apparent to me that people would end up being surprised if a `GString` came out from that kind of conversion.

I'll leave a decision up to you all on how to proceed here. Feel free to build on my PR, implement something else entirely or whatever else – you all have been working on glib/gtk/etc for far longer than I have been and likely know better than me what's the best course of action is here.

sdroege:
I think the right thing to do here is to implement the `Value` traits for `PathBuf` and `OsString` here like you did.

The overall string situation is really hard to use and generally a problem in all C/FFI situations with Rust and I would consider that out of scope for this PR at least. I'm not sure how to solve that nicely either, other than panicking on interior `\0` and similar runtime solutions.

GuillaumeGomez:
The main issue here is that Rust has specific types whereas glib doesn't. Even though it's not great to have `GString` for everything, I think it's the best way for now. The next thing we can do (since we have conversions thanks to this PR) would be to enforce that each function/method is returning the appropriate Rust type (like `PathBuf` for pathes, etc).

sdroege:
> The main issue here is that Rust has specific types whereas glib doesn't. Even though it's not great to have `GString` for everything, I think it's the best way for now. The next thing we can do (since we have conversions thanks to this PR) would be to enforce that each function/method is returning the appropriate Rust type (like `PathBuf` for pathes, etc).

You can't use a `GString` for a `glib::Value` that contains a path/filename though. That's why we would need the value trait impls that are part of this PR.

glib/src/translate.rs
```diff
@@ -652,6 +674,12 @@ impl<'a> ToGlibPtr<'a, *const c_char> for OsStr {
         let tmp = os_str_to_c(self);
         Stash(tmp.as_ptr(), tmp)
     }
+
+    #[inline]
+    fn to_glib_full(&self) -> *const c_char {
+        let stash = self.to_glib_none();
+        unsafe { ffi::g_strdup(stash.0) as *mut c_char }
```
GuillaumeGomez, line 681:
```suggestion
        unsafe { ffi::g_strdup(stash.0) as *const c_char }
```

glib/src/value.rs
```diff
@@ -1162,6 +1222,51 @@ mod tests {
...
+
+        let path = Path::new("apple");
+        let pathbuf = path.to_owned();
+        let path_value = path.to_value();
+        let pathbuf_value = pathbuf.to_value();
+
+        assert_eq!(pathbuf_value.get::<PathBuf>(), Ok(Some(pathbuf.clone())));
+        assert_eq!(path_value.get::<PathBuf>(), Ok(Some(pathbuf.clone())));
+        // Drop the backing storage, and compare with str to ensure valid memory management.
+        drop(pathbuf);
```
GuillaumeGomez, line 1237: Even though it's dropped, the memory can remain accessible (depending on the page allocation, etc). Not sure if this part of the test is that useful... Or maybe you could simply update the path value?

glib/src/value.rs
```diff
@@ -1162,6 +1222,51 @@ mod tests {
...
+        use std::ffi::{OsStr, OsString};
+        let osstr = OsStr::new("banana");
+        let osstring = osstr.to_owned();
+        let osstr_value = osstr.to_value();
+        let osstring_value = osstring.to_value();
+
+        assert_eq!(osstring_value.get::<OsString>(), Ok(Some(osstring.clone())));
+        assert_eq!(osstr_value.get::<OsString>(), Ok(Some(osstring.clone())));
+        // Drop the backing storage, and compare with str to ensure valid memory management.
+        drop(osstring);
```
GuillaumeGomez, line 1259: Same as above.

glib/src/value.rs
sdroege: `valgrind` would complain, and at least I'm regularly running our tests with `valgrind`

GuillaumeGomez: `valgrind` doesn't work on mac and it's really annoying to make it run with `cargo test`. Updating the value and checking that it was updated seems simpler.

sdroege: Yes

GuillaumeGomez:

PR #331, opened 2021-03-01
Implement ListBoxRowImplExt for activate method

matzipan:


gtk/src/subclass/list_box_row.rs
```diff
@@ -19,6 +19,21 @@ pub trait ListBoxRowImplExt: ObjectSubclass {
     fn list_box_row_activate(&self, list_box_row: &Self::Type);
 }
 
+impl<T: ListBoxRowImpl> ListBoxRowImplExt for T {
+    fn list_box_row_activate(&self, list_box_row: &Self::Type) {
```
bilelmoussaoui, line 23: This should be renamed to `parent_activate` to follow the same naming conventions we use everywhere.

matzipan: Should this be part of this PR? Or a subsequent one?

matzipan: I've done it here.

bilelmoussaoui: As part of this PR is fine

bilelmoussaoui (approved):

Issue #322, opened 2021-02-25
Magic proc macro to simplify implementing ObjectSubclass


ids1024:
Implementing `ObjectSubclass` is rather verbose. In fact more so than with the previous release due to the addition of `Type` and `Interfaces`:

```rust
impl ObjectSubclass for FooBarInner {
    const NAME: &'static str = "FooBar";

    type ParentType = gtk::Widget;
    type Type = FooBar;
    type Interfaces = ();

    type Instance = subclass::simple::InstanceStruct<Self>;
    type Class = subclass::simple::ClassStruct<Self>;

    glib::object_subclass!();

    fn new() -> Self {
        Self::default()
    }
}
```

There are a few annoying things there that seem unnecessary, but there's no way to handle them better in standard Rust (yet):
* `Interfaces`, `Instance`, and `Class` should use associated type defaults, but that isn't stable.
* Calling `glib::object_subclass!();` here is annoying, but there's no way in Rust to associate static data with instances of a trait.
* Recently it was suggested on IRC that `new()` should default to `Default::default()`, but there's no way to do that without requiring a `Default` bound for all `ObjectSubclass` implementations.

Perhaps a magic proc macro could handle of of this:

```rust
#[glib::object_subclass]
impl ObjectSubclass for FooBarInner {
    const NAME: &'static str = "FooBar";
    type ParentType = gtk::Widget;
    type Type = FooBar;
}
```

This seems a lot neater, and is sufficient for most cases where I've implemented `ObjectSubclass`. It would still allow custom values of the associated type, or custom `new()`.

sdroege:
That seems useful, yes. A workaround for the missing language features :)

ids1024:
Closing now that https://github.com/gtk-rs/gtk-rs/pull/335 is merged.

Issue #215, opened 2020-12-14
Exposing type_data() together with register_type() is unsound


YaLTeR:
```rust
#![forbid(unsafe_code)]
use std::ptr::NonNull;

use glib::glib_wrapper;
use glib::subclass;
use glib::subclass::prelude::*;

#[derive(Debug)]
pub struct FooPrivate {}

impl ObjectSubclass for FooPrivate {
    const NAME: &'static str = "Foo";
    type Type = Foo;
    type ParentType = glib::Object;
    type Instance = glib::subclass::simple::InstanceStruct<Self>;
    type Class = glib::subclass::simple::ClassStruct<Self>;

    fn type_data() -> NonNull<subclass::TypeData> {
        NonNull::dangling()
    }

    fn get_type() -> glib::Type {
        unimplemented!()
    }
}

impl ObjectImpl for FooPrivate {}

glib_wrapper! {
    pub struct Foo(ObjectSubclass<FooPrivate>);
}

fn main() {
    subclass::register_type::<FooPrivate>();
}
```

sdroege:
I guess we should mark a few more things `unsafe`. Want to provide a PR for that? The functions generated by the `glib_object_subclass!` macro should be.

Also I'm surprised `#![forbid(unsafe_code)]` works here: `glib_wrapper!` expands into quite a bit of `unsafe` code.

YaLTeR:
> I guess we should mark a few more things `unsafe`. Want to provide a PR for that? The functions generated by the `glib_object_subclass!` macro should be.

Isn't `unsafe` on functions "unsafe to call" rather than "unsafe to implement" (as it is on traits)? Because I think `type_data()` is "unsafe to implement".

> Also I'm surprised `#![forbid(unsafe_code)]` works here: `glib_wrapper!` expands into quite a bit of `unsafe` code.

My guess is that it doesn't take other crate macro expansion into account exactly for use-cases like in glib-rs. Similar to how you can export a safe API from a crate and call unsafe under the hood.

sdroege:
> Isn't `unsafe` on functions "unsafe to call" rather than "unsafe to implement" (as it is on traits)? Because I think `type_data()` is "unsafe to implement".

I think we'd have to move those into a new `unsafe` trait (and could then replace the macro with a derive macro). Not sure how that could work in practice.

ids1024:
[...]

Issue #41, opened 2020-07-27
Remove unused features in syn


GuillaumeGomez:
A lot of features in `syn` are enabled by default even though we don't use them. We should check which features we actually use and opt out all the others.

pbor:
I opened https://github.com/gtk-rs/gtk-rs/pull/432 mostly for consistency.

I do not think there is much else we can do, since we depend from many other crates like `futures-macro` which also enable the `full` feature so even if we further customize the feature set, they would still get compiled:

https://github.com/rust-lang/futures-rs/blob/master/futures-macro/Cargo.toml

sdroege:
Why do we depend on `futures-macro` (except for the tests via dev-dependencies)?

bilelmoussaoui:
> Why do we depend on `futures-macro` (except for the tests via dev-dependencies)?

It comes from `futures-util` with the default `async-await-macro` feature. Do we need that one?

sdroege:
No we don't. But why do we depend on `futures-util` still outside of `dev-dependencies` for the tests?

bilelmoussaoui:
Aha, it comes from the default features of `futures-executor` https://github.com/rust-lang/futures-rs/blob/master/futures-executor/Cargo.toml#L16

sdroege:
So should make that `default-features = false` too, but we also only use that from `dev-dependencies` too :)

bilelmoussaoui:
It's used in https://github.com/gtk-rs/gtk-rs/blob/master/glib/src/main_context_futures.rs

PR #335, opened 2021-03-04
glib: `#[object_subclass]` proc macro.

ids1024:
Proposed in https://github.com/gtk-rs/gtk-rs/issues/322.

This replaces `object_subclass!` while also magically adding support for associated type defaults, and making `new()` call `Default::default()`.

A draft because I still need to update the other crates to use this, test it a bit more, and perhaps tweak the errors it produces and the documentation.

A proc macro like this here could be useful for other things. For instance, to address https://github.com/gtk-rs/gtk-rs/issues/215 by splitting `type_data` and `get_type` into a separate `unsafe trait`. That is not implemented here.

### DIFF
--- a/glib-macros/Cargo.toml
+++ b/glib-macros/Cargo.toml
@@ -16,7 +16,7 @@ heck = "0.3"
 proc-macro-error = "1.0"
 proc-macro2 = "1.0"
 quote = "1.0"
-syn = "1.0"
+syn = { version = "1.0", features = ["full"] }
 proc-macro-crate = "0.1"
 
 [lib]
--- a/glib-macros/src/object_subclass_attribute.rs
+++ b/glib-macros/src/object_subclass_attribute.rs
@@ -1,0 +1,94 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use proc_macro2::TokenStream;
+use proc_macro_error::abort_call_site;
+use quote::quote;
+
+pub const WRONG_PLACE_MSG: &str =
+    "This macro should be used on `impl` block for `glib::ObjectSubclass` trait";
+
+pub fn impl_object_subclass(input: &syn::ItemImpl) -> TokenStream {
+    let mut has_new = false;
+    let mut has_interfaces = false;
+    let mut has_instance = false;
+    let mut has_class = false;
+    for item in &input.items {
+        match item {
+            syn::ImplItem::Method(method) => {
+                let name = method.sig.ident.to_string();
+                if name == "new" || name == "with_class" {
+                    has_new = true;
+                }
+            }
+            syn::ImplItem::Type(type_) => {
+                let name = type_.ident.to_string();
+                if name == "Interfaces" {
+                    has_interfaces = true;
+                } else if name == "Instance" {
+                    has_instance = true;
+                } else if name == "Class" {
+                    has_class = true;
+                }
+            }
+            _ => {}
+        }
+    }
+
+    let syn::ItemImpl {
+        attrs,
+        generics,
+        trait_,
+        self_ty,
+        items,
+        ..
+    } = &input;
+
+    let interfaces_opt = if has_interfaces {
+        None
+    } else {
+        Some(quote!(
+            type Interfaces = ();
+        ))
+    };
+
+    let new_opt = if has_new {
+        None
+    } else {
+        Some(quote! {
+            fn new() -> Self {
+                std::default::Default::default()
+            }
+        })
+    };
+
+    let crate_ident = crate::utils::crate_ident_new();
+
+    let class_opt = if has_class {
+        None
+    } else {
+        Some(quote!(type Class = #crate_ident::subclass::simple::ClassStruct<Self>;))
+    };
+
+    let instance_opt = if has_instance {
+        None
+    } else {
+        Some(quote!(type Instance = #crate_ident::subclass::simple::InstanceStruct<Self>;))
+    };
+
+    let trait_path = match &trait_ {
+        Some(path) => &path.1,
+        None => abort_call_site!(WRONG_PLACE_MSG),
+    };
+
+    quote! {
+        #(#attrs)*
+        impl#generics #trait_path for #self_ty {
+            #interfaces_opt
+            #class_opt
+            #instance_opt
+            #new_opt
+            #crate_ident::object_subclass_internal!();
+            #(#items)*
+        }
+    }
+}
--- a/glib/src/subclass/mod.rs
+++ b/glib/src/subclass/mod.rs
@@ -68,6 +68,7 @@
 //!     // ObjectSubclass is the trait that defines the new type and
 //!     // contains all information needed by the GObject type system,
 //!     // including the new type's name, parent type, etc.
+//!     #[glib::object_subclass]
 //!     impl ObjectSubclass for SimpleObject {
 //!         // This type name must be unique per process.
 //!         const NAME: &'static str = "SimpleObject";
@@ -78,16 +79,6 @@
 //!
 //!         // Interfaces this type implements
 //!         type Interfaces = ();
-//!
-//!         // The C/FFI instance and class structs. The simple ones
-//!         // are enough in most cases and more is only needed to
-//!         // expose public instance fields to C APIs or to provide
-//!         // new virtual methods for subclasses of this type.
-//!         type Instance = subclass::simple::InstanceStruct<Self>;
-//!         type Class = subclass::simple::ClassStruct<Self>;
-//!
-//!         // This macro defines some boilerplate.
-//!         glib::object_subclass!();
 //!
 //!         // Called every time a new instance is created. This should return
 //!         // a new instance of our type with its basic values.
--- a/glib/src/subclass/object.rs
+++ b/glib/src/subclass/object.rs
@@ -223,6 +223,7 @@ mod test {
     use super::super::super::subclass;
     use super::super::super::value::{ToValue, Value};
     use super::*;
+    use crate as glib;
     use crate::{StaticType, Type};
 
     use std::cell::RefCell;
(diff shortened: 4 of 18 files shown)


glib/src/subclass/mod.rs
```diff
@@ -86,9 +87,6 @@
 //!         type Instance = subclass::simple::InstanceStruct<Self>;
 //!         type Class = subclass::simple::ClassStruct<Self>;
```
bilelmoussaoui, line 88:
```suggestion
```
Those can be omitted now, no?

glib/src/subclass/object.rs
```diff
@@ -224,54 +224,39 @@ mod test {
...
     mod imp {
         use super::*;
 
         // A dummy `Object` to test setting an `Object` property and returning an `Object` in signals
+        #[derive(Default)]
         pub struct ChildObject;
+
+        #[glib::object_subclass]
         impl ObjectSubclass for ChildObject {
             const NAME: &'static str = "ChildObject";
```
bilelmoussaoui, line 240: I wonder if we could have a default for the `NAME` as well?

glib/src/subclass/object.rs
```diff
@@ -224,54 +224,39 @@ mod test {
...
     use crate::{StaticType, Type};
+    use crate as glib;
 
     use std::cell::RefCell;
 
     mod imp {
         use super::*;
 
         // A dummy `Object` to test setting an `Object` property and returning an `Object` in signals
+        #[derive(Default)]
```
sdroege, line 235: Does it only require `Default` if no `fn new()` or `fn with_class()` is provided in the `impl` block? That would be great

glib/src/subclass/object.rs
```diff
@@ -224,54 +224,39 @@ mod test {
...
         pub struct ChildObject;
+
+        #[glib::object_subclass]
         impl ObjectSubclass for ChildObject {
             const NAME: &'static str = "ChildObject";
             type Type = super::ChildObject;
             type ParentType = Object;
-            type Interfaces = ();
-            type Instance = subclass::simple::InstanceStruct<Self>;
-            type Class = subclass::simple::ClassStruct<Self>;
```
sdroege, line 241: Does it make these optional but if they're provided nonetheless in the `impl` block it will use those instead?

glib/src/subclass/types.rs
```diff
@@ -299,7 +299,7 @@ impl TypeData {
 /// Macro for boilerplate of [`ObjectSubclass`] implementations.
 ///
 /// [`ObjectSubclass`]: subclass/types/trait.ObjectSubclass.html
-macro_rules! object_subclass {
+macro_rules! object_subclass_internal {
```
sdroege, line 302: Let's still keep this public API for people who don't want or can't use the macro. Doesn't hurt :)

glib/src/subclass/object.rs
sdroege: That default should probably include the crate and module name though

ids1024: Oh, I hadn't even thought of that. It would be great, but I don't see a way for a proc macro to get the equivalent of `std::module_path!()` (which it could manipulate to generate a name). `Span` can get the source file path (with unstable methods), but I don't see anything that could really be used for this.

It can be added later though, if we find a way or Rust adds one.

glib/src/subclass/object.rs
ids1024: Yes, that's the idea. The proc macro iterates through its input and sees if `new` or `with_class` are provided. If it finds neither, it adds a `new` definition.

The only issue is that I don't know if there's a way to print a nice warning specifically saying to implement `Default` or `new` (it errors saying the type doesn't implement `Default`, in the case where no `new` or `with_class` is defined, and `Default` isn't implemented).

glib/src/subclass/object.rs
ids1024: Yes.

glib/src/subclass/types.rs
ids1024: If so, they can't have the same name.

sdroege: Same as?

glib/src/subclass/object.rs
sdroege: As macros don't know any type information that won't be possible. You can't know whether `Default` is implemented or not

glib/src/subclass/types.rs
ids1024: It could be kept as part of the public API, but it should be exactly the same as the proc macro, when the associated types and functions are provided and it doesn't use the defaults.

glib/src/subclass/mod.rs
ids1024: Probably. I guess the documentation should cover the way they have default values and how they can be set to something else, but they can probably be omitted in a simple example.

glib/src/subclass/types.rs
sdroege: Yes

ids1024: > Same as?

A proc macro called `#[glib::object_subclass]` and a declarative macro `glib::object_subclass!`. They seem to be in the same namespace.

ids1024: I don't think there should be any reason not to use the proc macro. If the caller specifies `Interfaces`, `Instance`, `Class`, and either `new` or `with_class`, the macro does nothing but add the `object_subclass_internal!` call. Perhaps someone doesn't like the fact there are implicit default values, but then if Rust allowed things like associated type defaults on stable, we would already be using that without a macro.

Maybe someone can think of a reason not to want to use the proc macro, but I don't really see one.

glib/src/subclass/object.rs
ids1024: Would things like `crate::foo::bar::SomeObject` work and be reasonable as a default value for `NAME`? I think that could be done by having the proc macro generate `concat!(std::module_path!(), "::", type_name)`. But if it had to remove/replace the `::` in the module path, it couldn't do that at compile time.

glib/src/subclass/types.rs
sdroege: Let's not worry about that then until someone complains

glib/src/subclass/object.rs
sdroege: I don't think that's a valid GType name

ids1024: [It appears so](https://developer.gnome.org/gobject/stable/gtype-conventions.html):
>Type names (including object names) must be at least three characters long and start with ‘a–z’, ‘A–Z’ or ‘_’.

ids1024: So probably not really possible to provide a default for `NAME` now, but it would be great to add later if someone finds a way to provide a good default.

sdroege: Or go with just the type name. At least for applications that's often going to be fine

ids1024: Often, but it doesn't seem great to me to end up with a crash at run time when a user uses too generic a name for their type (I assume that's what would happen), which isn't normally a concern in Rust.

sdroege: It would cause an assertion, yes. I agree

glib-macros/Cargo.toml
```diff
@@ -16,7 +16,7 @@ heck = "0.3"
 proc-macro-error = "1.0"
 proc-macro2 = "1.0"
 quote = "1.0"
-syn = "1.0"
+syn = { version = "1.0", features = ["full"] }
```
sdroege, line 19: It's probably better to only opt-in to the features we actually use instead of enabling all :) See also https://github.com/gtk-rs/gtk-rs/issues/41

glib-macros/src/object_subclass_attribute.rs
```diff
@@ -0,0 +1,88 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use proc_macro2::TokenStream;
+use quote::quote;
+
+pub fn impl_object_subclass(input: &syn::ItemImpl) -> TokenStream {
+    let mut has_new = false;
```
sdroege, line 7: You probably want to check that this is only added to an `impl .*ObjectSubclass for T` and not on anything else

glib/src/subclass/mod.rs
sdroege: Yeah, that's something for the docs of the macro. Describing what it does and what the "defaults" are that it's going to use

glib/src/subclass/object.rs
```diff
@@ -231,47 +232,31 @@ mod test {
...
             construct_name: RefCell<Option<String>>,
             constructed: RefCell<bool>,
         }
 
+        #[glib::object_subclass]
         impl ObjectSubclass for SimpleObject {
             const NAME: &'static str = "SimpleObject";
             type Type = super::SimpleObject;
             type ParentType = Object;
             type Interfaces = (DummyInterface,);
```
sdroege, line 259: If we are friendly we could also allow for a single interface either
  - Passing it in without the trailing comma and add one then
  - Give a more friendly error if the trailing comma is missing

sdroege:
Looks good to me otherwise (missing docs and the things mentioned above). Is there anything else you want to do before you consider it ready as a first version?

glib-macros/src/object_subclass_attribute.rs
ids1024: Technically someone could `use glib::ObjectSubclass as FooBar`, and we'd have no way to know that. But yeah, it's probably reasonable to restrict it to `ObjectSubclass` or paths ending in that.

glib-macros/Cargo.toml
ids1024: According to https://docs.rs/syn/1.0.61/syn/enum.ImplItem.html:
> This type is available only if Syn is built with the "full" feature.

I'm not that familiar with Syn, but looking at it's `Cargo.toml`, "full" doesn't actually imply any other features, somewhat counterintuitively. It [actually](https://docs.rs/syn/1.0.61/syn/index.html#optional-features) means this:
> full — Data structures for representing the syntax tree of all valid Rust source code, including items and expressions.

sdroege: How confusing

glib-macros/src/object_subclass_attribute.rs
sdroege: Well, at least check that it's some trait impl block :)

glib/src/subclass/object.rs
ids1024: Perhaps it could offer a more friendly error message. Though I'm inclined not to add any special handling here. Personally I don't really mind how it's specified now since I'm familiar with how single element tuples are written, and I like that everything the macro is currently doing might ideally be a feature that a future Rust version could natively support.

sdroege: Yeah I agree. A more friendly error then maybe, but that can also happen in a future PR if it's not trivial to add.

ids1024:
Added a custom error message when it isn't used on an `impl` block for a trait, and a doc comment for the macro. I'm sure the documentation could be improved (for this macro and subclassing in general). But I'm fairly happy with how this is working now.

GuillaumeGomez (approved):
Looks good to me! Only thing I'm worried about: how much does it impact the compilation time?

bilelmoussaoui (approved):
Looks good to me. This is a nice step forward simplifying the subclassing code a bit, thanks for that!

sdroege:
I'll review it tonight but we should first get PRs ready for gstreamer-rs/gst-plugins-rs and gtk4-rs or otherwise they will stop building once this is merged :) @bilelmoussaoui you do gtk4-rs, I do the others later?

sdroege (approved):

sdroege:
Hmm, sad. For GStreamer elements I still need to specify `type Instance = gst::subclass::ElementInstanceStruct<Self>` of course.

@ids1024 Any ideas how to make that nicer?

ids1024:
Hm. I can't really think of a way to improve that. The `glib` crate probably shouldn't have any logic specific for gstreamer, not that I could think of a clean way to do that anyway. Maybe the default could in some way depend on the parent type, but doing that through the type system would probably at least require GATs.

sdroege:
I found a clean way that's completely independent from this here. I noticed that using a special instance struct for it is broken anyway in special cases, so I'm going to do a different solution.

I'll update my repos for this macro and then implement that after merging this. It will be a new addition to `ObjectSubclass` and `IsSubclassable`

PR #338, opened 2021-03-07
glib: Move `type_data()` and `get_type()` to new `unsafe trait`

ids1024:
Since this is implemented by `#[object_subclass]`, it mostly doesn't require changes in dependent crates.

Fixes https://github.com/gtk-rs/gtk-rs/issues/215.


GuillaumeGomez (approved):

glib-macros/src/object_subclass_attribute.rs
```diff
@@ -87,8 +87,36 @@ pub fn impl_object_subclass(input: &syn::ItemImpl) -> TokenStream {
             #class_opt
             #instance_opt
             #new_opt
-            #crate_ident::object_subclass_internal!();
             #(#items)*
         }
+
+        unsafe impl #crate_ident::subclass::types::ObjectSubclassType for #self_ty {
```
sdroege, line 93: The same change is also needed for `ObjectInterface` and the corresponding macro. Can you do that?

Otherwise looks like a good improvement to me, thanks :)

sdroege (approved):

bilelmoussaoui (approved):

PR #340, opened 2021-03-07
CI: use the new docker image

bilelmoussaoui:
Won't pass, requires making the new image public

cc @GuillaumeGomez

### DIFF
--- a/.github/workflows/image.yml
+++ b/.github/workflows/image.yml
@@ -4,8 +4,6 @@ on:
   push:
     paths:
       - "Dockerfile"
-    branches:
-      - "master"
 
 jobs:
   docker:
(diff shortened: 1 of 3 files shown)


sdroege (approved):
Seems good to me, @GuillaumeGomez ?

.github/workflows/image.yml
```diff
@@ -4,8 +4,6 @@ on:
   push:
     paths:
       - "Dockerfile"
-    branches:
```
GuillaumeGomez, line 7: Don't you want this to be done only when pushed on the master branch?

bilelmoussaoui: We want the docker image to be updated only when we modify the dockerfile. There's no reason to re-build the image all the time

GuillaumeGomez (approved):

PR #342, opened 2021-03-09
Add a way for the subclassing infrastructure to add per-instance management data

sdroege:


GuillaumeGomez (approved):
I know pretty much nothing about subclassing so not sure if my review is worth anything but at least the code looks good.

PR #344, opened 2021-03-10
object: Return a pointer from get_qdata, not a reference

heftig:
A returned reference can become dangling by a concurrent `set_data` or
`steal_data`. Return a pointer instead.


GuillaumeGomez (approved):
Thanks!

PR #349, opened 2021-03-11
glib: Add _once variants to many methods

### LINKED ISSUE #341: glib::timeout_add_local accept FnOnce
I sometimes find myself in the situation of adding a timeout/callback that will be called exactly once. But the respective `glib` methods accept a `FnMut` at best. One could wrap a `FnOnce` like this and expose this as an alternative (untested code):

```rust
fn wrap_once_local(func: F) -> impl (FnMut() -> Continue + 'static)
where
    F: FnOnce() 'static,  
{
    let mut func = Some(func);
    move || {
        func.take().unwrap();
        Continue(false)
    }
}
```

piegamesde:
Closes #341. Please don't mind the copy-paste blueprints in the code ^^

CC @sdroege

### DIFF
--- a/glib/src/source.rs
+++ b/glib/src/source.rs
@@ -142,6 +142,41 @@ fn into_raw_unix_fd<F: FnMut(RawFd, IOCondition) -> Continue + 'static>(func: F)
     Box::into_raw(func) as gpointer
 }
 
+/// Transform a generic FnOnce into a closure that can be used as callback in various glib methods
+///
+/// The resulting function can only be called once and will panic otherwise. It will return `Continue(false)`
+/// in order to prevent being called twice.
+#[inline(always)]
+fn fnmut_callback_wrapper(
+    func: impl FnOnce() + Send + 'static,
+) -> impl FnMut() -> Continue + Send + 'static {
+    let mut func = Some(func);
+    move || {
+        func.take()
+            .expect("GSource closure called after returning glib::Continue(false)");
+        Continue(false)
+    }
+}
+
+/// Transform a generic FnOnce into a closure that can be used as callback in various glib methods
+///
+/// The resulting function can only be called once and will panic otherwise. It will return `Continue(false)`
+/// in order to prevent being called twice.
+///
+/// Different to `fnmut_callback_wrapper()`, this does not require `func` to be
+/// `Send` but can only be called from the thread that owns the main context.
+#[inline(always)]
+fn fnmut_callback_wrapper_local(
+    func: impl FnOnce() + 'static,
+) -> impl FnMut() -> Continue + 'static {
+    let mut func = Some(func);
+    move || {
+        func.take()
+            .expect("GSource closure called after returning glib::Continue(false)");
+        Continue(false)
+    }
+}
+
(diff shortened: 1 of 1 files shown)


glib/src/source.rs
```diff
@@ -142,6 +142,56 @@ fn into_raw_unix_fd<F: FnMut(RawFd, IOCondition) -> Continue + 'static>(func: F)
     Box::into_raw(func) as gpointer
 }
 
+///
+/// In comparisong to `timeout_add()`, this only requires `func` to be
+/// `FnOnce`, and will automatically return `Continue(false)`.
+where
+    F: FnOnce() + Send + 'static,
```
sdroege, line 149: Something seemed to have gone missing here

glib/src/source.rs
```diff
@@ -142,6 +142,56 @@ fn into_raw_unix_fd<F: FnMut(RawFd, IOCondition) -> Continue + 'static>(func: F)
...
+/// in order to prevent being called twice.
+#[inline(always)]
+fn fnmut_callback_wrapper(func: F) -> G
+where
+    F: FnOnce() + Send + 'static,
+    G: FnMut() -> Continue + Send + 'static,
+{
+    let mut func = Some(func);
+    move || {
+        func.take().unwrap();
```
sdroege, line 164:
```suggestion
        func.take().expect("GSource closure called after returning glib::Continue(false)");
```

Or something like that

glib/src/source.rs
```diff
@@ -142,6 +142,56 @@ fn into_raw_unix_fd<F: FnMut(RawFd, IOCondition) -> Continue + 'static>(func: F)
...
+        Continue(false)
+    }
+}
+
+///
+/// In comparisong to `timeout_add()`, this only requires `func` to be
```
sdroege, line 170:
```suggestion
/// In comparison to `timeout_add()`, this only requires `func` to be
```

glib/src/source.rs
```diff
@@ -142,6 +142,56 @@ fn into_raw_unix_fd<F: FnMut(RawFd, IOCondition) -> Continue + 'static>(func: F)
...
+/// `FnOnce`, and will automatically return `Continue(false)`.
+where
+    F: FnOnce() + 'static,
```
sdroege, line 173: Also here something went missing

sdroege (commented):
Otherwise the approach looks good, just need to get it to compile :)

glib/src/source.rs
piegamesde: No, that's my copy-paste blueprint that I accidentally commited as well.

piegamesde:
I had to replace the generics with `impl Trait` types, otherwise the compiler didn't infer the types correctly for some reason.

sdroege:
> I had to replace the generics with `impl Trait` types, otherwise the compiler didn't infer the types correctly for some reason.

That's strange. This should be exactly the same apart from not having the generic types not nameable. But this way is fine too, all the same!

sdroege (approved):
Looks good to me, can you squash the commits?

GuillaumeGomez (approved):

piegamesde:
Done.

PR #350, opened 2021-03-12
glib: add ObjectExt::connect_id & other variants

### LINKED ISSUE #320: Add Object::connect_id() and similar that work based on the signal id instead of a string signal name
CC @bilelmoussaoui

bilelmoussaoui:
Similar to connect/connect_local/connect_unsafe but takes a SignalId instead of a signal name
Fixes #320

### DIFF
--- a/glib/src/object.rs
+++ b/glib/src/object.rs
@@ -1839,6 +1839,15 @@ impl<T: ObjectType> ObjectExt for T {
         let type_ = self.get_type();
         let return_type: Type = signal_query.return_type().into();
         let signal_name = signal_id.name();
+
+        let signal_query_type = signal_query.type_();
+        if !type_.is_a(signal_query_type) {
+            panic!(
+                "Signal '{}' of type '{}' but got type '{}'",
+                signal_name, type_, signal_query_type
+            );
+        }
+
         let closure = if return_type == Type::UNIT {
             Closure::new_unsafe(move |values| {
                 let ret = callback(values);
--- a/glib/src/object.rs
+++ b/glib/src/object.rs
@@ -1254,6 +1254,17 @@ pub trait ObjectExt: ObjectType {
     where
         N: Into<&'a str>,
         F: Fn(&[Value]) -> Option<Value> + Send + Sync + 'static;
+    /// Same as `connect` but takes a `SignalId` instead of a signal name.
+    fn connect_id<F>(
+        &self,
+        signal_id: SignalId,
+        details: Option<Quark>,
+        after: bool,
+        callback: F,
+    ) -> Result<SignalHandlerId, BoolError>
+    where
+        F: Fn(&[Value]) -> Option<Value> + Send + Sync + 'static;
+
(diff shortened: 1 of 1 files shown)


glib/src/object.rs
```diff
@@ -1843,6 +1899,19 @@ impl<T: ObjectType> ObjectExt for T {
...
+        &self,
+        signal_id: SignalId,
+        after: bool,
+        callback: F,
+    ) -> Result<SignalHandlerId, BoolError>
+    where
+        F: Fn(&[Value]) -> Option<Value>,
+    {
+        let signal_name = signal_id.name();
+        self.connect_unsafe(signal_name, after, callback)
```
sdroege, line 1912: That's rather useless :) You already have the id here, get the name, then query based on the name again to get the id again.

Can you refactor this a bit so that `connect_unsafe()` calls `connect_unsafe_id()` after querying the signal id?

sdroege (changes requested):

glib/src/object.rs
bilelmoussaoui: Ah right, but we still need the signal details which we can only get by calling `SignalId::parse_name` I think?

bilelmoussaoui: Unless we want the _id ones to take a `details: Option<&str>` param. This way we can parse the signal name, and call `connect_unsafe_id`. What do you think? :)

sdroege: You need to take the details as quark parameter

sdroege: In short: `connect_id()` takes `SignalId` and `Option<Quark>`, `connect()` takes a single `&str` (which is then parsed into the previous two things)

bilelmoussaoui: Makes sense, i will implement that :)

glib/src/object.rs
```diff
@@ -1756,12 +1820,25 @@ impl<T: ObjectType> ObjectExt for T {
...
+        signal_id: SignalId,
+        details: Option<Quark>,
+        after: bool,
+        callback: F,
+    ) -> Result<SignalHandlerId, BoolError>
+    where
+        F: Fn(&[Value]) -> Option<Value>,
+    {
+        let signal_query = signal_id.query();
+        let type_ = self.get_type();
```
sdroege, line 1839: You could check here that the query gives a type that is compatible with `self`

glib/src/object.rs
```diff
@@ -1839,6 +1839,15 @@ impl<T: ObjectType> ObjectExt for T {
         let type_ = self.get_type();
         let return_type: Type = signal_query.return_type().into();
         let signal_name = signal_id.name();
+
+        let signal_query_type = signal_query.type_();
+        if type_ != signal_query_type {
```
sdroege, line 1844: Use `Type::is_a()` here :)

sdroege (changes requested):

sdroege (approved):

Issue #348, opened 2021-03-11
glib::Class also allows casting to interfaces, which is wrong


sdroege:
`upcast_ref()` etc only looks at the `IsA` relationship but that's also true for interfaces. But for interfaces the "class" (vtable) struct has to be retrieved differently and is not just a cast of the proper class struct.

This can lead to crashes in safe code.

CC @ids1024

sdroege:
`glib_wrapper!` should probably implement yet another new trait that marks things as classes vs. interfaces. The macro knows about that already, just the type system doesn't.

sdroege:
Same issue for `Object::new()`, can be fixed the same way. I'm on it :)

PR #352, opened 2021-03-13
glib: Distinguish between classes and interfaces at the type level

sdroege:
Without this it was possible to cast from a Class<T> to a Class<U> where
U is an interface type. This doesn't work as interface structs are to be
retrieved differently than just casting and would cause memory unsafety
issues.

Fixes https://github.com/gtk-rs/gtk-rs/issues/348

### DIFF
--- a/gio/src/subclass/action_group.rs
+++ b/gio/src/subclass/action_group.rs
@@ -301,7 +301,7 @@ pub trait ActionGroupImpl: ObjectImpl {
 }
 
 unsafe impl<T: ActionGroupImpl> IsImplementable<T> for ActionGroup {
-    fn interface_init(iface: &mut glib::Class<Self>) {
+    fn interface_init(iface: &mut glib::Interface<Self>) {
         let iface = iface.as_mut();
 
         iface.action_added = Some(action_group_action_added::<T>);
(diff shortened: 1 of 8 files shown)


sdroege:
Maybe we should also distinguish further and have `glib::Class<T>` and `glib::Interface<T>` instead of using `glib::Class<T>` for both? Opinions?

bilelmoussaoui:
> Maybe we should also distinguish further and have `glib::Class<T>` and `glib::Interface<T>` instead of using `glib::Class<T>` for both? Opinions?

Would there be any differences between  `Class<T>` & the future `Interface<T>` other than making them distinct?

sdroege:
> Would there be any differences between `Class<T>` & the future `Interface<T>` other than making them distinct?

That and you can't really use any of the `Class` API on interfaces anyway (see this MR). It seems cleaner so I'm tending towards doing that.

sdroege:
Also would be needed for proper reference counting for a `InterfaceRef` kind of type, so let's go with that

GuillaumeGomez:
> Maybe we should also distinguish further and have `glib::Class<T>` and `glib::Interface<T>` instead of using `glib::Class<T>` for both? Opinions?

II think it'd be better to do it like this.

sdroege:
Done, please review @bilelmoussaoui @GuillaumeGomez :)

GuillaumeGomez (approved):

sdroege:
@bilelmoussaoui Also this needs some change in gtk4 so that stuff continues to compile (see gio commit here)

bilelmoussaoui (approved):

Issue #128, opened 2020-11-10
Add automatic signal connection in combination with gtk::Builder


sdroege:
The idea here would be to annotate the signal handler functions with a procedural macro that would export the signature of the function so it can be checked at runtime.

E.g.

```rust
#[gtk::signal_handler("my_button_clicked")]
fn button_clicked_handler(&self, button: &gtk::Button) { ... }
```

This would then expand to something like

```rust
#[no_mangle]
pub fn gtk_rs_priv_my_button_clicked(&self, args: &[glib::Value]) -> Option<glib::Value> {
    let button = args[0].get::<gtk::Button>().expect("Wrong type").expect("Unexpected None");
    self.button_clicked_handler(&button);
}

fn button_clicked_handler(&self, button: &gtk::Button) { ... }
```

The nullability of the arguments would have to be detected/handled automatically based on the types.

Now for connecting the whole thing we would do a new function on `BuilderExt` that calls `builder.connect_signals()`, assumes/requires that `&self` is cloneable, and then via `dlsym` (etc) checks for each name if `gtk_rs_priv_NAME` does exist, e.g. `gtk_rs_priv_my_button_clicked`.

Instead of only supporting a `&self` first parameter we should also allow arbitrary other types that are cloneable, which shouldn't be too hard in addition either.

piegamesde:
Some prior art: https://idanarye.github.io/woab/woab/derive.BuilderSignal.html

It's a derive macro that for an enum with one variant per signal type.

ids1024:
Looking at `pygobject`, this seems to be done with `gtk_widget_class_set_connect_func` on GTK3 and `GtkBuilderScope` with GTK4.

> Could maybe make use of https://crates.io/crates/label for this

That uses the `ctor` crate. Not sure if that limits platform compatibility in an important way, but the approach seems potentially problematic: https://github.com/jonay2000/label/issues/28

It seems better to instead use a proc macro on the `impl` block, with something like this:

```rust
#[gtk::composite_template]
impl Foo {
    #[signal]
    foo_bar(&self, widget: &Self::Class) {}
}
```

Though if it's on `impl Foo` instead of `impl WidgetImpl for Foo`, that also leaves the issue of how to automatically call the function registering the signals, other than explicitly like in https://github.com/gtk-rs/gtk-rs/pull/269.

jonay2000:
[...]

PR #359, opened 2021-03-16
Signal connection wth Builder

andy128k:
Addresses https://github.com/gtk-rs/gtk-rs/issues/128

Implementation is similar to #266 

This looks not very useful. `FromValue` is not implemented for a `glib::Object` (and `gtk::Widget`), so handler's arguments are limited to primitive types.

### DIFF
--- a/gtk/tests/builder_handlers.rs
+++ b/gtk/tests/builder_handlers.rs
@@ -1,0 +1,145 @@
+#[derive(glib::Downgrade)]
+pub struct MyWidget(gtk::Widget);
+
+#[gtk3_macros::builder_handlers]
+impl MyWidget {
+    fn handler0(&self) {}
+
+    fn handler1(&self, _p: u32) {}
+
+    fn handler2(&self, _x: f64, _y: f64) {}
+
+    fn handler3(&self, x: f64, y: f64) -> f64 {
+        x + y
+    }
+
+    // fn handler4(&self, x: f64, y: f64) -> Option<f64> {
+    //     if x >= 0.0 && y >= 0.0 {
+    //         Some(x + y)
+    //     } else {
+    //         None
+    //     }
+    // }
+}
+
+// Generated code
+/*
+impl MyWidget {
+    #[allow(clippy)]
+    fn get_handler(
+        &self,
+        signal: &str,
+    ) -> Option<Box<dyn Fn(&[glib::Value]) -> Option<glib::Value> + 'static>> {
+        match signal {
+            "handler0" => Some({
+                #[allow(unused_variables)]
+                Box::new(
+                    glib::clone!(@weak self as this => move |values: &[glib::Value]| {
+                        this.handler0();
+                        None
+                    }),
+                )
+            }),
+            "handler1" => Some({
+                #[allow(unused_variables)]
+                Box::new(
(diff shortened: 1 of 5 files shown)


sdroege:
> This looks not very useful. `FromValue` is not implemented for a `glib::Object` (and `gtk::Widget`), so handler's arguments are limited to primitive types.

This should be less of a problem once I finally finished https://github.com/gtk-rs/gtk-rs/pull/31 . Then you can treat all the types the same without the distinction of nullable/non-nullable types like now.

gtk3-macros/src/builder_handlers.rs
```diff
@@ -0,0 +1,200 @@
...
+        .iter()
+        .skip(1)
+        .enumerate()
+        .map(|(index, arg)| {
+            let arg_type = argument_type(arg)?;
+            Ok(quote_spanned! { arg.span() =>
+                match values[#index].get_some() {
+                    Ok(value) => value,
+                    Err(error) => {
+                        glib::g_critical!("builder handler", "Handler {} expects an argument of type {} but received `{:?}`: {}.", #handler_name, stringify!(#arg_type), values[#index], error);
```
sdroege, line 42: This should be a panic in one way or another. Returning `None` can easily cause memory safety issues in the C code calling this signal handler.

gtk/tests/builder_handlers.rs
```diff
@@ -0,0 +1,145 @@
+#[derive(glib::Downgrade)]
+pub struct MyWidget(gtk::Widget);
+
+#[gtk3_macros::builder_handlers]
+impl MyWidget {
+    fn handler0(&self) {}
```
sdroege, line 6: It would seem also useful to provide a `#[signal(name = "name_of_the_signal_handler_in_the_builder_file")]` kind of attribute so that the names don't necessarily have to match exactly.

Not sure what other attributes could be useful, should check what other languages like Vala and Python are doing here.

gtk3-macros/src/builder_handlers.rs
```diff
@@ -0,0 +1,200 @@
...
+        .collect::<Result<_, Error>>()?;
+
+    let signal = &info.name;
+    let method = &info.sig.ident;
+    let is_unit = matches!(info.sig.output, syn::ReturnType::Default);
+    let handler = if is_unit {
+        quote_spanned! { info.sig.span() =>
+            #signal => Some({
+                #[allow(unused_variables)]
+                Box::new(glib::clone!(@weak self as this => move |values: &[glib::Value]| {
```
sdroege, line 57: I think instead of having to work with a `struct Foo(Widget)` kind of type, it would make more sense to integrate this with subclassing and especially composite templates.

That way you could have the macro on the impl struct, and use the `ObjectSubclass::from_instance()` / `::get_instance()` functions to get to a widget and back.

sdroege (commented):
Looks generally like going in the right direction, thanks a lot :)

gtk/tests/builder_handlers.rs
andy128k: It is already there. I just was too lazy to create an example.

sdroege: Even better :)

Issue #358, opened 2021-03-15
`drag-motion` signal handler return value should be `bool`, not `Inhibit`


SolraBizna:
`WidgetExt::connect_drag_motion` accepts a function returning `Inhibit`, because the `drag-motion` signal handler returns a boolean. It was probably assumed that this signal's boolean returned value is used to inhibit signal propagation, like most other boolean-returning signals. However, this is not the case.

From the documentation for the D binding (which puts it more clearly than any other documentation I was able to find):

> The `::drag-motion` signal is emitted on the drop site when the user moves the cursor over the widget during a drag. The signal handler must determine whether the cursor position is in a drop zone or not. If it is not in a drop zone, it returns `FALSE` and no further processing is necessary. Otherwise, the handler returns `TRUE`. In this case, the handler is responsible for providing the necessary information for displaying feedback to the user, by calling `gdk_drag_status()`.

I can confirm that, with the current state of `gtk-rs`, returning `Inhibit(true)` results in drag acceptance, and `Inhibit(false)` results in drag rejection.

SolraBizna:
The same thing also applies to the `drag-drop` signal. `true` → "the drop is succeeding", `false` → "the drop is failing".

sdroege:
The same probably also applies to GTK4 in one way or another (CC @bilelmoussaoui).

@SolraBizna, do you want to submit a PR for changing this to a `bool`?

bilelmoussaoui:
> The same probably also applies to GTK4 in one way or another (CC @bilelmoussaoui).

There are probably some functions where it's needed for GTK 4, but not those as they were removed/replaced

SolraBizna:
As @sdroege suggested, I've made a PR that would resolve this issue. Hopefully it's adequate. I wasn't able to test it, but the changes are pretty trivial and everything does build.

PR #361, opened 2021-03-16
Make `drag-motion` and `drag-drop` signals return `bool` instead of `Inhibit`

SolraBizna:
As described in #358, unlike most `bool`-returning signals on `Widget`, the `drag-motion` and `drag-drop` signals' return values are *not* interpreted as inhibiting further signal propagation. However, the person who wrote their Gir assumed that they were. (Understandably, as the official documentation for these signals is pretty sparse.) This PR modifies them to return `bool` instead of `Inhibit`.

- `drag-motion`: `true` → a drop here will succeed, `false` → a drop here will fail; not `Inhibit`-like, modified
- `drag-drop`: `true` → a drop here succeeded, `false` → a drop here failed; not `Inhibit`-like, modified
- `drag-failed`: `true` → carry out the standard "drag failed" animation, `false` → don't carry out that animation; I left this as `Inhibit` since it seems to fit

I wasn't able to test this PR myself, since my application has a lot of 0.9isms and doesn't compile with 0.13, but the changes do build and (in Rust, anyway) that's strong evidence that they're not *fundamentally* broken...

### DIFF
--- a/gtk/Gir.toml
+++ b/gtk/Gir.toml
@@ -2706,13 +2706,7 @@ manual_traits = ["WidgetExtManual"]
         name = "data"
         transformation = "borrow"
     [[object.signal]]
-    name = "drag-drop"
-    inhibit = true
-    [[object.signal]]
     name = "drag-failed"
-    inhibit = true
-    [[object.signal]]
-    name = "drag-motion"
     inhibit = true
     [[object.signal]]
     name = "draw"
--- a/gtk/src/auto/widget.rs
+++ b/gtk/src/auto/widget.rs
@@ -909,9 +909,7 @@ pub trait WidgetExt: 'static {
         f: F,
     ) -> SignalHandlerId;
 
-    fn connect_drag_drop<
-        F: Fn(&Self, &gdk::DragContext, i32, i32, u32) -> glib::signal::Inhibit + 'static,
-    >(
+    fn connect_drag_drop<F: Fn(&Self, &gdk::DragContext, i32, i32, u32) -> bool + 'static>(
         &self,
         f: F,
     ) -> SignalHandlerId;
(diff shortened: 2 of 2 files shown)


gtk/Gir.toml
```diff
@@ -2706,14 +2706,12 @@ manual_traits = ["WidgetExtManual"]
...
-    inhibit = true
+    name = "drag-drop" # this bool return should not be an Inhibit
     [[object.signal]]
-    name = "drag-failed"
+    name = "drag-failed" # this one should
     inhibit = true
     [[object.signal]]
-    name = "drag-motion"
-    inhibit = true
+    name = "drag-motion" # this one should not
```
bilelmoussaoui, line 2714: You can just drop this

gtk/Gir.toml
```diff
@@ -2706,14 +2706,12 @@ manual_traits = ["WidgetExtManual"]
         name = "data"
         transformation = "borrow"
     [[object.signal]]
-    name = "drag-drop"
-    inhibit = true
+    name = "drag-drop" # this bool return should not be an Inhibit
```
bilelmoussaoui, line 2709: and this

gtk/Gir.toml
SolraBizna: There is an existing comment above:
```rust
# return Inhibit instead of bool
```
which is implied to cover all the subsequent signals. I put those comments in to make it clear that those are individual, deliberate exceptions.

(I would have reordered them, which would have allowed them to be clearly delineated with a single comment, but that would have made a much messier diff, and I would have worried more about breaking something.)

SolraBizna: oh, you were saying that those two entire `[[object.signal]]` sections can be dropped. (Sorry, I am sleep deprived and slow on the uptake today.)

bilelmoussaoui: No worries, thanks!

bilelmoussaoui:
Do you mind squashing the two commits into one? Otherwise looks good to me, thanks!

SolraBizna:
Done. I think.

bilelmoussaoui (approved):

sdroege (approved):

PR #364, opened 2021-03-18
Update to GLib 2.68.0

sdroege:


gio/src/auto/dbus_proxy.rs
```diff
@@ -251,6 +251,13 @@ impl DBusProxy {
     }
 }
 
+impl fmt::Display for DBusProxy {
+    #[inline]
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
```
GuillaumeGomez, line 256: Wow that's cool!

gio/src/auto/dbus_proxy.rs
```diff
@@ -935,9 +942,3 @@ impl<O: IsA<DBusProxy>> DBusProxyExt for O {
         }
     }
 }
-
-impl fmt::Display for DBusProxy {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
```
GuillaumeGomez, line 940: Ah that was already here. :laughing:

sdroege: Was just moved around for arbitrary reasons, yes

GuillaumeGomez (approved):

GuillaumeGomez:
`Gio` seems to have issues.

sdroege:
> `Gio` seems to have issues.

Already solved :)

bilelmoussaoui (approved):

PR #365, opened 2021-03-18
Update gir (Gir.toml format update for gir directories)

GuillaumeGomez:
From https://github.com/gtk-rs/gir/pull/1041.

I still wonder if we should keep the default value for `--gir-files-directory` in the `generator.py`...


sdroege (approved):

sdroege:
> I still wonder if we should keep the default value for `--gir-files-directory` in the `generator.py`...

I don't think so. It should only be provided if overriding the `Gir.toml` files is necessary.

GuillaumeGomez:
Ok, sending another PR to change that then.

PR #367, opened 2021-03-19
Add a GError derive macro

elmarco:


glib-macros/src/gerror_derive.rs
```diff
@@ -0,0 +1,68 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use proc_macro2::TokenStream;
+use proc_macro_error::abort_call_site;
+use quote::{quote, quote_spanned};
+use syn::{punctuated::Punctuated, spanned::Spanned, token::Comma, Data, Ident, Variant};
+
+use crate::utils::{crate_ident_new, parse_domain_name};
+
+// FIXME: merge with genum version
```
sdroege, line 10: Is that difficult? If not, maybe just do it here in a separate commit :)

sdroege (commented):
Looks mostly good to me, thanks!

glib-macros/src/lib.rs
```diff
@@ -249,6 +250,31 @@ pub fn genum_derive(input: TokenStream) -> TokenStream {
...
+///
+/// # Example
+///
+/// ```
+/// use glib::prelude::*;
+/// use glib::subclass::prelude::*;
+///
+/// #[derive(Debug, Copy, Clone, glib::GError)]
+/// #[gerror(domain_name = "ExFoo")]
+/// enum Foo {
```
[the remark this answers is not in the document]
elmarco: why do you think it's necessary? It's the rust implementation after, doesn't have an influence on the ffi, right? Afaict, "self as i32" in code() is the only case where it would matter, but it should be fine whatever representation, isn't it?

glib-macros/src/lib.rs
```diff
@@ -249,6 +250,31 @@ pub fn genum_derive(input: TokenStream) -> TokenStream {
...
+///     Blah,
+///     Baaz,
+/// }
+/// ```
+///
+/// [`ErrorDomain`]: error/trait.ErrorDomain.html
+#[proc_macro_derive(GError, attributes(gerror))]
```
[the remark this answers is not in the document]
elmarco: agree

glib-macros/src/gerror_derive.rs
```diff
@@ -0,0 +1,68 @@
...
+            fn from(code: i32) -> Option<Self>
+            where
+                Self: Sized
+            {
+                #from_glib
+                None
+            }
+        }
+    }
+    // TODO: generate the extern "C" fn {}_quark(void) name.to_string().to_snake_case() ?
```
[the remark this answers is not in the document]
elmarco: Well, unfortunately, cbindgen doesn't generate the ffi, the crate has to provide it. Or am I missing something?

glib-macros/src/gerror_derive.rs
```diff
@@ -0,0 +1,68 @@
...
+            "{}: derive(GError) requires #[gerror(domain_name = \"DomainName\")]",
+            e
+        ),
+    };
+    let from_glib = gen_enum_from_glib(name, enum_variants);
+
+    quote! {
+        impl #crate_ident::error::ErrorDomain for #name {
+            fn domain() -> #crate_ident::Quark {
+                #crate_ident::Quark::from_string(#domain_name)
```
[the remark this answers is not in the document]
elmarco: ok

glib-macros/src/gerror_derive.rs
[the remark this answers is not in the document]
elmarco: However, one somewhat big issue if/when we start generating extern fn or other FFI from macros is that cbindgen doesn't see them... By default, it process unexpanded crates. If you expand crates and glib macros, then everything becomes also more complicated for gbindgen (as it rely on the macros...! the expanded version would be an extra layer of complexity)

What would be interesting would be to be able expand macros selectively. But at this point, this not possible (it's a rustc today limitation). Another option would be for gbindgen to process unexpanded & expanded version...?

glib-macros/src/gerror_derive.rs
elmarco: done, hopefully in a correct way

glib-macros/src/lib.rs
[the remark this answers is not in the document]
sdroege: Right, if you always go via `FromGlib` / `ToGlib` it's irrelevant indeed

glib-macros/src/gerror_derive.rs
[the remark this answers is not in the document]
sdroege: For now I would do like we always did: write the FFI functions manually. At some point we could consider having a code generator for that, I guess. Unlike macros that would then also not run into that problem.

But I don't think it should ever be the job of this derive macro to generate the FFI functions.

glib-macros/src/gerror_domain_derive.rs
```diff
@@ -0,0 +1,48 @@
...
+            "{}: derive(GErrorDomain) requires #[gerror_domain(name = \"DomainName\")]",
+            e
+        ),
+    };
+    let from_glib = gen_enum_from_glib(name, enum_variants);
+
+    quote! {
+        impl #crate_ident::error::ErrorDomain for #name {
+            fn domain() -> #crate_ident::Quark {
+                #crate_ident::Quark::from_static_string(&concat!(#domain_name, "\0"))
```
sdroege, line 32: There is no such function in the safe bindings (because of Rust not having a NUL-terminator in strings), you'll have to call FFI.

glib/src/quark.rs
```diff
@@ -14,6 +14,11 @@ impl Quark {
         unsafe { from_glib(ffi::g_quark_from_string(s.to_glib_none().0)) }
     }
 
+    #[doc(alias = "g_quark_from_static_string")]
+    pub fn from_static_string(s: &'static str) -> Quark {
+        unsafe { from_glib(ffi::g_quark_from_static_string(s.to_glib_none().0)) }
```
sdroege, line 19: This is unsound: the string is not actually static. `to_glib_none().0` has to allocate a new temporary string, append a NUL, pass it to the C function and then frees it.

That's why you directly have to call the FFI function

glib-macros/src/gerror_domain_derive.rs
```diff
@@ -0,0 +1,48 @@
...
+
+            fn from(value: i32) -> Option<Self>
+            where
+                Self: Sized
+            {
+                #from_glib
+            }
+        }
+    }
+    // TODO: generate the extern "C" fn {}_quark(void) name.to_string().to_snake_case() ?
```
sdroege, line 47: (so remove this TODO comment, that's just confusing for now)

sdroege (commented):
Can you also add a small test to glib-macros/tests, like the existing ones? Then we can at least be sure that it compiles and kind of behaves (e.g. you can create a `glib::Error` from it).

glib-macros/src/gerror_derive.rs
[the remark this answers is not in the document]
elmarco: You are probably right.

glib/src/quark.rs
elmarco: ok

glib-macros/src/gerror_derive.rs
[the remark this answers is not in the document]
elmarco: sorry, my change got lost.. doing it again

glib-macros/src/gerror_domain_derive.rs
elmarco: ah, that's why we use the once_cell then :)

sdroege: No, the `Lazy` we use so we don't have to go through the global `GQuark` lock every time :)

elmarco:
> Can you also add a small test to glib-macros/tests, like the existing ones? Then we can at least be sure that it compiles and kind of behaves (e.g. you can create a `glib::Error` from it).

done (it was partially covered by doc already)

glib-macros/src/genum_derive.rs
```diff
@@ -112,10 +95,23 @@ pub fn impl_genum(input: &syn::DeriveInput) -> TokenStream {
             }
         }
 
+        impl #crate_ident::translate::TryFromGlib<i32> for #name {
```
sdroege, line 98: Why is this needed btw?

sdroege (approved):

glib-macros/src/genum_derive.rs
elmarco: ErrorDomain::from() -> Option, so it was simpler for the common code generator to also return Option. I could have not implemented TryFromGlib and just inlined the call and unwrap() in FromGlib, but I thought that could also be useful (more useful that FromGlib that may panic)

GuillaumeGomez (approved):

PR #371, opened 2021-03-21
Fix async clone closures handling

### LINKED ISSUE #369: clone! macro does not work with async blocks
The content of the async block is not retained by the clone! macro.

Minimal example:
This
```rust
button.connect_clicked(move |button| {
        let main_context = MainContext::default();
        // The main loop executes the asynchronous block
        main_context.spawn_local(glib::clone!(@strong button => async move {
            // Deactivate the button until the operation is done
            button.set_sensitive(false);
            timeout_future_seconds(10).await;
            // Activate the button again
            button.set_sensitive(true);
        }));
    });
```
expands to this
```rust
  button.connect_clicked(move |button| {
        let main_context = MainContext::default();
        main_context.spawn_local({
            let button = button.clone();
            async move { {} }
        });
    });
```

GuillaumeGomez:
Fixes #369.

cc @Hofer-Julian

### DIFF
--- a/glib/tests/clone.rs
+++ b/glib/tests/clone.rs
@@ -4,6 +4,7 @@ use std::panic;
 use std::rc::Rc;
 use std::sync::{Arc, Mutex};
 
+use futures_executor::block_on;
 use glib::{clone, Downgrade, Object};
 
 struct State {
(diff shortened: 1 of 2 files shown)


glib/tests/clone.rs
```diff
@@ -490,12 +491,15 @@ fn test_clone_macro_body() {
...
 
     // This one is still a rust unstable feature.
     // let _closure = clone!(@weak v => async move || 0);
-    let _closure = clone!(@weak v => move || async move { 0 });
-    let _closure = clone!(@weak v => async move { 0 });
+    let closure = clone!(@weak v => move || async move { *v.borrow_mut() += 1; });
+    block_on(closure());
+    assert_eq!(*v.borrow(), 2);
+    block_on(clone!(@weak v => async move { *v.borrow_mut() += 1; }));
+    assert_eq!(*v.borrow(), 3);
```
sdroege, line 502: For closures we also don't check if they're actually callable and do the right thing. Can you add that in a separate commit so we can be sure?

sdroege (approved):

glib/tests/clone.rs
GuillaumeGomez: What do you mean?

sdroege: We have e.g. `let _closure = clone!(@weak v => @default-return { let x = 12; x + 2 }, move || 19);` above. This doesn't test anything other than that the macro doesn't produce broken code. It could also produce an empty closure and we wouldn't know from the test.

So instead, call the closure and check that it actually contains the code it should :)

GuillaumeGomez: I see, adding it.

GuillaumeGomez: I found bugs so let's merge this and I'll open another PR.

PR #372, opened 2021-03-21
glib: implement DerefMut for ClassRef/InterfaceRef

bilelmoussaoui:
Needed to be able to write such code
```rust
let cell_class: glib::object::ClassRef<CellArea> =
    glib::Class::from_type(O::static_type()).unwrap();
let pspec: Option<glib::ParamSpec> =
    from_glib_none(ffi::gtk_cell_area_class_find_cell_property(
        cell_class.as_mut() as *mut _ as *mut ffi::GtkCellAreaClass,
        property_name.to_glib_none().0,
    ));
```

### DIFF
--- a/glib/src/object.rs
+++ b/glib/src/object.rs
@@ -2764,6 +2764,12 @@ impl<'a, T: IsClass> ops::Deref for ClassRef<'a, T> {
     }
 }
 
+impl<'a, T: IsClass> ops::DerefMut for ClassRef<'a, T> {
+    fn deref_mut(&mut self) -> &mut Class<T> {
+        unsafe { self.0.as_mut() }
+    }
+}
+
 impl<'a, T: IsClass> Drop for ClassRef<'a, T> {
     fn drop(&mut self) {
         if self.1 {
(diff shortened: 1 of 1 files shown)


sdroege:
> ```rust
>         cell_class.as_mut() as *mut _ as *mut ffi::GtkCellAreaClass,
> ```

Should be

```rust
cell_class.as_ref() as *const _ as *mut _ as *mut ffi::GtkCellAreaClass,
```

Or similar. The C function is not actually changing the class, is it? It just is not a `const` pointer in C because C.

bilelmoussaoui:
I'm wondering  why `glib::Class::from_type` doesn't have any reference to `T`. Which requires explicitly writing `glib::object::ClassRef<T>`. Any reason for not making it `glib::Class::from_type::<T>`?

sdroege:
> I'm wondering why `glib::Class::from_type` doesn't have any reference to `T`. Which requires explicitly writing `glib::object::ClassRef<T>`. Any reason for not making it `glib::Class::from_type::<T>`?

How about `glib::Class::<T>::from_type(typ)`?

bilelmoussaoui:
> > I'm wondering why `glib::Class::from_type` doesn't have any reference to `T`. Which requires explicitly writing `glib::object::ClassRef<T>`. Any reason for not making it `glib::Class::from_type::<T>`?
> 
> How about `glib::Class::<T>::from_type(typ)`?

That actually works. I somehow never thought you would write it that way. Also R-A seems to be unable to guess the type in such case, but that's not a problem as the code compiles :-) 

I suppose, this can be closed as it's not really useful, at least, I solved the use case I thought I had.

sdroege:
> I suppose, this can be closed as it's not really useful, at least, I solved the use case I thought I had.

Well, not just "not useful" but it's wrong :) With your change you can get mutable access to the class struct after type initialization. That must not happen.

PR #373, opened 2021-03-21
Fix some clone @default-return handlings

GuillaumeGomez:


glib/tests/clone.rs
```diff
@@ -457,23 +457,33 @@ fn test_clone_macro_default_return() {
...
-    #[allow(dead_code)]
+    #[derive(PartialEq, Debug)]
     struct Bar {
         x: i32,
     }
 
-    let _closure = clone!(@weak v => @default-return Bar { x: 0 }, move || Bar { x: 1 });
+    let closure = clone!(@weak v => @default-return Bar { x: 0 }, move || Bar { x: 1 });
+    let wrapper = clone!(@weak v => move |call: &(dyn Fn() -> Bar)| call());
+    assert_eq!(wrapper(&closure), Bar { x: 0 });
```
GuillaumeGomez, line 467: This test is failing because it seems like it can upgrade `v` twice, which is very weird...

glib/tests/clone.rs
```diff
@@ -457,23 +457,33 @@ fn test_clone_macro_default_return() {
...
         C { x: i32 },
     }
-    let _closure = clone!(@weak v => @default-return Enum::A, move || Enum::A);
-    let _closure = clone!(@weak v => @default-return Enum::B(0), move || Enum::A);
-    let _closure = clone!(@weak v => @default-return Enum::C { x: 0 }, move || Enum::A);
-    let _closure = clone!(@weak v => @default-return { let x = 12; x + 2 }, move || 19);
+    let closure1 = clone!(@weak v => @default-return Enum::A, move || Enum::B(0));
+    let closure2 = clone!(@weak v => @default-return Enum::B(0), move || Enum::A);
+    let closure3 = clone!(@weak v => @default-return Enum::C { x: 0 }, move || Enum::A);
+    let closure4 = clone!(@weak v => @default-return { let x = 12; x + 2 }, move || 19);
```
GuillaumeGomez, line 478: This test is failing too because it returns 19 instead of 14, no clue why either...

glib/tests/clone.rs
```diff
@@ -457,23 +457,33 @@ fn test_clone_macro_default_return() {
...
+
+    let wrapper = clone!(@weak v => move |call: &(dyn Fn() -> Enum)| call());
+    assert_eq!(wrapper(&closure1), Enum::B(0));
+    assert_eq!(wrapper(&closure2), Enum::A);
+    assert_eq!(wrapper(&closure3), Enum::A);
+
+    let wrapper = clone!(@weak v => move |call: &(dyn Fn() -> usize)| call());
+    assert_eq!(wrapper(&closure4), 14);
```
GuillaumeGomez, line 486: This test is failing too because it returns 19 instead of 14, no clue why either...

glib/tests/clone.rs
sdroege: Can you provide the `cargo expand` output for the failing tests?

GuillaumeGomez: Enjoy! So this is `closure`:

```rust
{
    let v = glib::clone::Downgrade::downgrade(&v);
    move || {
        let v = match glib::clone::Upgrade::upgrade(&v) {
            Some(val) => val,
            None => {
                glib::g_debug!(glib::CLONE_MACRO_LOG_DOMAIN, "Failed to upgrade v",);
                let ___return_value = || Bar { x: 0 };
                return ___return_value();
            }
        };
        {
            Bar { x: 1 }
        }
    }
}
```

And this is `wrapper`:

```
{
    let v = glib::clone::Downgrade::downgrade(&v);
    move |call: &(dyn Fn() -> Bar)| {
        let v = match glib::clone::Upgrade::upgrade(&v) {
            Some(val) => val,
            None => {
                panic!("failed to upgrade `v` (if you don't want to panic, use @default-return)",)
            }
        };
        {
            call()
        }
    }
}
```

sdroege: What's the problem here? I would expect it to return `Bar { x: 1 }` here and not `Bar { x: 0 }`. Why do you expect the other?

glib-macros/src/clone.rs
```diff
@@ -381,7 +381,12 @@ fn get_expr(parts: &mut Peekable<ProcIter>) -> String {
                 }
                 ret.push_str(&p_s);
             }
-            Some(TokenTree::Group(g)) => ret.push_str(&group_to_string(g)),
+            Some(TokenTree::Group(g)) => {
+                ret.push_str(&group_to_string(g));
+                if total == 0 {
```
GuillaumeGomez, line 386: I need to revert this change.

glib/tests/clone.rs
GuillaumeGomez: The problem was me. :)

GuillaumeGomez:
Fixed it.

glib/tests/clone.rs
```diff
@@ -451,29 +451,46 @@ fn test_clone_macro_typed_args() {
```
sdroege, line 451: Maybe the tests just above can also return something from their closures that can be tested?

sdroege (approved):

glib/tests/clone.rs
GuillaumeGomez: Another PR incoming then!

PR #375, opened 2021-03-22
Add layout.rs, and manual LayoutLine::x_to_index impl

### LINKED ISSUE #374: Pango LayoutLine::x_to_index doesn't report offset when x is outside of line
The C signature or this function is,

```c
gboolean
pango_layout_line_x_to_index (PangoLayoutLine *line,
                              int x_pos,
                              int *index_,
                              int *trailing);
```

Importantly, it 'returns' three values; `index`, `trailing`, and a `bool` indicating whether or not `x_pos` is within the horizontal bounds of the line.

The signature in the `pango` crate is,

```rust
pub fn x_to_index(&self, x_pos: i32) -> Option<(i32, i32)>;
```

In this case, the `Some` case represents an `x_pos` that is inside the line's horizontal bounds, and `None` represents outside.

My understanding is is that even in the case that the hit is outside, we may still want to know what the resulting offset is, and with the current signature, that information is discarded.

I can think of two reasonable alternatives:

```rust
pub fn x_to_index(&self, x_pos: i32) -> (bool, i32, i32);
//or
pub fn x_to_index(&self, x_pos: i32) -> Result<(i32, i32), (i32, i32)>;
```

I think realistically the first option makes the most sense. I would probably also return some `Position` type with named fields instead of a tuple, but that's a personal design preference.

If you agree that this is an issue I would be happy to prepare a patch.

cmyr:
Not sure if this is exactly right, let me know if I've missed anything.

- closes #374

### DIFF
--- a/pango/Gir.toml
+++ b/pango/Gir.toml
@@ -50,7 +50,6 @@ generate = [
     "Pango.GravityHint",
     "Pango.Item",
     "Pango.LayoutIter",
-    "Pango.LayoutLine",
     "Pango.LayoutRun",
     "Pango.Matrix",
     "Pango.Overline",
@@ -240,6 +239,13 @@ status = "generate"
         [[object.function.parameter]]
         name = "tabs"
         const = true
+
+[[object]]
+name = "Pango.LayoutLine"
+status = "generate"
+    [[object.function]]
+    name = "x_to_index"
+    manual = true
 
 [[object]]
 name = "Pango.ShapeFlags"
--- a/pango/src/layout.rs
+++ b/pango/src/layout.rs
@@ -1,0 +1,57 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use glib::translate::*;
+
+use crate::LayoutLine;
+
+// rustdoc-stripper-ignore-next
+/// The result of [`LayoutLine::x_to_index`].
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+pub struct HitPosition {
+    // rustdoc-stripper-ignore-next
+    /// The UTF-8 byte offset of the grapheme closest to the position.
+    ///
+    /// This position is relative to the start of the line, not the start
+    /// of the [`Layout`]'s text.
+    ///
+    /// [`Layout`]: crate::Layout
+    pub index: i32,
+    // rustdoc-stripper-ignore-next
+    /// The codepoint within the grapheme of the position.
+    ///
+    /// This will always be either `0`, or the number of `char`s (*not bytes!*)
+    /// in the grapheme. This represents whether the user clicked near the start
+    /// of the grapheme or near the end; this is important for things like
+    /// resolving cursor positions.
+    pub trailing: i32,
+    // rustdoc-stripper-ignore-next
+    /// Whether or not the position was within the bounds of the line.
+    ///
+    /// If this is `false`, then `index` and `trailing` will always resolve
+    /// to either the very first or the very last position in the line; this
+    /// behaviour is dependent on the line's resolved writing direction.
+    pub is_inside: bool,
+}
+
+impl LayoutLine {
+    #[doc(alias = "pango_layout_line_x_to_index")]
+    pub fn x_to_index(&self, x_pos: i32) -> HitPosition {
+        let mut index = 0;
+        let mut trailing = 0;
+
+        let is_inside = unsafe {
+            from_glib(ffi::pango_layout_line_x_to_index(
+                self.to_glib_none().0,
+                x_pos,
+                &mut index,
+                &mut trailing,
+            ))
+        };
+
+        HitPosition {
+            index,
+            trailing,
+            is_inside,
+        }
+    }
+}
--- a/pango/src/lib.rs
+++ b/pango/src/lib.rs
@@ -52,6 +52,8 @@ mod functions;
 pub mod item;
 pub mod language;
 pub use crate::language::Language;
+pub mod layout;
+pub use crate::layout::HitPosition;
 pub mod rectangle;
 pub use crate::rectangle::Rectangle;
 pub mod glyph;
--- a/pango/src/auto/layout_line.rs
+++ b/pango/src/auto/layout_line.rs
@@ -89,25 +89,4 @@ impl LayoutLine {
             x_pos
         }
     }
-
-    #[doc(alias = "pango_layout_line_x_to_index")]
-    pub fn x_to_index(&self, x_pos: i32) -> Option<(i32, i32)> {
-        unsafe {
-            let mut index_ = mem::MaybeUninit::uninit();
-            let mut trailing = mem::MaybeUninit::uninit();
-            let ret = from_glib(ffi::pango_layout_line_x_to_index(
-                self.to_glib_none().0,
-                x_pos,
-                index_.as_mut_ptr(),
-                trailing.as_mut_ptr(),
-            ));
-            let index_ = index_.assume_init();
-            let trailing = trailing.assume_init();
-            if ret {
-                Some((index_, trailing))
-            } else {
-                None
-            }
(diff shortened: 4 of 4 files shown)


pango/src/layout.rs
```diff
@@ -0,0 +1,51 @@
...
+    trailing: i32,
+    /// Whether or not the position was within the bounds of the line.
+    ///
+    /// If this is `false`, then `index` and `trailing` will always resolve
+    /// to either the very first or the very last position in the line; this
+    /// behaviour is dependent on the line's resolved writing direction.
+    is_inside: bool,
+}
+
+
```
GuillaumeGomez, line 29: Please remove this extra empty line.

pango/src/layout.rs
```diff
@@ -0,0 +1,51 @@
...
+impl LayoutLine {
+    /// Return the [`HitPosition`] corresponding to the provided `x_pos` in this line.
+    pub fn x_to_index(&self, x_pos: i32) -> HitPosition {
+        unsafe {
+            let mut index = mem::MaybeUninit::uninit();
```
GuillaumeGomez, line 34: There isn't much interest into using `MaybeUninit` here.

pango/src/layout.rs
```diff
@@ -0,0 +1,50 @@
...
+    ///
+    /// If this is `false`, then `index` and `trailing` will always resolve
+    /// to either the very first or the very last position in the line; this
+    /// behaviour is dependent on the line's resolved writing direction.
+    is_inside: bool,
+}
+
+impl LayoutLine {
+    /// Return the [`HitPosition`] corresponding to the provided `x_pos` in this line.
+    pub fn x_to_index(&self, x_pos: i32) -> HitPosition {
```
GuillaumeGomez, line 31: You forgot to add `#[doc(alias = "pango_layout_line_x_to_index")]`.

pango/src/layout.rs
cmyr: I just copy-pasted this from the existing impl. Would you prefer to just initialize to `0` and pass that?


This is a good reminder,  I wanted to document that the returned positions are relative to the first position in the line, not the larger layout.

GuillaumeGomez: In the original code, we assumed that if it returned `false`, something went wrong and we just returned `None`. The situation changed since then. ;)

cmyr: Oh, I see what you're saying. 👍

pango/src/layout.rs
cmyr: does that mean I should remove my own docs?

pango/src/layout.rs
```diff
@@ -0,0 +1,52 @@
...
+/// The result of [`LayoutLine::x_to_index`].
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+pub struct HitPosition {
+    /// The UTF-8 byte offset of the grapheme closest to the position.
+    ///
+    /// This position is relative to the start of the line, not the start
+    /// of the [`Layout`]'s text.
+    ///
+    /// [`Layout`]: crate::Layout
+    index: i32,
```
GuillaumeGomez, line 16: Just realized that you need to make those fields public, otherwise the type will be useless.

pango/src/lib.rs
```diff
@@ -52,6 +52,7 @@ mod functions;
 pub mod item;
 pub mod language;
 pub use crate::language::Language;
+pub mod layout;
```
GuillaumeGomez, line 55: You also need to reexport `HitPosition` (like we do just above for `Language`).

pango/src/layout.rs
GuillaumeGomez: https://doc.rust-lang.org/rustdoc/advanced-features.html?highlight=alias#add-aliases-for-an-item-in-documentation-search

But I just realized that your doc will be removed in any case because we generate documentation too. (Documentation is a dark place for gtk-rs...)

pango/src/layout.rs
```diff
@@ -0,0 +1,52 @@
...
+    /// Whether or not the position was within the bounds of the line.
+    ///
+    /// If this is `false`, then `index` and `trailing` will always resolve
+    /// to either the very first or the very last position in the line; this
+    /// behaviour is dependent on the line's resolved writing direction.
+    pub is_inside: bool,
+}
+
+impl LayoutLine {
+    /// Return the [`HitPosition`] corresponding to the provided `x_pos` in this line.
```
GuillaumeGomez, line 33: So like said below, please remove this doc comment. Documentation will be generated automatically like for the rest of the codebase.

GuillaumeGomez (approved):
Thanks!

cmyr:
@GuillaumeGomez thanks for the speedy review!

I have a follow-up PR at #376; I'm not sure i this makes sense or not, but it was easier to write the patch than to write up an issue. The relevant commit is cba86dac4e.

pango/src/layout.rs
```diff
@@ -0,0 +1,53 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use glib::translate::*;
+
+use crate::LayoutLine;
+
+/// The result of [`LayoutLine::x_to_index`].
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+pub struct HitPosition {
```
GuillaumeGomez, line 9: No need to re-declare this type, just re-use the other one.

GuillaumeGomez: Well, once it's merged of course.

GuillaumeGomez: Please just ignore this. I thought I was reviewing another PR of yours. Big brain time... :brain:

cmyr: sorry this is confusing, that PR includes the commit in #375, because it needs to also add custom methods in `layout.rs`.

pango/Gir.toml
```diff
@@ -241,6 +240,13 @@ status = "generate"
         name = "tabs"
         const = true
 
+[[object]]
+name = "Pango.LayoutLine"
+status = "generate"
+    [[object.function]]
+    name = "x_to_index"
+    ignore = true
```
sdroege, line 248:
```suggestion
    manual = true
```

pango/src/layout.rs
```diff
@@ -0,0 +1,53 @@
...
+    /// to either the very first or the very last position in the line; this
+    /// behaviour is dependent on the line's resolved writing direction.
+    pub is_inside: bool,
+}
+
+impl LayoutLine {
+    #[doc(alias = "pango_layout_line_x_to_index")]
+    pub fn x_to_index(&self, x_pos: i32) -> HitPosition {
+        let mut index = 0;
+        let mut trailing = 0;
```
sdroege, line 36: Could also use `std::mem::MaybeUninit` here but doesn't really matter

sdroege (commented):
Looks good otherwise, thanks!

pango/src/layout.rs
cmyr: Unless there's some error condition I need to handle where `MaybeUninit` would be more correct, I think I prefer it this way? I don't feel like I have a very strong understanding of the tradeoffs, however... 🤷

cmyr:
There's a funny CI failure I don't understand, any pointers?

cmyr:
Am I maybe just... not allowed to have docs?

GuillaumeGomez:
You can, you just need to tell rustdoc-stripper to not remove it by adding (before your doc comment):

```rust
// rustdoc-stripper-ignore-next
```

pango/src/layout.rs
sdroege: Doesn't really matter :)

cmyr:
@GuillaumeGomez okay I think this should be good to go? thanks for your help!

sdroege (approved):

PR #377, opened 2021-03-23
Fix panic when there is no @default-panic and add more checks for clone macro

GuillaumeGomez:


glib/tests/clone.rs
```diff
@@ -416,37 +416,64 @@ fn test_clone_macro_double_rename() {
 
 #[test]
 fn test_clone_macro_typed_args() {
```
sdroege, line 418: Same in the function above

glib/tests/clone.rs
```diff
@@ -416,37 +416,64 @@ fn test_clone_macro_double_rename() {
...
-    let closure = clone!(@weak v, @weak w as x => move |_: i8, _| {
-        println!("v: {}, w: {}", v, x);
+    let closure = clone!(@weak v, @weak w as x => move |arg: i8, arg2| {
+        *v.borrow_mut() = arg;
+        *x.borrow_mut() = arg2;
     });
-    closure(0, 'a');
+    closure(0, 9);
+    assert_eq!(0, *v.borrow());
+    assert_eq!(9, *w.borrow());
```
sdroege, line 476: Do we have tests somewhere that check dropping of the weak references so that upgrading fails?

GuillaumeGomez: Like [here](https://github.com/gtk-rs/gtk-rs/pull/373/files#diff-78ad6fe523da0fb4f9437f0d29ff5d476075b5ff590762b72cdbf13f229fa92bR460) you mean? I can add some to ensure that nothing was done on the values (that's actually a good idea).

sdroege: Yes, sounds good

sdroege:
There are still many tests here that don't check if the closures are actually doing what they're supposed to do but I guess this is an improvement already. I'm sure you'll find more bugs if you try checking the other closures :)

GuillaumeGomez:
Sending more PRs later then!

PR #379, opened 2021-03-23
Fix use of gflags attribute in the subclass example

federicomenaquintero:
This bit me in https://gitlab.gnome.org/GNOME/librsvg/-/merge_requests/490#note_1064620 - the example in the subclass docs seems wrong.


sdroege (approved):

GuillaumeGomez (approved):
Great addition, thanks!

bilelmoussaoui (approved):

PR #380, opened 2021-03-23
Add clone checks 3

GuillaumeGomez:
I also realized that we had a test that wasn't supposed to be here so I moved it to its own file.

### DIFF
--- a/glib/tests/derive.rs
+++ b/glib/tests/derive.rs
@@ -1,0 +1,33 @@
+use std::marker::PhantomData;
+
+use glib::{Downgrade, Object};
+
+#[test]
+fn derive_downgrade() {
+    #[derive(Downgrade)]
+    pub struct NewType(Object);
+
+    #[derive(Downgrade)]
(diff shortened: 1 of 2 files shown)


glib/tests/derive.rs
```diff
@@ -0,0 +1,33 @@
+use std::marker::PhantomData;
+
+use glib::{Downgrade, Object};
+
+#[test]
+fn derive_downgrade() {
```
sdroege, line 6: Maybe add some tests here that this actually derives something meaningful and can be used, not just that it compiles :)

Also don't we have derive tests for this inside `glib-macros/tests` already?

sdroege (commented):
Looks good otherwise, thanks!

glib/tests/derive.rs
GuillaumeGomez: I have no idea. I simply moved this test out of the clone file. Don't know why it was there in the first place...

Issue #381, opened 2021-03-23
Should Win32 streams use AsRawHandle rather than IntoRawHandle?


federicomenaquintero:
Warning: I don't know what I'm doing, I just ran into this while implementing librsvg's stdin/stdout for win32.

Gtk-rs has these:

```rust
impl Win32InputStream {
    pub unsafe fn new<T: IntoRawHandle>(handle: T) -> Win32InputStream { ... }
}

impl Win32OutputStream {
    pub unsafe fn new<T: IntoRawHandle>(handle: T) -> Win32OutputStream { ... }
}
```

However, `std::io::stdin()` implements `AsRawHandle`, not `IntoRawHandle`.  The latter is only for `ChildStdin`, which is from `std::process::Child`.  Stdout is analogous.

I think gtk-rs may want `T: AsRawHandle` in the `new()` functions, but I haven't had a chance to actually try to compile code like this on Windows:

```
    pub fn stream() -> InputStream {
        let stream = unsafe { Win32InputStream::new(io::stdin().as_raw_handle()) };
        stream.upcast::<InputStream>()
    }
```

federicomenaquintero:
Sort of proof here - https://github.com/gtk-rs/gtk-rs/pull/382/checks?check_run_id=2180023075

pbor:
Mmm... this is the same as `UnixInputStream` that takes a `IntoRawFd`. In both cases the pass `true` to the underlying `close` flag so they are taking ownership of the fd/handle.

Maybe the right thing to do while we are still in time is to have two constructors (eg _new and _take), one that takes ownership and one that does not. This would be functionally equivalent to the glib API which takes the close param.

sdroege:
I think that would be the correct approach here, also for the UNIX variant. @pbor you make a PR?

pbor:
Using _new() and _take() is problematic because _new() would change semantic and silently broke existing code.

I will go with adding _without_close() variant of the constructor, but if anyone has better ideas on naming, I would love to hear them :)

pbor:
On irc/matrix we decided to go with take_fd and with_fd

PR #384, opened 2021-03-24
Replace _new constructor for Unix/Win32 Input and Output streams

pbor:
Add two variants, take_fd/handle and with_fd/handle which respectively
take ownership or not of the fd/handle. This models the "close"
boolean of the C api.

Fixes https://github.com/gtk-rs/gtk-rs/issues/381

### DIFF
--- a/gio/src/win32_input_stream.rs
+++ b/gio/src/win32_input_stream.rs
@@ -41,9 +41,18 @@ impl fmt::Display for Win32InputStream {
 impl Win32InputStream {
     #[doc(alias = "g_win32_input_stream_new")]
     #[allow(clippy::missing_safety_doc)]
-    pub unsafe fn new<T: IntoRawHandle>(handle: T) -> Win32InputStream {
+    pub unsafe fn take_handle<T: IntoRawHandle>(handle: T) -> Win32InputStream {
         let handle = handle.into_raw_handle();
         let close_handle = true.to_glib();
+        InputStream::from_glib_full(ffi::g_win32_input_stream_new(handle, close_handle))
+            .unsafe_cast()
+    }
+
+    #[doc(alias = "g_win32_input_stream_new")]
+    #[allow(clippy::missing_safety_doc)]
+    pub unsafe fn with_handle<T: AsRawHandle>(handle: T) -> Win32InputStream {
+        let handle = handle.as_raw_handle();
+        let close_handle = false.to_glib();
         InputStream::from_glib_full(ffi::g_win32_input_stream_new(handle, close_handle))
             .unsafe_cast()
     }
--- a/gio/src/win32_output_stream.rs
+++ b/gio/src/win32_output_stream.rs
@@ -41,9 +41,18 @@ impl fmt::Display for Win32OutputStream {
 impl Win32OutputStream {
     #[doc(alias = "g_win32_output_stream_new")]
     #[allow(clippy::missing_safety_doc)]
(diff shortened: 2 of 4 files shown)


gio/src/win32_input_stream.rs
```diff
@@ -41,12 +41,21 @@ impl fmt::Display for Win32InputStream {
 impl Win32InputStream {
     #[doc(alias = "g_win32_input_stream_new")]
     #[allow(clippy::missing_safety_doc)]
-    pub unsafe fn new<T: IntoRawHandle>(handle: T) -> Win32InputStream {
+    pub unsafe fn take_handle<T: IntoRawHandle>(handle: T) -> Win32InputStream {
```
federicomenaquintero, line 44: Should this be `T: AsRawHandle`?

gio/src/win32_output_stream.rs
```diff
@@ -41,12 +41,21 @@ impl fmt::Display for Win32OutputStream {
 impl Win32OutputStream {
     #[doc(alias = "g_win32_output_stream_new")]
     #[allow(clippy::missing_safety_doc)]
-    pub unsafe fn new<T: IntoRawHandle>(handle: T) -> Win32OutputStream {
+    pub unsafe fn take_handle<T: IntoRawHandle>(handle: T) -> Win32OutputStream {
```
federicomenaquintero, line 44: Same here.

gio/src/win32_input_stream.rs
pbor: No, the idea is that `take_handle` takes ownership, while `with_handle` takes a ref. Or am I misunderstanding your comment?

federicomenaquintero: Oh, I mean my original question from #381.  For rsvg-convert's stdin/stdout on Windows I want to do something like `Win32InputStream::new(io::stdin())` but that doesn't work, because [`IntoRawHandle`](https://doc.rust-lang.org/std/os/windows/io/trait.IntoRawHandle.html) is only implemented for things like `ChildStdin` and `ChildStdout`.  However, `io::stdin()` returns a `Stdin`, which implements [`AsRawHandle`](https://doc.rust-lang.org/std/os/windows/io/trait.AsRawHandle.html).

I don't really know the Windows API and am just looking at which Rust API bits I can use to get from `io::stdin()/stdout()` to a gio stream.  So, I'm not sure if gtk-rs uses `IntoRawHandle` incorrectly there, when (I think) it should be `AsRawHandle`.  That would make `Win32InputStream::new(io::stdin())` work, since `Stdin` impls `AsRawHandle`.  (`new` or `with_handle` eventually, I think.)

federicomenaquintero: To be clear: while the "Into" and "As" prefixes from those traits indeed match the desired behavior of consuming/borrowing, `IntoRawHandle` is not implemented for `Stdin/Stdout` (looks like an oversight in `std`?).

pbor: I do not know if it is an oversight or a limitation in std.

However Win32InputStream etc can be used for more type of handles beside stdin/stdout, so I think it is correct to have both variants.

For librsvg just use the `with_handle` constructor

sdroege (approved):
Makes sense to me and looks correct.

@federicomenaquintero this should cover your use-case unless I'm missing something?

gio/src/win32_input_stream.rs
federicomenaquintero: You are right, `with_handle` should work fine for librsvg.  Thanks!

PR #385, opened 2021-03-24
generator: Iterate gir directories passed on the command line

MarijnS95:
This change cleans up handling of the default gir directory, and makes sure the `-d` flag is prepended to every path. Before the command would contain a single `-d` flag followed by a `list` of paths, which cannot be passed to an executable verbatim:

    The following error occurred: expected str, bytes or os.PathLike object, not list
    Do you want to continue? [y/N] ^CTraceback (most recent call last):
      File "./generator.py", line 189, in <module>

Fixes: c33495c ("Don't set a default value for the -d option")

CC @GuillaumeGomez

### DIFF
--- a/generator.py
+++ b/generator.py
@@ -79,9 +79,9 @@ def regen_crates(path, conf):
     elif path.match("Gir*.toml"):
         print('==> Regenerating "{}"...'.format(path))
 
-        args = [conf.gir_path, "-c", path, "-o", path.parent]
-        if len(conf.gir_files_paths) > 0:
-            args += ["-d"] + [conf.gir_files_paths]
+        args = [conf.gir_path, "-c", path, "-o", path.parent] + [
+            d for path in conf.gir_files_paths for d in ("-d", path)
+        ]
         if path.parent.name.endswith("sys"):
             args.extend(["-m", "sys"])
         error = False
@@ -136,7 +136,7 @@ def parse_args():
     )
     parser.add_argument(
         "--gir-files-directories",
-        nargs='+', # If the option is used, we expect at least one folder!
+        nargs="+",  # If the option is used, we expect at least one folder!
         dest="gir_files_paths",
         default=[],
         type=directory_path,
@@ -165,7 +165,7 @@ def parse_args():
 def main():
     conf = parse_args()
 
-    if len(conf.gir_files_paths) == 0:
+    if not conf.gir_files_paths:
         if def_check_submodule(DEFAULT_GIR_FILES_DIRECTORY, conf) == FAILURE:
             return 1
 


GuillaumeGomez:
The option is definitely not doing the same anymore. Before it was to provide the path for gir directories, now it's just to prepend before all paths, meaning that we have no way to override the `gir-directories` option anymore in `generator.py`.

MarijnS95:
> GuillaumeGomez changed the title generator: Iteraterate gir directories passed on the command line generator: Iterate gir directories passed on the command line 1 hour ago

Whoa, that was a good catch (also in the commit title)!



> The option is definitely not doing the same anymore. Before it was to provide the path for gir directories, now it's just to prepend before all paths, meaning that we have no way to override the `gir-directories` option anymore in `generator.py`.

@GuillaumeGomez What do you mean with "no way to override"? The `default=[DEFAULT_GIR_FILES_DIRECTORY]` _is_ overridden if an argument is supplied, try for yourself with `print(conf.gir_files_paths)` and pass an existing directory to `--gir-files-directories`.

Keep in mind that python-argparse works different from the parser in `gir`. Help for the generator shows: `[--gir-files-directories GIR_FILES_PATHS [GIR_FILES_PATHS ...]]`, meaning multiple options for this argument must be specified with spaces (weird if you want to pass positional arguments after that).

Rusts `getopts` on the other hands expects repetition by specifying the flag multiple times:

```console
me:gstreamer-rs/ $ ./gir/target/release/gir -d foo,bar
[src/main.rs:96] &matches.opt_strs("d") = [
    "foo,bar",
]
me:gstreamer-rs/ $ ./gir/target/release/gir -d foo bar
[src/main.rs:96] &matches.opt_strs("d") = [
    "foo",
]
me:gstreamer-rs/ $ ./gir/target/release/gir -d foo -d bar
[src/main.rs:96] &matches.opt_strs("d") = [
    "foo",
    "bar",
]
```

This is why this commit prefixes every path in the `gir_files_paths` array with `-d`. That option doesn't work without arguments

```console
me:gstreamer-rs/ $ ./gir/target/release/gir -d
Argument to option 'd' missing
```

`Config::new` _only looks up gir directories in toml files if nothing was specified on the cmdline_. You wrote it that way:

https://github.com/gtk-rs/gir/blob/de641b48053213a84b5440772bf4f5f16addc9d3/src/config/config.rs#L163-L177

What exactly do you want to override?

GuillaumeGomez:
My bad, it was actually the opposite: I removed the default value set in the `-d` option so that by default, it was using the one presen t in the `Gir.toml` file. Now, you simply override it by default every time; which we don't want.

MarijnS95:
> Now, you simply override it by default every time; which we don't want.

That was happening before already. More concretely I `DEFAULT_GIR_FILES_DIRECTORY` should be removed entirely?

GuillaumeGomez:
> > Now, you simply override it by default every time; which we don't want.
> 
> That was happening before already (but indeed, removed by your PR as the title implies, except that a stray `-d` was left in place).

Yes, which is why I removed it. :)

> More concretely `DEFAULT_GIR_FILES_DIRECTORY` should be removed entirely?

Is there a reason to remove it?

MarijnS95:
@GuillaumeGomez Well, not entirely after all because of the default case where this directory is synced, so it has been left in place.

Now this PR is merely fixing the `-d` that should come in pairs with paths.

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

GuillaumeGomez:
Can you send this fix to the gtk4-rs repository too please?

PR #386, opened 2021-03-25
Move examples to separate folders

Hofer-Julian:


GuillaumeGomez:
Why advantage does it bring? We still have the same amount of entries in the source folder but now we have more levels.

Hofer-Julian:
The advantage of the intermediate step is indeed limited. It is not true though, that we have the same amount of entries, you probably forgot about the ui files.

The next step would be to add one README per example.

I would then suggest to discuss within a follow-up issue:
1. If we really want to keep all examples
2. How we could structure them, so that especially newcomers know where to look.

sdroege:
> The next step would be to add one README per example.

And to also add a screenshot to each UI example.

I think overall this makes going through the examples much more friendly and less confusing for users. There's not just a collection of source files but a proper overview and pre-example documentation, and everything belong to an example is in the same directory instead of having everything put together.

Hofer-Julian:
I think CI will be fixed as soon as https://github.com/gtk-rs/gtk-rs/pulls is merged and I rebased.

Hofer-Julian:
@sdroege you requested a review, this is still a a draft PR though.

One other thing, I don't really know how to resolve the gir conflict.

sdroege:
> One other thing, I don't really know how to resolve the gir conflict.

Reset it to the version that is in master :)

> @sdroege you requested a review, this is still a a draft PR though.

It would still be good to get some opinions on it before it's finished IMHO

Hofer-Julian:
> Reset it to the version that is in master :)

Could you please tell me the command?
The web does not seem to be very helpful here.

sdroege:
Just rebasing on top of master should fix this already in your case.

examples/build_travis.sh
```diff
@@ -1,42 +0,0 @@
-#!/bin/sh
```
GuillaumeGomez, line 1: Good catch!

GuillaumeGomez:
So, overall this seems nice. Now I have a big issue with adding images directly in this repository. It increases the size a lot for finally small gain. I like the idea of having screenshots though, so if someone has an idea on how to go around this issue...

Hofer-Julian:
> So, overall this seems nice. Now I have a big issue with adding images directly in this repository. It increases the size a lot for finally small gain. I like the idea of having screenshots though, so if someone has an idea on how to go around this issue...

[Git-lfs](https://git-lfs.github.com/)

Not sure if adding images which don't change are a big problem though

GuillaumeGomez:
On the repository definitely. I don't want to download images to update code. Also, unless stricly needed, we'd better stay as far away as possible from git-lfs.

Hofer-Julian:
Host the images on a website and link them there.

I would like to mention that most screenshots were already there, so the repo size would only shrink if we change history (unless I misunderstand how git works)

Hofer-Julian:
> Also, unless stricly needed, we'd better stay as far away as possible from git-lfs.

Wouldn't mind if you'd have the same policy for git submodules :P

GuillaumeGomez:
> I would like to mention that most screenshots were already there, so the repo size would only shrink if we change history (unless I misunderstand how git works)

Damn, I didn't clean up things before merging repositories... T_T

git submodules are necessary evil here. :p

Hofer-Julian:
Not sure why the regen check fails.
Hints would be very welcome :)

Hofer-Julian:
Apart from the CI failure it is ready for review now.

sdroege:
The CI failure should be gone if you rebase on top of latest master

sdroege:
@GuillaumeGomez What do we do with the images? They're there already so let's keep them for now and get rid of them at a later time separately?

GuillaumeGomez:
For the existing images, there is nothing to do unless rewriting the whole history. But I really don't know if this is a good idea to add new ones...

sdroege:
Doesn't seem like much of a problem to me, they're small and not changing often.

GuillaumeGomez:
30KB x the number of added images is definitely not small in a git repository. I really don't know what to do here...

sdroege:
How much is it in sum that is added as part of this PR?

GuillaumeGomez:
Around 88KB.

sdroege:
Why are we even talking about this then?

Hofer-Julian:
I rebased and squashed the commits now.

I hope that the CI is fine now and that we can move the binaries discussion to a separate issue.

sdroege (approved):

Hofer-Julian:
regen check still fails :(

Hofer-Julian:
If one of you knows a magic command which might fix the CI, please feel free to push to this branch.

sdroege:
I assume this already fails before your PR then, i.e. master currently fails the CI. There's nothing in your PR here that should break it.

GuillaumeGomez (approved):

GuillaumeGomez:
Just rebasing on master should be enough for the failure so that's not a problem. I'm really sad not having see the images before. Well, what's done is done. Thanks for moving the examples, I really like how they look now!

Hofer-Julian:
> Thanks for moving the examples, I really like how they look now!

Glad to hear that you like it as well :)

PR #395, opened 2021-03-28
Fix #392 export gdk::event::FromEvent

### LINKED ISSUE #392: Export gdk::event::FromEvent
My usecase - I'm using builder's [`connect_signals`](https://gtk-rs.org/docs/gtk/prelude/trait.BuilderExtManual.html#tymethod.connect_signals) and need to convert the `glib::Value`s to their concrete types. For events, I can't convert from the value to the concrete event type (`EventButton`, `EventConfigure` etc.) directly - I have to first use `to_value()` to convert the general `gdk::Event` type and then use [`downcast()`](https://docs.rs/gdk/0.13.2/gdk/struct.Event.html#method.downcast) to convert to the concrete event type.

When I'm doing this manually it works well, but when I'm trying to add some mechanisms to make it more ergonomic I have a problem, because `FromEvent` is private and my generic functions can't use it as a trait bound.

Ideally I want to write something like this, on my custom type that contains the signal information:
```rust
pub fn param_event<E: gdk::FromEvent>(&self, index: usize) -> Result<E, crate::Error> {
    let event: gdk::Event = self.param(index)?;
    event.downcast().map_err(|err| {
        // Format a proper error
    })
}
```

Which will get used like this:
```rust
let event: gdk::EventButton = msg.param_event(1)?;
```

But I can't - because `FromEvent` is private - so instead my API looks like this:
```rust
let event: gdk::EventButton = msg.param::<gdk::Event>(1)?.downcast().unwrap();
```

Is there a reason why `FromEvent` is not exported? Can it be exported?

idanarye:


gdk/src/lib.rs
```diff
@@ -100,6 +100,7 @@ pub use crate::atom::TARGET_PIXMAP;
 pub use crate::atom::TARGET_STRING;
 pub use crate::change_data::ChangeData;
 pub use crate::event::Event;
+pub use crate::event::FromEvent;
```
bilelmoussaoui, line 103:
```suggestion
pub use crate::event::{Event, FromEvent};
```

idanarye: I was following the `atom` exports example from above, but OK.

sdroege (approved):

PR #397, opened 2021-03-30
Examples fixes

Hofer-Julian:
We missed a bit during https://github.com/gtk-rs/gtk-rs/pull/386

Now, the commands don't look so weird anymore.

Also the description of `list_box_model` is now fixed.

I also took the opportunity to split up `list_box_model` into separate files which improves readability quite a bit in my opinion.
If you want me to, I can do this for other examples as well.


examples/list_box_model/main.rs
```diff
@@ -1,88 +1,13 @@
+mod model;
+pub mod row_data;
```
sdroege, line 2: Why `pub`?

sdroege:
> I also took the opportunity to split up `list_box_model` into separate files which improves readability quite a bit in my opinion.

Personally I even go one step further in my code and have the `imp` module not inline but as a separate `imp.rs` file next to `mod.rs` with the outer part.

But this is also a huge improvement already.

>  If you want me to, I can do this for other examples as well.

That would be great!

examples/list_box_model/main.rs
Hofer-Julian: model accesses it IIRC

Hofer-Julian: It could be pub (crate) though

sdroege: Oh of course, sorry :)

sdroege:
> If you want me to, I can do this for other examples as well.

Want to do this as part of this PR or separately?

Hofer-Julian:
I'll probably not work on it today so maybe let's merge this one and I'll open another one :)

PR #398, opened 2021-03-30
WIP: CI: use gnome image

bilelmoussaoui:


bilelmoussaoui:
This can't work as we can't install new things inside the image :/ Will figure out something else using the current image we have.

sdroege:
Can't we use it as a base image and install other things on top of it?

Issue #399, opened 2021-03-30
ci: `regen check` fails currently


sdroege:
See e.g. https://github.com/gtk-rs/gtk-rs/pull/397/checks?check_run_id=2227220881

PR #400, opened 2021-03-31
be able to inspect pango attribute types and downcast them

bvinc:
Previously there was no way to determine the type of a pango attribute.  I added support for this, and the ability to downcast to actually read pango attributes.

Do you think I should go ahead and modify each attribute constructor to return a typed attribute and modify `AttrList`'s `change`, `insert`, and `insert_before` to take in any `IsAttribute`?

Any suggestions welcome.

### DIFF
--- a/pango/src/attribute.rs
+++ b/pango/src/attribute.rs
@@ -243,7 +243,7 @@ macro_rules! define_attribute_struct {
         #[cfg(any(feature = "v1_44", feature = "dox"))]
         #[cfg_attr(feature = "dox", doc(cfg(feature = "v1_44")))]
         glib::wrapper! {
-            #[derive(Debug, PartialOrd, Ord, Hash)]
+            #[derive(Debug)]
             pub struct $rust_type(Boxed<$ffi_type>);
 
             match fn {
@@ -255,7 +255,7 @@ macro_rules! define_attribute_struct {
 
         #[cfg(not(any(feature = "v1_44", feature = "dox")))]
         glib::wrapper! {
-            #[derive(Debug, PartialOrd, Ord, Hash)]
+            #[derive(Debug)]
             pub struct $rust_type(Boxed<$ffi_type>);
 
             match fn {
@@ -263,27 +263,6 @@ macro_rules! define_attribute_struct {
                 free => |ptr| ffi::pango_attribute_destroy(ptr as *mut ffi::PangoAttribute),
             }
         }
-
-        impl $rust_type {
-            #[doc(alias = "pango_attribute_equal")]
-            fn equal(&self, attr2: &$rust_type) -> bool {
-                unsafe {
-                    from_glib(ffi::pango_attribute_equal(
-                        self.to_glib_none().0 as *const ffi::PangoAttribute,
-                        attr2.to_glib_none().0 as *const ffi::PangoAttribute,
-                    ))
-                }
-            }
-        }
-
-        impl PartialEq for $rust_type {
-            #[inline]
-            fn eq(&self, other: &Self) -> bool {
-                self.equal(other)
-            }
-        }
-
-        impl Eq for $rust_type {}
 
         impl IsAttribute for $rust_type {
             const ATTR_TYPES: &'static [AttrType] = $attr_types;
--- a/pango/src/attribute.rs
+++ b/pango/src/attribute.rs
@@ -1,9 +1,13 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
 use crate::AttrClass;
+use crate::AttrType;
 use crate::Attribute;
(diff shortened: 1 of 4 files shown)


bvinc:
I don't think the regen check failure is my fault.

sdroege:
> I don't think the regen check failure is my fault.

Yes, that's https://github.com/gtk-rs/gtk-rs/issues/399

pango/src/attribute.rs
```diff
@@ -200,4 +204,200 @@ impl Attribute {
...
+            if T::ATTR_TYPES.contains(&self.get_attr_class().type_()) {
+                Some(&*(self as *const Attribute as *const T))
+            } else {
+                None
+            }
+        }
+    }
+}
+
+pub trait IsAttribute:
```
sdroege, line 229:
```suggestion
pub unsafe trait IsAttribute:
```

pango/src/attribute.rs
```diff
@@ -200,4 +204,200 @@ impl Attribute {
...
+            match fn {
+                copy => |ptr| ffi::pango_attribute_copy(ptr as *const ffi::PangoAttribute) as *mut $ffi_type,
+                free => |ptr| ffi::pango_attribute_destroy(ptr as *mut ffi::PangoAttribute),
+                get_type => || ffi::pango_attribute_get_type(),
+            }
+        }
+
+        #[cfg(not(any(feature = "v1_44", feature = "dox")))]
+        glib::wrapper! {
+            #[derive(Debug, PartialOrd, Ord, Hash)]
```
sdroege, line 258: As clippy suggests, if you implement `PartialEq` / `Eq` manually, you must implement `Hash` manually too and it must be consistent with those manual `PartialEq` / `Eq` implementations (equal must mean same hash).

sdroege (commented):
Looks good to me otherwise

pango/src/attribute.rs
```diff
@@ -200,4 +204,200 @@ impl Attribute {
...
+                        attr2.to_glib_none().0 as *const ffi::PangoAttribute,
+                    ))
+                }
+            }
+        }
+
+        impl PartialEq for $rust_type {
+            #[inline]
+            fn eq(&self, other: &Self) -> bool {
+                self.equal(other)
```
bilelmoussaoui, line 282: Why not call pango_attribute_equal here directly ?

pango/src/attribute.rs
bvinc: What makes it need unsafe?  I'm following the examples of `ParamSpec` and `RenderNode`.  They both have a similar type and don't need unsafe.

pango/src/attribute.rs
```diff
@@ -200,4 +204,200 @@ impl Attribute {
...
+
+pub trait IsAttribute:
+    FromGlibPtrFull<*const ffi::PangoAttribute>
+    + FromGlibPtrFull<*mut ffi::PangoAttribute>
+    + std::convert::AsRef<crate::Attribute>
+    + 'static
+{
+    const ATTR_TYPES: &'static [AttrType];
+    fn upcast(self) -> Attribute;
+    fn upcast_ref(&self) -> &Attribute;
```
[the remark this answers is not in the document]
bvinc: I'm not sure what you mean by "implement the casts generically".

pango/src/attribute.rs
bvinc: I see the problem.  I basically copied this from `src/auto/attribute.rs`, so the same problem exists there already.

We have this nice equals function provided for us, but no hash function. It would be silly to try to match a Hash function to work on all the same values.

I think the only thing that makes sense is to not derive `Hash` or `PartialEq` / `Eq` and just provide the equals function as a method.

pango/src/attribute.rs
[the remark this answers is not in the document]
sdroege: There's only one possible and correct implementation for the casts. But if you make the trait unsafe it doesn't really matter anymore, it's up to the implementer to do it right then.

pango/src/attribute.rs
sdroege: Maybe don't derive `Hash` then, and if someone needs it they can figure out a way around it.

sdroege: Why did you remove the `PartialEq` / `Eq` implementation around `pango_attribute_equal()` together with the others? IMHO that one should stay, `Hash` / `PartialOrd` / `Ord` should go away.

pango/src/attribute.rs
sdroege: I created issues for marking those other two traits as unsafe. Can you update this PR accordingly? Thanks :)

GuillaumeGomez:

PR #408, opened 2021-04-02
Refactor CSS and dialog example

Hofer-Julian:
- Put CSS into separate file
- Rename and simplify dialog example a bit


examples/css/style.css
```diff
@@ -0,0 +1,28 @@
...
+
+combobox button.combo box {
+    padding: 5px;
+}
+combobox box arrow {
+    -gtk-icon-source: none;
+    border-left: 5px solid transparent;
+    border-right: 5px solid transparent;
+    border-top: 5px solid black;
+}
```
sdroege, line 28: Better to have a newline at the end

sdroege (commented):
Looks good to me otherwise

examples/dialog_async/main.rs
```diff
@@ -36,30 +20,41 @@ fn build_ui(application: &gtk::Application) {
         .visible(true)
         .build();
 
-    let window = Rc::new(
```
sdroege, line 39: The `Rc` here was very strange

examples/css/style.css
Hofer-Julian: Done.
Also added the setting to my editor, even though I never understood what difference it makes.

examples/dialog_async/main.rs
Hofer-Julian: Yep...

I also found the image holder in cairo_threads a bit too over-engineered.
Maybe it should also be simplified
https://github.com/gtk-rs/gtk-rs/blob/master/examples/cairo_threads/image/image_holder.rs

Hofer-Julian: To be more explicit, remove the image_holder altogether and wrap `Image.data` in a Rc<Refcell<>>
https://github.com/gtk-rs/gtk-rs/blob/master/examples/cairo_threads/image/mod.rs

sdroege: Can you show what you're thinking of?

Hofer-Julian: Sure will open another PR

examples/css/style.css
sdroege: It make diffs more useful to read later at least :)

PR #411, opened 2021-04-04
cairo: don't include freetype by default

bilelmoussaoui:
with this, all the gtk apps pulls in the freetype-crate with it dependencies without actually using it.


cairo/Cargo.toml
```diff
@@ -25,7 +25,7 @@ use_glib = ["glib", "ffi/use_glib"]
 embed-lgpl-docs = ["gtk-rs-lgpl-docs"]
 v1_14 = ["ffi/v1_14"]
 v1_16 = ["v1_14", "ffi/v1_16"]
-default = ["use_glib", "freetype"]
+default = ["use_glib"]
```
sdroege, line 28: Then `dox` needs to include `freetype`

sdroege (approved):
Otherwise, a good idea!

PR #412, opened 2021-04-05
Add ValueArray::len()/is_empty()

elmarco:
Those are expected for an array type, since we have getters on
ValueArray itself (along with the [Value] deref).


sdroege (changes requested):
Makes sense but please remove the `gir-files` update from the commit :)

sdroege (approved):

bilelmoussaoui (approved):

PR #414, opened 2021-04-05
glib: Implement StaticType, FromValue(Optional)? and SetValue(Optional)? for Value

jplatte:
I still feel like I have no idea what I'm doing, but at least this compiles.

### DIFF
--- a/glib/src/value.rs
+++ b/glib/src/value.rs
@@ -336,6 +336,12 @@ impl Uninitialized for Value {
     }
 }
 
+impl StaticType for Value {
+    fn static_type() -> Type {
+        unsafe { from_glib(gobject_ffi::g_value_get_type()) }
+    }
+}
+
 #[doc(hidden)]
 impl<'a> ToGlibPtr<'a, *const gobject_ffi::GValue> for Value {
     type Storage = &'a Value;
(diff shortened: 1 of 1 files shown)


glib/src/value.rs
```diff
@@ -336,6 +336,12 @@ impl Uninitialized for Value {
     }
 }
 
+impl StaticType for Value {
+    fn static_type() -> Type {
+        Type::value()
```
sdroege, line 341:
```suggestion
        unsafe { from_glib(gobject_ffi::g_value_get_type()) }
```

Why not just call the function here directly? That seems easier :)

glib/src/value.rs
```diff
@@ -930,6 +936,33 @@ impl SetValueOptional for str {
     }
 }
 
+impl<'a> FromValue<'a> for Value {
+    unsafe fn from_value(value: &'a Value) -> Self {
+        from_glib_none(
+            gobject_ffi::g_value_get_boxed(value.to_glib_none().0) as *mut gobject_ffi::GValue
```
sdroege, line 942: That works, is correct and also has the lowest number of copies :)

For more correctness:

```suggestion
            gobject_ffi::g_value_get_boxed(value.to_glib_none().0) as *const gobject_ffi::GValue
```


You could also add a

```rust
impl<'a> FromValue<'a> for &'a Value {
    unsafe fn from_value(value: &'a Value) -> Self {
        &*(
            gobject_ffi::g_value_get_boxed(value.to_glib_none().0) as *const Value
        )
    }
}
```

And the corresponding `FromValueOptional` impl

jplatte, line 942: Not sure why I wrote `*mut`, this probably has to be `*const`?

glib/src/value.rs
```diff
@@ -930,6 +936,33 @@ impl SetValueOptional for str {
...
+
+impl<'a> FromValueOptional<'a> for Value {
+    unsafe fn from_value_optional(value: &'a Value) -> Option<Self> {
+        Some(FromValue::from_value(value))
+    }
+}
+
+impl SetValue for Value {
+    unsafe fn set_value(value: &mut Value, this: &Self) {
+        gobject_ffi::g_value_set_boxed(value.to_glib_none_mut().0, this.to_glib_full() as _)
```
sdroege, line 955:
```suggestion
        gobject_ffi::g_value_set_boxed(value.to_glib_none_mut().0, this.to_glib_none().0 as ffi::gpointer)
```

Otherwise you leak memory: `full` creates a copy, `set_boxed` does not take ownership and also creates a copy

glib/src/value.rs
```diff
@@ -930,6 +936,33 @@ impl SetValueOptional for str {
...
+    }
+}
+
+impl SetValueOptional for Value {
+    unsafe fn set_value_optional(value: &mut Value, this: Option<&Self>) {
+        let ptr = this.map_or_else(ptr::null, |t| t.to_glib_full() as _);
```
sdroege, line 961: Same as above here. For simplicity I would probably just go with

```rust
if let Some(this) = this {
    gobject_ffi::g_value_set_boxed(value.to_glib_none_mut().0, this.to_glib_none().0 as ffi::gpointer);
} else {
    gobject_ffi::g_value_set_boxed(value.to_glib_none_mut().0, ptr::null_mut());
}
```

glib/src/value.rs
jplatte: Because `Type`s representation is private and I couldn't find a constructor that takes a `GType`.

glib/src/value.rs
sdroege: See below. `const` is more correct but it also doesn't really make a difference

glib/src/value.rs
sdroege: Ah. `from_glib` is what you're looking for

glib/src/value.rs
jplatte: This pattern from above is common in other parts of gtk-rs, except it's an explicit `match` for the pointer rather than `map_or_else`. Why is what you wrote "simpler", just because `map_or_else` is an uncommon combinator?

sdroege: Because in theory you have to keep the stash returned by `this.to_glib_none()` alive and only then pass the `.0` to the function. So the correct way with `map_or_else()` would be

```rust
let ptr = this.map(|t| t.to_glib_none()); // so that it stays alive
let ptr = ptr.map_or_else(ptr::null, |t|  t.0 as ffi::gconstpointer);
gobject_ffi::g_value_set_boxed(value.to_glib_none_mut().0, ptr);
```

jplatte: But your first suggestion also has `to_glib_none()` in a temporary that is dropped before the call to `g_value_get_boxed` 😅

sdroege: No, because it's in the same statement the return value of `to_glib_none()` is only dropped afterwards :)

jplatte: Huh, I did not know that. Thanks for the explanation!

sdroege (approved):

glib/src/value.rs
sdroege: Otherwise you couldn't e.g. do `foo(&*some_ref_cell.borrow())` for example.

PR #416, opened 2021-04-06
glib: Only allow canonical names in property and signal builders

### LINKED ISSUE #409: glib: Auto conversion of property and signal names
Currently if you define a property in your ObjectImpl with an underscored name:
```rust
glib::ParamSpec::boolean(
    "underscored_name",
    "underscored_name",
    "underscored_name",
    true,
    glib::ParamFlags::READWRITE,
)
```
It is silently converted into underscored-name (with a hyphen[|dash|minus]).

If you try to use the property:
```rust
glib::Object::new(&[("underscored_name",&false)])
```
This results in an error when you try to use the property of `Can't find property "underscored_name"`

We should be consistent in applying the auto-conversion of property signal names.

I suggest disallowing the creation of invalid property/signal names in the ObjectImpl rather than applying auto-conversion in the rest of the API.

mehmooda:
Fixes #409
Panics if handed a non canonical name

### DIFF
--- a/glib/src/param_spec.rs
+++ b/glib/src/param_spec.rs
@@ -1,6 +1,7 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
 use crate::translate::*;
+use crate::utils::is_canonical_pspec_name;
 use crate::value;
 use crate::ParamFlags;
 use crate::StaticType;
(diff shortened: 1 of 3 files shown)


sdroege (approved):

GuillaumeGomez (approved):

GuillaumeGomez:
Just waiting on @bilelmoussaoui now. :)

bilelmoussaoui (approved):

Issue #418, opened 2021-04-06
The trait `glib::IsA<O>` is not implemented


elmarco:
The following code should work, but fails with missing IsA<O> trait implementations errors:

```rust
        impl<O: IsA<Foo>> FooExt for O {
            fn test(&self) {
                let self_ = imp::Foo::from_instance(self.downcast_ref::<Foo>());
                unimplemented!()
            }
        }
```

See related test I am about to submit.

sdroege:
I think this is missing an `impl<T: ObjectType + etc> IsA<T> for T {}` in glib, or otherwise this can't really work correctly.

sdroege:
@elmarco Want to give it a try if that solves your problem?

elmarco:
> @elmarco Want to give it a try if that solves your problem?

I tried, and it doesn't work. I tried to play with the traits and generic impl, but it's out of my league I am afraid :)

sdroege:
What did you try and what did the compiler tell you?

elmarco:
I tried adding `unsafe impl<T: ObjectType + AsRef<T>> IsA<T> for T {}` (and removing `unsafe impl $crate::object::IsA<$name> for $name { }`). That seems ok. But I still have to write:

`impl<O: IsA<Foo> + IsA<O>> FooExt for O where Foo: IsA<O>`

And I tried various ways to have generic IsA<O> for O somehow, but tripped on various errors and confusion...

mehmooda:
Your not mixing and matching git glib vs crates.io glib are you? This seems to be working for me

mehmooda:
although I did have to use
```rust
imp::LoginWidgetImp::from_instance(self.as_ref())
```

to get around the expected &MyWidget found &O

bilelmoussaoui:
Is this still an issue @elmarco ?

PR #419, opened 2021-04-06
macros: add a subclassable example that should work

elmarco:
Test for issue #418


elmarco:
moved to https://github.com/gtk-rs/gtk-rs-core/pull/81

PR #420, opened 2021-04-06
Improve Debug output for GString, SendValue

### LINKED ISSUE #417: Debug impl for Value is misleading
Example of the `Debug` output for a `SendValue` containing a `guint`:

```rust
SendValue(Value(GString(Foreign(0x7f6c008620, 24))))
```

This is the [Debug impl](https://docs.rs/glib/0.10.3/src/glib/value.rs.html#300-309):

```rust
impl fmt::Debug for Value {
    fn fmt(&self, f: &mut fmt::Formatter) -> Result<(), fmt::Error> {
        unsafe {
            let s: GString =
                from_glib_full(gobject_sys::g_strdup_value_contents(self.to_glib_none().0));

            f.debug_tuple("Value").field(&s).finish()
        }
    }
}
```

I don't know what it does exactly, but it looks like it always outputs a string for some reason. Also, IMHO the `SendValue()` and `Value()` wrapping in the output is not very useful.

What I'd love to see instead: `guint(10)` for both a `Value` and `SendValue`.

There is precedent for not only printing inner values in the standard library: [NonZero*](https://doc.rust-lang.org/src/core/num/nonzero.rs.html#11-23), [Rc](https://doc.rust-lang.org/src/alloc/rc.rs.html#1633-1637), even [Cow](https://doc.rust-lang.org/src/alloc/borrow.rs.html#377-387).

jplatte:
Fixes #417.

### DIFF
--- a/glib/src/gstring.rs
+++ b/glib/src/gstring.rs
@@ -15,10 +15,8 @@ use std::string::String;
 
 use crate::value::{FromValueOptional, SetValue, SetValueOptional, Value};
 
-#[derive(Debug)]
 pub struct GString(Inner);
 
-#[derive(Debug)]
 enum Inner {
     Native(Option<CString>),
     Foreign(*mut c_char, usize),
@@ -85,6 +83,12 @@ impl Clone for GString {
         let cstring = CString::new(self.as_str().to_string()).expect("CString::new failed");
 
         GString(Inner::Native(Some(cstring)))
+    }
+}
+
+impl fmt::Debug for GString {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        <&str as fmt::Debug>::fmt(&self.as_str(), f)
     }
 }
 
--- a/glib/src/value.rs
+++ b/glib/src/value.rs
@@ -271,6 +271,10 @@ impl Value {
     ) -> Result<SendValue, Self> {
         self.downcast::<T>().map(TypedValue::into_send_value)
     }
+
+    fn content_debug_string(&self) -> GString {
+        unsafe { from_glib_full(gobject_ffi::g_strdup_value_contents(self.to_glib_none().0)) }
+    }
 }
 
 impl Clone for Value {
@@ -295,12 +299,7 @@ impl Drop for Value {
 
 impl fmt::Debug for Value {
     fn fmt(&self, f: &mut fmt::Formatter) -> Result<(), fmt::Error> {
-        unsafe {
-            let s: GString =
-                from_glib_full(gobject_ffi::g_strdup_value_contents(self.to_glib_none().0));
-
-            f.debug_tuple("Value").field(&s).finish()
-        }
+        write!(f, "({}) {}", self.type_(), self.content_debug_string())
     }
 }
 
@@ -804,7 +803,7 @@ impl SendValue {
 
 impl fmt::Debug for SendValue {
     fn fmt(&self, f: &mut fmt::Formatter) -> Result<(), fmt::Error> {
-        f.debug_tuple("SendValue").field(&self.0).finish()
+        <Value as fmt::Debug>::fmt(&self.0, f)
     }
 }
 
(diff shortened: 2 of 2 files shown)


glib/src/gstring.rs
```diff
@@ -88,6 +86,12 @@ impl Clone for GString {
     }
 }
 
+impl fmt::Debug for GString {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{:?}", self.as_str())
```
sdroege, line 91:
```suggestion
       <&str as fmt::Debug>::fmt(self.as_str(), f)
```

That way you pass through all formatting options too.

sdroege (commented):
Looks good to me otherwise

glib/src/value.rs
```diff
@@ -271,6 +271,10 @@ impl Value {
     ) -> Result<SendValue, Self> {
         self.downcast::<T>().map(TypedValue::into_send_value)
     }
+
+    fn content_debug_string(&self) -> GString {
+        unsafe { from_glib_full(gobject_ffi::g_strdup_value_contents(self.to_glib_none().0)) }
```
sdroege, line 276: I completely forgot about this function. That's of course very useful here :)

GuillaumeGomez (approved):
Looks good to me minus @sdroege's comment.

glib/src/value.rs
jplatte: That was being used already 😉

jplatte: Given that we use that, do we still need the tuple wrapping?

glib/src/gstring.rs
jplatte: That's what I had first, but I thought it would be less readable 😄

glib/src/value.rs
jplatte: > It seems useful to show that this is actually a `glib::Value`.

Why though? Again, `Rc` and `NonZero*` from the stdlib don't show it either, even `Cow` doesn't show that it's `Cow` (and whether it's borrowed or owned).

> Does this also print the type btw?

I have no idea. I guess I'll write a unit test.

sdroege: Well, first printing the type and then the value :)

sdroege: > Why though? Again, Rc and NonZero* from the stdlib don't show it either, even Cow doesn't show that it's Cow (and whether it's borrowed or owned).

Personally I prefer that so I know what I'm looking at when debugging things, but you have a point there.

@bilelmoussaoui @GuillaumeGomez What do you think?

GuillaumeGomez: The whole point of a debug impl is to get information so I'd prefer what @sdroege is suggesting.

jplatte: The debug string only contains the value. Also I just realized that `debug_tuple("foo").field()` will quote the string again, which is not what we want here. With the current code,

```rust
let v = "foo".to_value();
assert_eq!(format!("{:?}", v), "foo");
```

fails with

```
thread 'value::tests::test_debug' panicked at 'assertion failed: `(left == right)`
  left: `"Value(\"\\\"foo\\\"\")"`,
 right: `"foo"`', glib/src/value.rs:1234:9
```

So now we know we can print both type (through `type_()`?) and value, does that mean I can go ahead with `{type}({debug_content})` and remove the `Value()` / `SendValue()`?

sdroege: Or maybe C syntax: `(type)value`? But OK with me then

jplatte: > The whole point of a debug impl is to get information

Yes, information that's not present in the code itself. You already know that you're debugging a `Value` because that's the type you declared yourself. But if you insist, I can print `Value()` around everything. In that case, how should the type be printed? Should it be `Value(<{type}> {debug_string})` / `Value({type}({debug_string}))` / something else?

jplatte: Is an extra space okay? I know that's not the usual formatting in C, but I find it really ugly otherwise. I've committed that for now, but if you want I'll remove it.

sdroege: I also thought of that but :man_shrugging: Whatever you prefer

glib/src/value.rs
```diff
@@ -295,12 +299,7 @@ impl Drop for Value {
 
 impl fmt::Debug for Value {
     fn fmt(&self, f: &mut fmt::Formatter) -> Result<(), fmt::Error> {
-        unsafe {
-            let s: GString =
-                from_glib_full(gobject_ffi::g_strdup_value_contents(self.to_glib_none().0));
-
-            f.debug_tuple("Value").field(&s).finish()
-        }
+        write!(f, "({}) {}", self.type_(), self.content_debug_string())
```
sdroege, line 302: If you feel like it, make `{:?}` print only the value and `{:#?}` with type (or the other way around). That's `Formatter::alternate()`

jplatte: I don't feel like it 😅

I would generally expect `#` to make things more verbose and here I wouldn't want to remove any verbosity from the regular debug output or add any for the alternate debug output.

sdroege: No strong opinion here, seems fine either way for me :)

sdroege (approved):

GuillaumeGomez (approved):

PR #421, opened 2021-04-06
gio: do not take args in application.run()

pbor:
Passing the args is a C-ism of the API since in C the args are
passed to main(), but in rust they are in std::env, so we can
keep the common case simple.
A run_with_args variant is added for the cases where a caller
wants to process the arg vector before running the app.

### DIFF
--- a/gio/src/application.rs
+++ b/gio/src/application.rs
@@ -12,12 +12,19 @@ use std::mem::transmute;
 
 pub trait ApplicationExtManual {
     #[doc(alias = "g_application_run")]
-    fn run(&self, argv: &[String]) -> i32;
+    fn run(&self) -> i32;
+    #[doc(alias = "g_application_run")]
+    fn run_with_args<S: AsRef<str>>(&self, args: &[S]) -> i32;
     fn connect_open<F: Fn(&Self, &[File], &str) + 'static>(&self, f: F) -> SignalHandlerId;
 }
 
 impl<O: IsA<Application>> ApplicationExtManual for O {
-    fn run(&self, argv: &[String]) -> i32 {
+    fn run(&self) -> i32 {
+        self.run_with_args(&std::env::args().collect::<Vec<_>>())
+    }
+
+    fn run_with_args<S: AsRef<str>>(&self, args: &[S]) -> i32 {
+        let argv: Vec<&str> = args.iter().map(|a| a.as_ref()).collect();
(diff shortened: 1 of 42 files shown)


gio/src/application.rs
```diff
@@ -12,12 +12,18 @@ use std::mem::transmute;
 
 pub trait ApplicationExtManual {
     #[doc(alias = "g_application_run")]
-    fn run(&self, argv: &[String]) -> i32;
+    fn run(&self) -> i32;
+    #[doc(alias = "g_application_run")]
+    fn run_with_args(&self, argv: &[String]) -> i32;
```
sdroege, line 17: Ugh, this is annoying :)

```suggestion
    fn run_with_args<S: AsRef<str>>(&self, argv: &[S]) -> i32;
```

please

sdroege (commented):
Makes sense to me otherwise, thanks!

sdroege (approved):

sdroege (approved):

gio/src/application.rs
```diff
@@ -12,12 +12,19 @@ use std::mem::transmute;
 
 pub trait ApplicationExtManual {
     #[doc(alias = "g_application_run")]
-    fn run(&self, argv: &[String]) -> i32;
+    fn run(&self) -> i32;
+    #[doc(alias = "g_application_run")]
+    fn run_with_args<S: AsRef<str>>(&self, args: &[S]) -> i32;
```
GuillaumeGomez, line 17: I'm not a big fan of the naming. What do you think about `run_with_environment_args`?

GuillaumeGomez (commented):
This is super cool! Just one issue with the naming but otherwise I strongly approve this PR!

gio/src/application.rs
sdroege: Instead of plain `run()`? That's a lot of additional typing for the normal case :)

GuillaumeGomez: I made my comment on `run_with_args`. :stuck_out_tongue:

sdroege: It's not using the environment args, but the ones you pass in.

GuillaumeGomez: Oh indeed, just realized that it was using argc and argv, not the env. My bad!

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #422, opened 2021-04-09
gio: Mark `DBusProxy` as `send+sync` and regenerate

ids1024:
https://developer.gnome.org/gio/stable/GDBusProxy.html says the proxy can be used from multiple threads, so I assume this is correct.

### DIFF
--- a/gio/Gir.toml
+++ b/gio/Gir.toml
@@ -421,6 +421,7 @@ generate_builder = true
 
 [[object]]
 name = "Gio.DBusConnection"
+concurrency = "send+sync"
 status = "generate"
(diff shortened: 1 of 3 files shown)


sdroege (approved):
Did you check if any of the other related DBus types are also `Send+Sync`?

ids1024:
Amended the commit to do the same with `DBusConnection`, which has a similar note in it's documentation. Don't know if there are other relevant types that should be `Send` and/or `Sync`.

sdroege (approved):
Ok, let's get this in then and other things can be marked accordingly when needed :)

Issue #69, opened 2019-08-14
Manual implementation of enumerate_*


bilelmoussaoui:
Gio::File lacks enumerate_* that needs some manual bindings I guess. It would be nice to have them back
- [x] enumerate_children (attributes, flags, cancellable)
- [ ]  enumerate_children_async (attributes, flags, io_priority, cancellable, callback, *user_data)
- [ ] enumerate_children_finish (res)

pbor:
working on it

PR #425, opened 2021-04-10
gio: implement enumerate_children_async

pbor:
Fixes https://github.com/gtk-rs/gtk-rs/issues/69

### DIFF
--- a/gio/src/file.rs
+++ b/gio/src/file.rs
@@ -3,6 +3,8 @@
 use crate::Cancellable;
 use crate::File;
 use crate::FileCreateFlags;
+use crate::FileEnumerator;
+use crate::FileQueryInfoFlags;
 use glib::object::IsA;
 use glib::translate::*;
 use std::pin::Pin;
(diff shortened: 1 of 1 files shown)


pbor:
To test I modified one of the examples in this way (I do not think we want to add an example for each function)

https://gist.github.com/pbor/6a8dfc1511b9688d057e0e6c9190a4f9

sdroege (approved):

gio/src/auto/file.rs
```diff
@@ -251,6 +251,26 @@ pub trait FileExt: 'static {
         cancellable: Option<&P>,
     ) -> Result<FileEnumerator, glib::Error>;
 
+    #[doc(alias = "g_file_enumerate_children_async")]
```
sdroege, line 254: Oh, you'd add this to `src/file.rs` into the `FileExtManual` trait.

pbor: oops, will fix

sdroege (approved):

bilelmoussaoui (approved):

PR #426, opened 2021-04-10
pango: Add is_paragraph_start and resolved_dir methods on LayoutLine

### LINKED ISSUE #378: Missing is_paragraph_start and resolved_dir fields on LayoutLine
These are both C bitfields, which do not have a stable representation in rust; accessing them will involve doing something fancy.

This came out of the discussion in #376.

GuillaumeGomez:
Fixes #378.

### DIFF
--- a/pango/src/layout.rs
+++ b/pango/src/layout.rs
@@ -48,6 +48,21 @@ impl LayoutLine {
         unsafe { (*self.to_glib_none().0).length }
     }
 
+    // rustdoc-stripper-ignore-next
+    /// Returns `true` if this is the first line of the paragraph.
+    pub fn is_paragraph_start(&self) -> bool {
+        let field = unsafe { (*self.to_glib_none().0).is_paragraph_start };
+        (field >> 24) != 0 // We only keep the bigger byte to get "is_paragraph".
+    }
+
+    // rustdoc-stripper-ignore-next
+    /// Resolved [`PangoDirection`] of line.
+    pub fn resolved_dir(&self) -> crate::Direction {
+        let field = unsafe { (*self.to_glib_none().0).is_paragraph_start };
+        let field = (field << 8) >> 8; // We remove the bigger byte to only keep "resolved_dir".
+        unsafe { crate::Direction::from_glib(field as _) }
+    }
+
     #[doc(alias = "pango_layout_line_x_to_index")]
     pub fn x_to_index(&self, x_pos: i32) -> HitPosition {
         let mut index = 0;


pango/src/layout.rs
```diff
@@ -48,6 +48,21 @@ impl LayoutLine {
         unsafe { (*self.to_glib_none().0).length }
     }
 
+    // rustdoc-stripper-ignore-next
+    /// Returns `true` if this is the first line of the paragraph.
+    pub fn is_paragraph_start(&self) -> bool {
+        let field = unsafe { (*self.to_glib_none().0).is_paragraph_start };
+        (field >> 24) != 0 // We only keep the bigger byte to get "is_paragraph".
```
sdroege, line 55: It's one bit, not one byte. And AFAIU it's implementation defined where exactly that would be located (might be different per platform, endianness, toolchain and especially MSVC vs. gcc/clang do things differently)

pango/src/layout.rs
```diff
@@ -48,6 +48,21 @@ impl LayoutLine {
...
+    }
+
+    // rustdoc-stripper-ignore-next
+    /// Resolved [`PangoDirection`] of line.
+    pub fn resolved_dir(&self) -> crate::Direction {
+        let field = unsafe { (*self.to_glib_none().0).is_paragraph_start };
+        let field = (field << 8) >> 8; // We remove the bigger byte to only keep "resolved_dir".
```
sdroege, line 62: And this is 3 bits, not 3 bytes and see above :)

GuillaumeGomez:
So, I think this is not doable without making a C library which would expose through functions those bitfields. And I don't want us to require a C compiler (especially on windows!) so I think it'll remain as is for now. Should I close it then?

sdroege:
I think this just needs some research how this is exactly layed out in memory for at least: MSVC 32 bit, MSVC 64 bit (both arm and x86), gcc x86/x86-64/ppc/ppc64/arm/arm64. Based on that we can probably find a pattern that works with as few as possible conditions.

Looking at bindgen might be a good idea for this as it somehow handles bitfields (but generated code differs per target).

pbor:
Shall we file an upstream bug to add those methods?

sdroege:
That would be useful. Accessor methods for all the fields, ideally. There are also bindings that can't access (normal, non-bitfield) fields at all.

GuillaumeGomez:
Closing this PR until then.

sdroege:
@pbor If you create a pango issue/MR, please link it from https://github.com/gtk-rs/gtk-rs/issues/378

Issue #55, opened 2018-12-06
Want Error::steal() -> *mut GError, or a general way to steal() boxed types


federicomenaquintero:
The specific situation is this:  I have a `glib::Error`, and I want to transfer-full it up to C code as a `*mut glib_sys::GError`.  There is no easy way to do it.

@sdroege's way is [this](https://gitlab.gnome.org/GNOME/librsvg/merge_requests/161?diff_id=28325&start_sha=4c954886214ad89b5d2d1d79c6b257d7ed48fa98#note_380159):

```rust
#[no_mangle]
pub fn something_called_from_c(error: *mut *mut GError) -> gboolean {
    match something_that_returns_glib_error() {
        Ok(()) => true.to_glib(),

        Err(e) => {
            if !error.is_null() {
                *error = e.to_glib_none_mut().0;
                mem::forget(e);
            }

            false.to_glib()
        }
}
```

I had a hackier way of doing it, with a cast, but this involves a `g_error_copy()` internally:
```rust
if !error.is_null() {
    *error = e.to_glib_full() as *mut _; // cast the *const GError to *mut GError
```

So, in general: a way to take a `Boxed<T>` and have a `fn steal(self) -> *mut T`.  This would be really useful for returning `GError` up to C, and probably useful for boxed types in general.  Should this be called `steal_ptr` or `steal_glib_ptr` instead?

PR #428, opened 2021-04-10
glib: add Error::into_raw()

pbor:
Fixes https://github.com/gtk-rs/gtk-rs/issues/55

### DIFF
--- a/glib/src/error.rs
+++ b/glib/src/error.rs
@@ -69,7 +69,7 @@ impl Error {
     /// }
     /// ```
     pub fn kind<T: ErrorDomain>(&self) -> Option<T> {
-        if self.0.domain == T::domain().to_glib() {
+        if self.is::<T>() {
             T::from(self.0.code)
         } else {
             None
@@ -215,6 +215,13 @@ impl error::Error for BoolError {}
 mod tests {
     use super::*;
     use std::ffi::CString;
+
+    #[test]
+    fn test_error_kind() {
+        let e = Error::new(crate::FileError::Failed, "Failed");
+        assert_eq!(e.kind::<crate::FileError>(), Some(crate::FileError::Failed));
+        assert_eq!(e.kind::<crate::KeyFileError>(), None);
+    }
 
     #[test]
     fn test_into_raw() {
--- a/glib/src/error.rs
+++ b/glib/src/error.rs
@@ -8,6 +8,7 @@ use std::borrow::Cow;
 use std::error;
 use std::ffi::CStr;
 use std::fmt;
+use std::mem;
 use std::str;
 
 wrapper! {
@@ -81,6 +82,12 @@ impl Error {
             str::from_utf8(bytes)
                 .unwrap_or_else(|err| str::from_utf8(&bytes[..err.valid_up_to()]).unwrap())
         }
+    }
+
+    /// Consumes the `Error` and returns the corresponding `GError` pointer.
+    pub fn into_raw(self) -> *mut ffi::GError {
+        let mut e = mem::ManuallyDrop::new(self);
+        e.to_glib_none_mut().0
     }
 }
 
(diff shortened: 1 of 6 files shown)


glib/src/error.rs
```diff
@@ -82,6 +83,12 @@ impl Error {
                 .unwrap_or_else(|err| str::from_utf8(&bytes[..err.valid_up_to()]).unwrap())
         }
     }
+
+    pub fn into_raw(mut self) -> *mut ffi::GError {
```
GuillaumeGomez, line 87: I think documentation is needed here, considering that it's calling "mem::forget" on `self`. Also: shouldn't this be an `unsafe` function?

glib/src/error.rs
```diff
@@ -82,6 +83,12 @@ impl Error {
...
+        let e = self.to_glib_none_mut().0;
+        mem::forget(self);
```
sdroege, line 89: Take a look at `Value::into_raw`. It's nicer to use `mem::ManuallyDrop` :)

glib/src/error.rs
sdroege: No, doesn't have to be `unsafe`

glib/src/error.rs
```diff
@@ -82,6 +83,12 @@ impl Error {
...
+        e
+    }
```
sdroege, line 91: These can all be updated to use the new function. Maybe I missed some elsewhere

```
gio/src/subclass/io_stream.rs
176:            *err = e.to_glib_none_mut().0;

gio/src/subclass/seekable.rs
213:            *err = e.to_glib_none_mut().0;
248:            *err = e.to_glib_none_mut().0;

gio/src/subclass/output_stream.rs
234:            *err = e.to_glib_none_mut().0;
258:            *err = e.to_glib_none_mut().0;
282:            *err = e.to_glib_none_mut().0;
314:            *err = e.to_glib_none_mut().0;

gio/src/subclass/input_stream.rs
193:            *err = e.to_glib_none_mut().0;
217:            *err = e.to_glib_none_mut().0;
251:            *err = e.to_glib_none_mut().0;
```

pbor:
By the way, value has `#[doc(hidden)]`... any reason for that? should I go ahead and document both?

sdroege:
> By the way, value has `#[doc(hidden)]`... any reason for that? should I go ahead and document both?

Sounds like a good idea

glib/src/error.rs
bilelmoussaoui: We also have some in gtk4-rs, it would be very nice if you could update those as well otherwise an issue at least would be enough

pbor:
Updated, but I guess I am still missing something: I added a commit that calls `g_error_free` on the raw pointer, as I would expect a C caller to do, and it crashes with a double free

glib/src/value.rs
```diff
@@ -1224,6 +1224,15 @@ mod tests {
         assert_eq!(v2.get::<&str>(), Ok(Some("123")));
     }
 
+    #[test]
+    fn test_into_raw() {
+        unsafe {
+            let v = 123.to_value().into_raw();
+            assert_eq!(gobject_ffi::g_type_check_value(&v), ffi::GTRUE);
+            assert_eq!(gobject_ffi::g_value_get_int(&v), 123);
```
sdroege, line 1232: Should probably for completeness also call `gobject_ffi::g_value_unset(&mut v)` here

glib/src/error.rs
```diff
@@ -206,6 +213,25 @@ impl error::Error for BoolError {}
...
+        unsafe {
+            assert_eq!((*e).domain, ffi::g_file_error_quark());
+            assert_eq!((*e).code, ffi::G_FILE_ERROR_FAILED);
+            assert_eq!(
+                CString::from_raw((*e).message),
+                CString::new("Failed").unwrap()
+            );
+
+            // THIS CAUSES double free
+            ffi::g_error_free(e);
```
sdroege, line 231: Code looks correct, unclear why this fails. Will debug now

glib/src/error.rs
```diff
@@ -206,6 +213,25 @@ impl error::Error for BoolError {}
...
+
+    #[test]
+    fn test_into_raw() {
+        let e = Error::new(crate::FileError::Failed, "Failed").into_raw();
+        unsafe {
+            assert_eq!((*e).domain, ffi::g_file_error_quark());
+            assert_eq!((*e).code, ffi::G_FILE_ERROR_FAILED);
+            assert_eq!(
+                CString::from_raw((*e).message),
+                CString::new("Failed").unwrap()
```
sdroege, line 227:
```suggestion
                CStr::from_ptr((*e).message),
                CString::new("Failed").unwrap().as_c_str()
```

sdroege: You were stealing the message from the error here, and that's why it failed freeing below with a double free.

pbor: It is *always* my fault :)

sdroege (approved):

PR #431, opened 2021-04-10
[docs] Some changes to gtk intro

sophie-h:


gtk/src/lib.rs
```diff
@@ -16,71 +16,71 @@
...
-//!                        MessageType::Info,
-//!                        ButtonsType::Ok,
-//!                        "Hello World").run();
+//!     gtk::init().expect("Failed to initialize GTK.");
+//!
+//!     MessageDialogBuilder::new()
+//!         .text("Hello, World!")
+//!         .buttons(ButtonsType::Close)
+//!         .build()
+//!         .run();
```
bilelmoussaoui, line 38: Should use show here

sdroege: Should maybe be something more useful here like the Hello World on https://gtk.org (which is about the same number of lines...)

gtk/src/lib.rs
```diff
@@ -129,9 +129,9 @@
 //! not have easy access to the latest ones.** The higher the version, the fewer
 //! users will have it installed.
 //!
-//! ## Lgpl-docs
+//! ### Lgpl-docs
```
sdroege, line 132:
```suggestion
//! ### LGPL docs
```

sdroege (commented):
Seems good to me otherwise :)

gtk/src/lib.rs
sophie-h: The example would not work with show because there is no main loop. If you think it's worth it, I can restructure the examples. For gtk4 I would just give one example and mostly link to the book for the rest.

sophie-h: I will just drop this example, there is a proper example.

bilelmoussaoui: > The example would not work with show because there is no main loop

We should replace that with something that the user can copy/paste and have something working.

gtk/src/lib.rs
```diff
@@ -16,71 +16,71 @@
...
-//! - [Gtk-rs documentation overview](https://gtk-rs.org/docs-src/)
+//! - [gtk-rs documentation overview](https://gtk-rs.org/docs-src/)
 //!
 //! - [General `GLib` family types and object system overview](../glib/index.html)
 //!
-//! - [GTK+ documentation](https://www.gtk.org/docs/)
+//! - [GTK documentation](https://www.gtk.org/docs/)
 //!
-//! # Hello World
+//! ## “Hello, World!” example program
```
GuillaumeGomez, line 25: There is no need for the "fancy" (don't know what they're called) double-quote, rustdoc is changing them automatically (as you can see [here](https://doc.rust-lang.org/nightly/std/primitive.never.html)).

gtk/src/lib.rs
```diff
@@ -16,71 +16,71 @@
 //!
 //! See also:
 //!
-//! - [Gtk-rs documentation overview](https://gtk-rs.org/docs-src/)
+//! - [gtk-rs documentation overview](https://gtk-rs.org/docs-src/)
```
GuillaumeGomez, line 19: In most places, we use `Gtk-rs` (for example <https://opencollective.com/gtk-rs> and <https://gtk-rs.org>). So for now, I think it'd be better to keep as is until we make a decision about it.

gtk/src/lib.rs
```diff
@@ -3,10 +3,10 @@
 #![allow(clippy::needless_doctest_main)]
 #![allow(clippy::upper_case_acronyms)]
 
-//! # GTK+ 3 bindings
+//! # GTK 3 bindings
 //!
-//! This library contains safe Rust bindings for [GTK+ 3](http://www.gtk.org), a
-//! multi-platform GUI toolkit. It's a part of [Gtk-rs](http://gtk-rs.org/).
+//! This library contains safe Rust bindings for [GTK 3](http://www.gtk.org), a
+//! multi-platform GUI toolkit. It's a part of [gtk-rs](http://gtk-rs.org/).
```
GuillaumeGomez, line 9: Same as below.

gtk/src/lib.rs
```diff
@@ -129,9 +129,9 @@
 //! not have easy access to the latest ones.** The higher the version, the fewer
 //! users will have it installed.
 //!
-//! ## Lgpl-docs
+//! ### Lgpl-docs
 //!
-//! The Gtk-rs crates come with API docs missing because of licensing
+//! The gtk-rs crates come with API docs missing because of licensing
```
GuillaumeGomez, line 134: Same as above.

gtk/src/lib.rs
sdroege: Yeah dropping the example and just keeping the one below makes most sense, I guess (and maybe replacing the example below with the gtk.org one).

gtk/src/lib.rs
sdroege: Should probably be `gtk-rs` or `GTK-rs`, but let's handle that separately I guess...

GuillaumeGomez: Agreed (for context: `gtk-rs` refer to the repository whereas `Gtk-rs` currently somewhat refers to the organization. It's very unclear).

gtk/src/lib.rs
```diff
@@ -20,67 +20,81 @@
...
 //!
-//! # Initialization
+//! ## Initialization
 //!
-//! GTK+ needs to be initialized before use by calling [`init`](fn.init.html) or
-//! [`Application::new`](struct.Application.html#method.new). You only need to
-//! do it once and there is no 'finalize'.
+//! GTK needs to be initialized before use by calling [function@init] or by
+//! creating an [struct@Application]. You only need to do it once and there
+//! is no 'finalize'.
```
sdroege, line 57: > You only need to do it once and there is no 'finalize'.

I don't know what this means (even before your changes) :) @GuillaumeGomez?

gtk/src/lib.rs
```diff
@@ -20,67 +20,81 @@
 //!
 //! - [General `GLib` family types and object system overview](../glib/index.html)
 //!
-//! - [GTK+ documentation](https://www.gtk.org/docs/)
+//! - [GTK documentation](https://www.gtk.org/docs/)
 //!
-//! # Hello World
+//! ## "Hello, World!" example program
```
sdroege, line 25: I think this can be merged with the next section, the code is the same anyway

gtk/src/lib.rs
```diff
@@ -102,20 +112,20 @@
...
 //!
-//! - Any `&str` or `&Path` parameter with an interior null (`\0`) character will
+//! - Any [&str] or [&Path](std::path::Path) parameter with an interior null (`\0`) character will
 //! cause a panic.
 //!
 //! - Some functions will panic if supplied out-of-range integer parameters. All
 //! such cases will be documented individually but they're not yet.
 //!
-//! **A panic in a closure will abort the process.**
+//! **A panic in a closure will abort the main thread.**
```
sdroege, line 121: That's not true. It will actually kill the whole process

gtk/src/lib.rs
GuillaumeGomez: I meant that you don't need to free it. (Yep, can definitely be improved haha)

sdroege: The sentence is gone now as it makes not much sense :) Memory is managed automatically in the bindings everywhere.

sophie-h: Yes! This is rust here! :laughing:

gtk/src/lib.rs
sophie-h: I think you can spawn a thread an do the init there? But I can still revert it.

sophie-h: I was thinking about specifying "in a closure" a bit more. The sentence as it is seems a bit unclear.

sdroege: Any panic in a signal handler closure or any other closure passed to bindings functions will directly kill the whole process. The panic would unwind until an `extern "C"` function, and there it would then call `abort()`.

With current stable Rust it's actually UB (but in almost all cases does exactly that already) but next (or next-next?) release it's guaranteed to do that.

sdroege: But yes, making the "in a closure" part a bit more clear is a good idea :)

sophie-h:
- Removed all the bold formatting because it's not helping imo. I some cases it might make things worse.
- I have removed the statement about missing bindings because I think gtk-rs is quite complete?
- I think it would be good to have an explicit statement about gtk-dev being an external requirement for gtk-rs to work.
- Looks like we can almost copy paste this for gtk4. Just at a prominent mention of the book.

sdroege (approved):

sdroege:
Seems all good to me like this. For `gtk-dev`, no opinion here :)

sophie-h:
> For `gtk-dev`, no opinion here :)

Let's merge this first. Potentially mentioning distro packages etc. is tedious. (I still think it's important though.)

sdroege:
> Let's merge this first. Potentially mentioning distro packages etc. is tedious. (I still think it's important though.)

I agree. Also why is `gtk-dev` not required? That's where the pkg-config files are, or not?

sophie-h:
> Also why is `gtk-dev` not required?

I'm not following. Who said it's not required?

sdroege:
Nobody, I somehow read a "not" in your initial statement that doesn't actually exist. Sorry for the confusion :)

sophie-h:
Can this and #435 be merged?

sdroege (approved):

bilelmoussaoui (approved):

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #432, opened 2021-04-10
gtk3 macros: specify syn features

pbor:
This enables the same syn features in gtk3-macros as it is in glib-macros

### DIFF
--- a/gtk3-macros/Cargo.toml
+++ b/gtk3-macros/Cargo.toml
@@ -23,5 +23,5 @@ heck = "0.3"
 proc-macro-error = "1.0"
 proc-macro2 = "1.0"
 quote = "1.0"
-syn = "1.0"
+syn = { version = "1.0", features = ["full"] }
 proc-macro-crate = "0.1"


gtk3-macros/Cargo.toml
```diff
@@ -23,5 +23,5 @@ heck = "0.3"
 proc-macro-error = "1.0"
 proc-macro2 = "1.0"
 quote = "1.0"
-syn = "1.0"
+syn = { version = "1.0", features = ["full"] }
```
GuillaumeGomez, line 26: Unless needed, I don't think this is a good idea to enable all features by default.

pbor: Yeah, I was about to comment about this on https://github.com/gtk-rs/gtk-rs/issues/41

I have filed this just for consistency among the two internal macro crates, but feel free to drop it.
As written in the ticket I do not think there is much else we can shave.

bilelmoussaoui: iirc, `full` in syn doesn't mean that all the features are enabled.

GuillaumeGomez: I think it's better to drop it then. Feel free to re-open if new information come in!

sdroege: Yes, `full` is not "enable all features". So this together with `default-features = false` probably still makes a difference.

Same thing inside `glib-macros/Cargo.toml`

sdroege: FWIW: `full` only enables all AST types. It doesn't enable macros, lots of trait impls, etc.

GuillaumeGomez: It's still more than what we currently use. So why adding it?

sdroege: My point is that we should do `default-features = false` and then only enable what we need here. If we need `full` then that's ok, that's still less than the default features AFAICS ;)

GuillaumeGomez: I see, completely agree!

sdroege: And yes, it's very confusing with the syn features :)

The same thing with glib-macros btw: `default-features = false` and only enable what we need.

PR #435, opened 2021-04-10
[docs] Some changes to glib intro

sophie-h:


glib/src/lib.rs
```diff
@@ -1,61 +1,53 @@
...
 //!
 //! It is the foundation for higher level libraries with uniform Rusty (safe and
 //! strongly typed) APIs. It avoids exposing GLib-specific data types where
 //! possible and is not meant to provide comprehensive GLib bindings, which
 //! would often amount to duplicating the Rust Standard Library or other utility
 //! crates.
 //!
-//! The library is a work in progress: expect missing functionality and breaking
-//! changes.
+//! The library is a work in progress. Expect breaking changes between releases.
```
bilelmoussaoui, line 15: Is this still true?

sdroege: We have that in all the crates. I think it should be written a bit differently. There will be breaking changes but they should be rather minimal. Not very different to those crates that release a new major version every half year and are considered stable because they're >= 1.0.

sdroege (commented):
Looks good to me apart from that one comment, but that should probably be handled separately for all crates at once.

glib/src/lib.rs
```diff
@@ -64,11 +56,11 @@
 //! reference counted structures so the bindings have to implement corresponding
 //! smart pointers (wrappers), which encapsulate resource management and safety
 //! checks. Such wrappers are defined via the
-//! [`wrapper!`](macro.wrapper!.html) macro, which uses abstractions
-//! defined in the [`wrapper`](wrapper/index.html), [`boxed`](boxed/index.html),
-//! [`shared`](shared/index.html) and [`object`](object/index.html) modules.
+//! [`macro@wrapper`] macro, which uses abstractions
```
GuillaumeGomez, line 59: Just wondering about this: do you want to keep the "macro@" (and "mod@" following) in the doc display?

sophie-h: They aren't visible in the docs. Or what do you mean?

Screenshot from 2021-04-11 13-19-33

GuillaumeGomez: Oh nice. Then nevermind my comment. :)

sdroege (approved):

glib/src/lib.rs
bilelmoussaoui: I guess let's handle that another time as we have the same comment on all the crates :)

bilelmoussaoui (approved):

glib/src/lib.rs
sdroege: Yes, can you create an issue for that?

GuillaumeGomez (approved):

glib/src/lib.rs
sophie-h: I have posted gtk-rs/release#147 for this.

bilelmoussaoui: Thanks Sophie :)

sdroege: https://github.com/gtk-rs/release/issues/147

PR #436, opened 2021-04-10
Improve README files

### LINKED ISSUE #117: Supported Rust version?
The documentation doesn't seem to reflect the minimum Rust version requirements. Or am I missing something?

### LINKED ISSUE #244: Improve README.md
See https://github.com/gtk-rs/gtk4-rs/issues/74 . Especially the minimum supported Rust version should be documented.

CC @bilelmoussaoui

GuillaumeGomez:
Fixes #244.
Fixes #117.

I took example of what we had in gtk4-rs.

### DIFF
--- a/atk/README.md
+++ b/atk/README.md
@@ -1,12 +1,42 @@
 # atk 
 [Project site](https://gtk-rs.org/) | [Online documentation](https://gtk-rs.org/docs-src/)
 
-__Rust__ bindings and wrappers for __Atk__.
+__Rust__ bindings and wrappers for __Atk__, part of [gtk-rs](https://github.com/gtk-rs/gtk-rs).
 
 ## Minimum supported Rust version
 
 Currently, the minimum supported Rust version is `1.48.0`.
 
+## Documentation
+
+ * [Rust API - Stable](https://gtk-rs.org/docs/atk/)
+ * [Rust API - Development](https://gtk-rs.org/gtk-rs/git/docs/atk)
+ * [GTK Installation instructions](https://www.gtk.org/docs/installations/)
+
+## Using
+
+We recommend using [crates from crates.io](https://crates.io/keywords/gtk-rs),
+as [demonstrated here](https://gtk-rs.org/#using).
+
+If you want to track the bleeding edge, use the git dependency instead:
+
+```toml
+[dependencies]
+atk = { git = "https://github.com/gtk-rs/gtk-rs.git", package = "atk" }
+```
+
+Avoid mixing versioned and git crates like this:
+
+```toml
+# This will not compile
+[dependencies]
+atk = "0.13"
+atk = { git = "https://github.com/gtk-rs/gtk-rs.git", package = "atk" }
+
+### See Also
+
+ * [glib](https://crates.io/crates/glib)
+
 ## License
 
-__Atk__ is available under the MIT License, please refer to it.
+__atk__ is available under the MIT License, please refer to it.
--- a/cairo/README.md
+++ b/cairo/README.md
@@ -1,13 +1,6 @@
 # cairo 
 
-Cairo bindings for Rust.
-
-- [Gtk-rs project site](https://gtk-rs.org/)
-
-- [Online documentation](https://gtk-rs.org/docs-src/)
-
-- [Readme](https://github.com/gtk-rs/gtk/blob/master/README.md) in our
-  [main repo](https://github.com/gtk-rs/gtk)
+__Rust__ bindings for Rust and wrappers for __Cairo__.
 
 ![screenshot](https://guillaume-gomez.fr/image/cairo.png)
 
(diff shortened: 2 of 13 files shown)


atk/README.md
```diff
@@ -1,12 +1,41 @@
...
-__Rust__ bindings and wrappers for __Atk__.
+__Rust__ bindings and wrappers for __Atk__, part of [gtk-rs](https://github.com/gtk-rs/gtk-rs).
 
 ## Minimum supported Rust version
 
 Currently, the minimum supported Rust version is `1.48.0`.
 
+## Documentation
+
+ * [Rust API - Stable](https://gtk-rs.org/docs/atk/)
```
bilelmoussaoui, line 12: we do have devel version of the docs as well, maybe it's worth mentioning it here ? especially that building the docs locally require knowing about the `dox,embed-lgpl-docs` trick

cairo/README.md
```diff
@@ -1,20 +1,42 @@
...
+cairo-rs = { git = "https://github.com/gtk-rs/gtk-rs.git", package = "cairo-rs" }
+```
+
+Avoid mixing versioned and git crates like this:
+
+```toml
+# This will not compile
+[dependencies]
+cairo-rs = "0.13"
+cairo-rs = { git = "https://github.com/gtk-rs/gtk-rs.git", package = "cairo-rs" }
```
bilelmoussaoui, line 34: this should probably use a different crate to show up that mixing versions can't work

gtk/README.md
```diff
@@ -83,6 +83,16 @@ When opening a PR please put the changes to the `src/auto` directory in a separa
 You may also wish to run `cargo clippy -- -D warnings` and check that you're clean because
 otherwise you may be surprised when CI fails.
 
+## See Also
+
+ * [atk](https://crates.io/crates/atk)
+ * [cairo](https://crates.io/crates/cairo-rs)
+ * [gdk](https://crates.io/crates/gdk)
+ * [gdk-pixbuf](https://crates.io/crates/gdk-pixbuf)
```
bilelmoussaoui, line 91: could mention gdk-x11 as well, as none of the "See also" can lead you to it

atk/README.md
GuillaumeGomez: We do for gtk-rs too? :o Where is it?

bilelmoussaoui: https://gtk-rs.org/gtk-rs/atk/, they are built by the CI and published at gh-pages

bilelmoussaoui: Actually, we should probably move them to use something similar to the url format of what we have in gtk4-rs with gtk-rs.org/gtk4-rs/git/docs & we can do the stable/docs as well if you want.

sdroege: Yes we should definitely do that :) Ideally before merging this so that it can be correct from the beginning.

atk/README.md
```diff
@@ -1,12 +1,41 @@
...
+atk = { git = "https://github.com/gtk-rs/gtk-rs.git", package = "atk" }
+
+### See Also
+
+ * [glib](https://crates.io/crates/glib)
+
 ## License
 
-__Atk__ is available under the MIT License, please refer to it.
+__atk__ is available under the MIT License, please refer to it.
```
sdroege, line 41: Why is this line rendered red and the one above with the glib link too? Looks like something is unhappy here

atk/README.md
```diff
@@ -1,12 +1,41 @@
...
 ## Minimum supported Rust version
 
 Currently, the minimum supported Rust version is `1.48.0`.
 
+## Documentation
+
+ * [Rust API - Stable](https://gtk-rs.org/docs/atk/)
+ * [GTK Installation instructions](https://www.gtk.org/docs/installations/)
+
+## Using
```
sdroege, line 15: A table with the features and a mention of the minimum version of the C library that is supported would be useful too here :)

atk/README.md
sdroege: And also many lines below.

sdroege (commented):
Otherwise looks like a good improvement to me :) Thanks!

atk/README.md
bilelmoussaoui: I opened a PR for the git docs https://github.com/gtk-rs/gtk-rs/pull/438

atk/README.md
GuillaumeGomez: That's just too annoying to maintain imo.

gtk/README.md
GuillaumeGomez: I only listed dependencies every time.

atk/README.md
sdroege: Why? Not very different from the minimum supported Rust version

GuillaumeGomez: It's very different: the msrv is one information and pretty easy to keep track of. The features are plenty.

atk/README.md
bilelmoussaoui: The git docs are available at https://gtk-rs.org/gtk-rs/git/docs/

atk/README.md
GuillaumeGomez: Dark magic I guess haha.

GuillaumeGomez:
Let's go with this for the moment (I added the links to the "development docs"). Further improvements can be added later on. :)

PR #440, opened 2021-04-11
gtk: optionally re-export gdkx11

### LINKED ISSUE #390: Expose gdk-x11 with a feature flag
It would be nice to be able to include both gdkx11 & gdkwayland  (once #101  was fixed if ever) directly into the re-exported crates from gtk if those features are enabled. 

Of course, the features shouldn't be set by default. 

It's mostly to avoid having to 
```toml
gdk3x11 = {package = "gdkx11", git = "https://github.com/gtk-rs/gtk-rs", optional = true}
gtk3 = {package = "gtk", git = "https://github.com/gtk-rs/gtk-rs", optional = true}
```

and instead it should probably be
```toml
gtk3 = {git = "https://github.com/gtk-rs/gtk-rs", optional = true, package="gtk", features = ["gdkx11"]}
```

Cogitri:
fixes #390

### DIFF
--- a/gtk/src/lib.rs
+++ b/gtk/src/lib.rs
@@ -161,6 +161,8 @@ pub use gdk_pixbuf;
 pub use gio;
 pub use glib;
 pub use pango;
+#[cfg(feature = "gdkx11")]
+pub use gdkx11;
 
 #[doc(hidden)]
 pub use field_offset::*;
--- a/gtk/Cargo.toml
+++ b/gtk/Cargo.toml
@@ -65,6 +65,7 @@ glib = { path = "../glib" }
 gdk = { path = "../gdk" }
 gdk-pixbuf = { path = "../gdk-pixbuf" }
 pango = { path = "../pango" }
+gdkx11 = { path = "../gdkx11", optional = true }
 
 [dev-dependencies]
 gir-format-check = "^0.1"


gtk/src/lib.rs
```diff
@@ -161,6 +161,8 @@ pub use gdk_pixbuf;
 pub use gio;
 pub use glib;
 pub use pango;
+#[cfg(feature = "gdkx11")]
```
bilelmoussaoui, line 164: I have done something like this in gtk4-rs
```
#[cfg(any(all(feature = "x11", target_os = "linux"), feature = "dox"))]
#[cfg_attr(feature = "dox", doc(cfg(feature = "x11", target_os = "linux")))]
```

GuillaumeGomez:
Reexporting used crate seems like a good idea but adding crates just to reexport them doesn't seem like a good idea at all.

bilelmoussaoui:
> Reexporting used crate seems like a good idea but adding crates just to reexport them doesn't seem like a good idea at all.

I agree on that point. Although, my thinking, at least of gtk4-rs, is to provide a way for the user to use gtk4 with simple
```
gtk = { version = "0.1", package = "gtk4", features = ["wayland", "x11"]}
```
and have everything set to use all the gtk APIs without having to figure out the various crates there's. The APIs are used indirectly in the case of gdkx11/gdkwayland, as you can cast gdk's specific types to either wayland/x11 ones.  Although, no strong opinion

Issue #61, opened 2020-11-09
Create tests for glib/src/translate.rs


sdroege:
It would be good to have tests for the different translations from/to C in `translate.rs`, especially the container and pointer translations.

This would be a good newcomer task. I'd start by writing tests for `String` <-> `gchar*`, then `GList` (containing strings) by creating one in `unsafe` code via the C API and checking if the translations give proper results in either direction.

PR #441, opened 2021-04-11
glib: add some unit tests for translate

pbor:
Some low hanging fruits for https://github.com/gtk-rs/gtk-rs/issues/61. Lists are more of a pain :)

### DIFF
--- a/glib/src/translate.rs
+++ b/glib/src/translate.rs
@@ -2253,6 +2253,53 @@ mod tests {
     use std::collections::HashMap;
 
     #[test]
+    fn boolean() {
+        assert_eq!(true.to_glib(), ffi::GTRUE);
+        assert_eq!(false.to_glib(), ffi::GFALSE);
+        assert_eq!(true, unsafe { bool::from_glib(ffi::GTRUE) });
+        assert_eq!(false, unsafe { bool::from_glib(ffi::GFALSE) });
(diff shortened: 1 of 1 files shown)


glib/src/translate.rs
```diff
@@ -2252,6 +2252,49 @@ mod tests {
...
+        let owned = "ABC".to_string();
+        let cstring = CString::new("ABC").unwrap();
+
+        let stash = s.to_glib_none();
+        assert_eq!(unsafe { CStr::from_ptr(stash.0) }, cstring.as_c_str());
+
+        let stash = owned.to_glib_none();
+        assert_eq!(unsafe { CStr::from_ptr(stash.0) }, cstring.as_c_str());
+
+        let ptr: *mut c_char = s.to_glib_full();
```
sdroege, line 2288: You're leaking `ptr` here

pbor: no, I consume it again in

```
        assert_eq!(owned, unsafe { String::from_glib_full(ptr) });
```

(at least as far as I understand)

sdroege: Yes there you consume the second one. You overwrite the first `ptr`

```rust
        let ptr: *mut c_char = s.to_glib_full();
        [...]
        let ptr: *mut c_char = owned.to_glib_full();
```

PR #444, opened 2021-04-11
glib: add Array wrapper

elmarco:
This allows me to write manual bindings for a boxed property returning a GArray

### DIFF
--- a/glib/src/array.rs
+++ b/glib/src/array.rs
@@ -1,12 +1,43 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
+use crate::translate::*;
+use std::fmt;
+
 wrapper! {
-    #[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+    #[derive(PartialEq, Eq, PartialOrd, Ord, Hash)]
     pub struct Array(Shared<ffi::GArray>);
 
     match fn {
         ref => |ptr| ffi::g_array_ref(ptr),
         unref => |ptr| ffi::g_array_unref(ptr),
         get_type => || ffi::g_array_get_type(),
+    }
+}
+
+impl Array {
+    pub fn len(&self) -> usize {
+        unsafe { (*self.to_glib_none().0).len as usize }
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+
+    pub fn data(&self) -> *mut libc::c_void {
+        unsafe { (*self.to_glib_none().0).data as _ }
+    }
+
+    pub fn element_size(&self) -> usize {
+        unsafe { ffi::g_array_get_element_size(self.to_glib_none().0) as usize }
+    }
+}
+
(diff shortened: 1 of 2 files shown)


glib/src/array.rs
```diff
@@ -0,0 +1,12 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+wrapper! {
```
GuillaumeGomez, line 3: That's surprisingly short. Well, the shorter the better I guess.

GuillaumeGomez (approved):

glib/src/array.rs
elmarco: The rest I do manually for now, by casting the array.data and pushing to a vec<>. We may eventually want FromGlibContainer traits implemented? Left for another day :)

glib/src/array.rs
```diff
@@ -0,0 +1,12 @@
...
+    #[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+    pub struct Array(Shared<ffi::GArray>);
+
+    match fn {
+        ref => |ptr| ffi::g_array_ref(ptr),
+        unref => |ptr| ffi::g_array_unref(ptr),
```
sdroege, line 9: That it's reference counted is so annoying...

glib/src/array.rs
sdroege: You might want to add at least some API though. Like at least for getting the length and for getting a raw pointer to the underlying data. That should simplify your manual bindings a bit too

glib/src/array.rs
```diff
@@ -0,0 +1,34 @@
...
+        get_type => || ffi::g_array_get_type(),
+    }
+}
+
+impl Array {
+    pub fn len(&self) -> usize {
```
sdroege, line 18: clippy will also want a `fn is_empty(&self) -> bool` here

glib/src/array.rs
```diff
@@ -0,0 +1,34 @@
...
+        unsafe { (*self.to_glib_none().0).len as usize }
+    }
+
+    pub fn data(&self) -> *mut libc::c_char {
```
sdroege, line 22:
```suggestion
    pub fn data(&self) -> *mut libc::c_void {
```

Seems more correct

bilelmoussaoui: Should be unsafe as well

sdroege: No, just like `as_ptr()` or `into_raw()` is not unsafe. All usage that would cause problems requires unsafe code, getting the pointer itself is not a problem.

glib/src/array.rs
elmarco: ok

glib/src/array.rs
elmarco: *mut or *const is preferrable?

sdroege: `mut` seems fine

Issue #137, opened 2020-11-10
Add bindings for GInitiable / GAsyncInitiable


sdroege:

sdroege:
Theoretically this would also need some integration into `glib::Object::new()` as the returned object for types implementing this interface is incomplete until the `GInitable` API is used... but as GIO and GObject are in separate libraries...

PR #446, opened 2021-04-11
[gtk] Non-panicking manual ApplicationBuilder impl

sophie-h:
Closes gtk-rs/gtk4-rs#287

### DIFF
--- a/gtk/src/application.rs
+++ b/gtk/src/application.rs
@@ -6,10 +6,99 @@ use gio::ApplicationExt;
 use gio::ApplicationFlags;
 use glib::signal::SignalHandlerId;
 use glib::translate::*;
-use glib::ObjectExt;
+use glib::{Cast, IsA, ObjectExt, ToValue};
 
 use std::cell::RefCell;
 use std::rc::Rc;
+
+#[derive(Clone, Default)]
+pub struct ApplicationBuilder {
+    app_menu: Option<gio::MenuModel>,
+    menubar: Option<gio::MenuModel>,
+    register_session: Option<bool>,
+    action_group: Option<gio::ActionGroup>,
+    application_id: Option<String>,
+    flags: Option<gio::ApplicationFlags>,
+    inactivity_timeout: Option<u32>,
+    resource_base_path: Option<String>,
+}
+
+impl ApplicationBuilder {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    pub fn build(self) -> Result<Application, glib::BoolError> {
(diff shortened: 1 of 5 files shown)


GuillaumeGomez:
Shouldn't we make this the implementation by default when generating builders in gir?

GuillaumeGomez (approved):

gtk/src/application.rs
```diff
@@ -6,11 +6,100 @@ use gio::ApplicationExt;
...
+        if let Some(ref flags) = self.flags {
+            properties.push(("flags", flags));
+        }
+        if let Some(ref inactivity_timeout) = self.inactivity_timeout {
+            properties.push(("inactivity-timeout", inactivity_timeout));
+        }
+        if let Some(ref resource_base_path) = self.resource_base_path {
+            properties.push(("resource-base-path", resource_base_path));
+        }
+        let ret = glib::Object::new::<Application>(&properties)?;
```
sdroege, line 57: What's the point of no panicking here? This can only fail if the properties or their types (and not their values!) are invalid, which can't really happen here.

Am I missing something?

sophie-h: The application id could be invalid I guess?

sdroege: That causes a critical warning but `glib::Object::new()` still returns an application. It can't possibly check validity of property values here, only their types and the names of the properties.

Failing construction (because of invalid property values or similar) is only possible for objects implementing `GInitable` (see https://github.com/gtk-rs/gtk-rs/issues/137)

sophie-h: Ah, great! In this case we can remove the manual call to init() from the gtk4 version and just use the "startup: signal there as well? (The signal can panic but that's just how gtk works I guess?)

And `gtk::Application::new` should just be auto generated. The `register_startup_hook` in the manual implementation should not be necessary.

bilelmoussaoui: The reason why we return Result is we are calling `rt::init()?` at `build`

sophie-h: @bilelmoussaoui not in gtk3. It sneakily uses `builder_postprocess = "Application::register_startup_hook(&ret);"`

bilelmoussaoui: Aha, well, I guess we should use something similar in gtk4. I wasn't aware of `builder_postprocess`.

sophie-h: Yeah, took me some time to figure out what's going on there.

I will take a break and suggest a new pair of patches for gtk3 and gtk4.

PR #447, opened 2021-04-11
[gtk] Do not return Result from Application::new()

sophie-h:
Follow-up to #446

### DIFF
--- a/gtk/src/application.rs
+++ b/gtk/src/application.rs
@@ -12,6 +12,19 @@ use std::cell::RefCell;
 use std::rc::Rc;
 
 impl Application {
+    #[doc(alias = "gtk_application_new")]
+    pub fn new(application_id: Option<&str>, flags: ApplicationFlags) -> Application {
+        skip_assert_initialized!();
+        let app: Application = unsafe {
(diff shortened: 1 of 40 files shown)


gtk/src/application.rs
```diff
@@ -30,21 +28,4 @@ impl Application {
...
-    ) -> Result<Application, glib::BoolError> {
-        skip_assert_initialized!();
-        let app: Application = unsafe {
-            Option::from_glib_full(ffi::gtk_application_new(
-                application_id.to_glib_none().0,
-                flags.to_glib(),
-            ))
-            .ok_or_else(|| glib::bool_error!("Failed to create application"))?
-        };
-        Application::register_startup_hook(&app);
```
sdroege, line 47: This is necessary so that GTK is considered initialized or not?

sophie-h: I think it's only necessary for the Builder because we are not using the constructor there.

sdroege: Same comment here as for the gtk4 version :)

sdroege (approved):

GuillaumeGomez (approved):

PR #449, opened 2021-04-11
glib: Remove glib::TypedValue

sdroege:
It's not very useful in practice and there seems to be no usage of it.

If a "typed value" is needed then the proper Rust type can be used as
well and there's no reason to keep it in a GValue wrapper.

----

Extracted from https://github.com/gtk-rs/gtk-rs/pull/31 for easier reviewing and smaller PRs.

### DIFF
--- a/glib/src/value.rs
+++ b/glib/src/value.rs
@@ -6,10 +6,6 @@
 //! isn't known at compile time but once created a `Value` can't change its
 //! type.
 //!
-//! [`TypedValue`](struct.TypedValue.html) has a statically known type and
-//! dereferences to `Value` so it can be used everywhere `Value` references are
-//! accepted.
-//!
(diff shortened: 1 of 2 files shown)


glib/src/value.rs
```diff
@@ -21,15 +17,14 @@
 //!
 //! ```
 //! use glib::prelude::*; // or `use gtk::prelude::*;`
-//! use glib::{TypedValue, Value};
+//! use glib::Value;
 //!
-//! // Value and TypedValue implement From<&i32>, From<&str>
-//! // and From<Option<&str>>. Another option is the `ToValue` trait.
+//! // Value implement From<&i32>, From<&str> // and From<Option<&str>>.
```
pbor, line 22: Spurious `//` in the middle of the line?

sdroege: Thanks!

bilelmoussaoui (approved):

PR #450, opened 2021-04-11
CI: publish release docs in a separate subdir

bilelmoussaoui:
With this the job will start automatically when a release is published and the docs will hopefully be deployed to the stable subdirectory instead of the git one.


GuillaumeGomez:
How do we get the "release" event?

bilelmoussaoui:
> How do we get the "release" event?

By creating a new release (draft releases are not included) https://github.com/gtk-rs/gtk-rs/releases/new; there are various release events such marking a release as a draft, removing one. If I understood the docs correctly, the published type event is when a not-draft release is created

bilelmoussaoui:
> How do we get the "release" event?

In order to ensure we are using the correct event I was thinking of doing a test case on gtk4-rs by releasing 0.0 from current master, see if the docs/book gets published properly and remove the release afterwards

bilelmoussaoui:
This can be merged, i did a test release on gtk4-rs (by creating a github release and removing it afterwards), the jobs were properly triggered and I got https://gtk-rs.org/gtk4-rs/stable/docs/gtk4/ at the end :-)

GuillaumeGomez:
Thanks!

PR #452, opened 2021-04-14
glib: glib-macros: add SharedType and Shared derive macro

mbiggio:
Added a SharedType trait that allows implementing reference-counted
types in terms os std::sync::Arc. In particular, the ffi methods that
increment reference counting can return always the same pointer value.
Also, added a `Shared` derive macro, similar to the existing `GBoxed`
one, that simplifies the creation of Shared objects.

### DIFF
--- a/glib-macros/src/lib.rs
+++ b/glib-macros/src/lib.rs
@@ -3,6 +3,7 @@
 mod clone;
 mod downgrade_derive;
 mod gboxed_derive;
+mod gboxed_shared_derive;
 mod genum_derive;
 mod gerror_domain_derive;
 mod gflags_attribute;
@@ -297,6 +298,34 @@ pub fn gerror_domain_derive(input: TokenStream) -> TokenStream {
 pub fn gboxed_derive(input: TokenStream) -> TokenStream {
     let input = parse_macro_input!(input as DeriveInput);
     let gen = gboxed_derive::impl_gboxed(&input);
+    gen.into()
+}
+
+/// Derive macro for defining a [`SharedType`]`::get_type` function and
+/// the [`glib::Value`] traits.
+///
+/// # Example
+///
+/// ```
+/// use glib::prelude::*;
+/// use glib::subclass::prelude::*;
+///
+/// #[derive(Clone, Debug, PartialEq, Eq)]
+/// struct MySharedInner {
+///   foo: String,
+/// }
+/// #[derive(Clone, Debug, PartialEq, Eq, glib::GSharedBoxed)]
+/// #[gshared_boxed(type_name = "MyShared")]
+/// struct MyShared(std::sync::Arc<MySharedInner>);
+/// ```
+///
+/// [`SharedType`]: subclass/shared/trait.SharedType.html
+/// [`glib::Value`]: value/struct.Value.html
+#[proc_macro_derive(GSharedBoxed, attributes(gshared_boxed))]
+#[proc_macro_error]
+pub fn gshared_boxed_derive(input: TokenStream) -> TokenStream {
+    let input = parse_macro_input!(input as DeriveInput);
+    let gen = gboxed_shared_derive::impl_gshared_boxed(&input);
     gen.into()
 }
 
--- a/glib/src/lib.rs
+++ b/glib/src/lib.rs
@@ -88,6 +88,7 @@ pub use once_cell;
 
 pub use glib_macros::{
     clone, gflags, object_interface, object_subclass, Downgrade, GBoxed, GEnum, GErrorDomain,
+    GSharedBoxed,
 };
 
 pub use self::array::Array;
--- a/glib/src/subclass/shared.rs
+++ b/glib/src/subclass/shared.rs
@@ -25,9 +25,8 @@ where
     type InnerType = T;
 
     unsafe fn ref_(this: *const Self::InnerType) -> *const Self::InnerType {
-        use std::mem::ManuallyDrop;
-        let this_rc = ManuallyDrop::new(std::rc::Rc::from_raw(this));
-        std::rc::Rc::into_raw(ManuallyDrop::take(&mut this_rc.clone()))
+        std::sync::Arc::increment_strong_count(this);
+        this
     }
 
     unsafe fn into_raw(self) -> *const Self::InnerType {
--- a/glib/src/subclass/shared.rs
+++ b/glib/src/subclass/shared.rs
@@ -1,0 +1,230 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+//! Module for registering shared types for Rust types.
+
+use crate::translate::*;
+
+pub unsafe trait RefCounted: Clone + Sized + 'static {
+    /// The inner type
+    type InnerType;
+
+    /// The function used to increment the inner type refcount
+    unsafe fn ref_(this: *const Self::InnerType) -> *const Self::InnerType;
+
+    /// Converts the RefCounted object to a raw pointer to InnerType
+    unsafe fn into_raw(self) -> *const Self::InnerType;
+
+    /// Converts a raw pointer to InnerType to a RefCounted object
+    unsafe fn from_raw(this: *const Self::InnerType) -> Self;
+}
+
+unsafe impl<T> RefCounted for std::sync::Arc<T>
+where
+    T: 'static,
+{
+    type InnerType = T;
+
+    unsafe fn ref_(this: *const Self::InnerType) -> *const Self::InnerType {
+        use std::mem::ManuallyDrop;
+        let this_rc = ManuallyDrop::new(std::rc::Rc::from_raw(this));
+        std::rc::Rc::into_raw(ManuallyDrop::take(&mut this_rc.clone()))
+    }
+
+    unsafe fn into_raw(self) -> *const Self::InnerType {
+        std::sync::Arc::into_raw(self)
+    }
+
+    unsafe fn from_raw(this: *const Self::InnerType) -> Self {
+        std::sync::Arc::from_raw(this)
+    }
+}
+
+unsafe impl<T> RefCounted for std::rc::Rc<T>
+where
+    T: 'static,
+{
+    type InnerType = T;
+
+    unsafe fn ref_(this: *const Self::InnerType) -> *const Self::InnerType {
+        use std::mem::ManuallyDrop;
+        let this_rc = ManuallyDrop::new(std::rc::Rc::from_raw(this));
+        std::rc::Rc::into_raw(ManuallyDrop::take(&mut this_rc.clone()))
+    }
+
+    unsafe fn into_raw(self) -> *const Self::InnerType {
+        std::rc::Rc::into_raw(self)
+    }
+
+    unsafe fn from_raw(this: *const Self::InnerType) -> Self {
+        std::rc::Rc::from_raw(this)
+    }
+}
+
+/// Trait for defining shared types.
+///
+/// Links together the type name with the type itself.
+///
+/// See [`register_shared_type`] for registering an implementation of this trait
+/// with the type system.
+///
+/// [`register_shared_type`]: fn.register_shared_type.html
+pub trait SharedType: Clone + Sized + 'static {
+    /// Shared type name.
+    ///
+    /// This must be unique in the whole process.
+    const NAME: &'static str;
+
+    /// The inner refcounted type
+    type RefCountedType: RefCounted;
+
+    /// Returns the type ID.
+    ///
+    /// This is usually defined via the [`Shared!`] derive macro.
+    ///
+    /// [`Shared!`]: ../../derive.Shared.html
+    fn get_type() -> crate::Type;
+
+    /// Converts the SharedType into its inner RefCountedType
+    fn into_refcounted(self) -> Self::RefCountedType;
+
+    /// Constructs a SharedType from a RefCountedType
+    fn from_refcounted(this: Self::RefCountedType) -> Self;
+}
+
+/// Register a boxed `glib::Type` ID for `T`.
+///
+/// This must be called only once and will panic on a second call.
+///
+/// See [`Shared!`] for defining a function that ensures that
+/// this is only called once and returns the type id.
+///
+/// [`Shared!`]: ../../derive.Shared.html
+pub fn register_shared_type<T: SharedType>() -> crate::Type {
+    unsafe {
+        use std::ffi::CString;
+        unsafe extern "C" fn shared_ref<T: SharedType>(v: ffi::gpointer) -> ffi::gpointer {
+            T::RefCountedType::ref_(v as *const <T::RefCountedType as RefCounted>::InnerType)
+                as ffi::gpointer
+        }
+        unsafe extern "C" fn shared_unref<T: SharedType>(v: ffi::gpointer) {
+            let _ = T::RefCountedType::from_raw(
+                v as *const <T::RefCountedType as RefCounted>::InnerType,
+            );
+        }
+
+        let type_name = CString::new(T::NAME).unwrap();
+        if gobject_ffi::g_type_from_name(type_name.as_ptr()) != gobject_ffi::G_TYPE_INVALID {
+            panic!(
+                "Type {} has already been registered",
+                type_name.to_str().unwrap()
+            );
+        }
+
+        from_glib(gobject_ffi::g_boxed_type_register_static(
+            type_name.as_ptr(),
+            Some(shared_ref::<T>),
+            Some(shared_unref::<T>),
+        ))
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    // We rename the current crate as glib, since the macros in glib-macros
+    // generate the glib namespace through the crate_ident_new utility,
+    // and that returns `glib` (and not `crate`) when called inside the glib crate
+    use crate as glib;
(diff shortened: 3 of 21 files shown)


glib-macros/src/shared_derive.rs
```diff
@@ -0,0 +1,191 @@
...
+    quote! {
+        impl #crate_ident::subclass::shared::SharedType for #name {
+            const NAME: &'static str = #gtype_name;
+
+            type InnerType = #inner_name;
+
+            fn ref_(this: *const Self::InnerType) -> *const Self::InnerType {
+                unsafe {
+                    use std::mem::ManuallyDrop;
+                    let this_arc = ManuallyDrop::new(std::sync::Arc::from_raw(this));
```
sdroege, line 136: There's `Arc::decrement_strong_count` (and increment) that directly works on the pointer. That should be easier :)

glib-macros/src/lib.rs
```diff
@@ -300,6 +301,34 @@ pub fn gboxed_derive(input: TokenStream) -> TokenStream {
...
+/// use glib::prelude::*;
+/// use glib::subclass::prelude::*;
+///
+/// #[derive(Clone, Debug, PartialEq, Eq)]
+/// struct MySharedInner {
+///   foo: String,
+/// }
+/// #[derive(Clone, Debug, PartialEq, Eq, glib::Shared)]
+/// #[shared(type_name = "MyShared")]
+/// struct MyShared(std::sync::Arc<MySharedInner>);
```
sdroege, line 319: I think a better way for this to work would be the require the inner struct to implement a trait, and then to implement this trait on `Arc<T>` and `Rc<T>`. That way it works for both, and users can define their own inner type by implementing a much simpler trait (only needs to provide ref/unref!) while still being able to make use of the derive macro here

glib/src/lib.rs
```diff
@@ -88,6 +88,7 @@ pub use once_cell;
 
 pub use glib_macros::{
     clone, gflags, object_interface, object_subclass, Downgrade, GBoxed, GEnum, GErrorDomain,
+    Shared,
```
sdroege, line 91: `GSharedBoxed` or similar maybe? Consistent with `GBoxed`, `GEnum`, etc. and it makes it clear that this is still a boxed type

glib/src/subclass/shared.rs
```diff
@@ -0,0 +1,125 @@
...
+///
+/// [`register_shared_type`]: fn.register_shared_type.html
+pub trait SharedType: Clone + Sized + 'static {
+    /// Shared type name.
+    ///
+    /// This must be unique in the whole process.
+    const NAME: &'static str;
+
+    /// The inner refcounted type
+    type InnerType;
```
sdroege, line 22: With that separate trait you'd then here require the inner type to implement that, and would move the `ref_` / `unref` functions into that separate trait

sdroege: Also I think for the whole thing to work in a safe way, you need a `fn from_inner(Self::InnerType) -> Self` and `fn into_inner(Self) -> Self::InnerType` or similar

sdroege (commented):
Looks nice apart from that, thanks! :)

glib-macros/src/lib.rs
mbiggio: That makes a lot of sense indeed :) Thanks for the suggestion

glib-macros/src/shared_derive.rs
pbor: That requires rust 1.51 though... do we want to bump the min rust version?

sdroege: I don't mind as many of my projects and/or their dependencies require that anyway, but maybe @GuillaumeGomez has an opinion here? :) tl;rd: 1.51 would allow to do things cleaner here, older versions require workarounds.

GuillaumeGomez: Let upgrade the version then. If that allows to remove dependencies, let's go for it!

pbor: It does not remove dependencies though.
I am personally ok with requiring new rustc if we have good use cases for const generics etc, but bumping to save one line here seems excessive tbh :)

sdroege: I had more places before where I would've liked to use newer APIs but didn't for the same reason :) At some point it makes sense to just update.

glib-macros/src/lib.rs
```diff
@@ -300,6 +301,34 @@ pub fn gboxed_derive(input: TokenStream) -> TokenStream {
...
+/// use glib::prelude::*;
+/// use glib::subclass::prelude::*;
+///
+/// #[derive(Clone, Debug, PartialEq, Eq)]
+/// struct MySharedInner {
+///   foo: String,
+/// }
+/// #[derive(Clone, Debug, PartialEq, Eq, glib::Shared)]
+/// #[shared(type_name = "MyShared")]
+/// struct MyShared(std::sync::Arc::<MySharedInner>);
```
sdroege, line 319:
```suggestion
/// struct MyShared(std::sync::Arc<MySharedInner>);
```

glib/src/subclass/shared.rs
```diff
@@ -0,0 +1,227 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+//! Module for registering shared types for Rust types.
+
+use crate::translate::*;
+
+pub trait RefCounted: Clone + Sized + 'static {
+    /// The inner type
+    type InnerType;
```
sdroege, line 9: Why as a associated type and not as a generic parameter (i.e. `pub trait RefCounted<T: SharedType>: Clone + Sized + 'static`)?

glib/src/subclass/shared.rs
```diff
@@ -0,0 +1,227 @@
...
+
+    /// The function used to increment the inner type refcount
+    fn ref_(this: *const Self::InnerType) -> *const Self::InnerType;
+
+    /// Converts the RefCounted object to a raw pointer to InnerType
+    fn into_ptr(self) -> *const Self::InnerType;
+
+    /// Converts a raw pointer to InnerType to a RefCounted object
+    fn from_ptr(this: *const Self::InnerType) -> Self;
```
sdroege, line 18: `into_raw()` and `from_raw()` might be better names here

glib/src/subclass/shared.rs
```diff
@@ -0,0 +1,227 @@
...
+where
+    T: 'static,
+{
+    type InnerType = T;
+
+    fn ref_(this: *const Self::InnerType) -> *const Self::InnerType {
+        unsafe {
+            use std::mem::ManuallyDrop;
+            let this_rc = ManuallyDrop::new(std::rc::Rc::from_raw(this));
+            std::rc::Rc::into_raw(ManuallyDrop::take(&mut this_rc.clone()))
```
sdroege, line 53: You can use `increment_strong_count` here too

glib/src/subclass/shared.rs
```diff
@@ -0,0 +1,227 @@
...
+            Some(shared_ref::<T>),
+            Some(shared_unref::<T>),
+        ))
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    // Shared macro assumes 'glib' is in scope
```
sdroege, line 137: It really shouldn't do that :) Look at the `#crate_ident` hack used in the other macros to avoid that.

sdroege: While at it, please fix the same for the `GBoxed` derive too

glib/src/subclass/shared.rs
```diff
@@ -0,0 +1,227 @@
...
+    #[test]
+    fn test_value_arc() {
+        assert_ne!(crate::Type::INVALID, MySharedArc::get_type());
+
+        let b = MySharedArc::from_refcounted(std::sync::Arc::new(MySharedInner {
+            foo: String::from("abc"),
+        }));
+        let v = b.to_value();
+        let b2 = v.get_some::<MySharedArc>().unwrap();
+        assert_eq!(b.0.foo, b2.0.foo);
```
sdroege, line 169: You probably also want to check everywhere for `Arc::ptr_eq` to make sure it's indeed the very same value and not just an equivalent copy

sdroege (commented):
Apart from the open comments this seems ready to me now

glib/src/subclass/shared.rs
mbiggio: I looked into that, but it seems [it hardcodes the `crate_ident` to `glib` in case the macro is being invoked inside the `glib` crate](https://github.com/gtk-rs/gtk-rs/blob/master/glib-macros/src/utils.rs#L155-L157). I guess it shoud be hardcoded to `crate` instead - I tried to change it but got a bunch of other errors in the `clone` macro:
```
error[E0433]: failed to resolve: maybe a missing crate `g_debug`?
  --> glib/tests/clone.rs:31:9
   |
31 | /         clone!(@weak ref_state => move || {
32 | |             ref_state.borrow_mut().started = true;
33 | |         })
   | |__________^ maybe a missing crate `g_debug`?

```
so I left it like that for the moment.
Btw [also the `GBoxed` macro does a similar thing currently](https://github.com/gtk-rs/gtk-rs/blob/master/glib/src/subclass/boxed.rs#L70-L71)

glib/src/subclass/shared.rs
mbiggio: For some reason, it seems `increment_strong_count` [is not implemented for `Rc`](https://doc.rust-lang.org/std/rc/struct.Rc.html) in rust 1.51.0

glib/src/subclass/shared.rs
mbiggio: Just because this way it turned out to be easier to implement the `SharedType` trait inside the macro. I'm OK to change it of course if we have a strong preference for the generic parameter :)

glib/src/subclass/shared.rs
sdroege: That should be indeed `crate` in that case. For the `clone!` macro, can you give the `cargo expand` code of that part?

glib/src/subclass/shared.rs
sdroege: Indeed. 1.53: https://github.com/rust-lang/rust/pull/83476

glib/src/subclass/shared.rs
sdroege: How is it more difficult with a generic parameter? I think I would prefer that because it maps directly to how `Arc` and `Rc` are defined, unless it makes things much more difficult.

But in either case, you can make `type SharedType: RefCounted<InnerType=Self>` or `type SharedType: RefCounted<Self>`.

mbiggio: I'm sorry but I'm getting a bit confused by:
> But in either case, you can make type SharedType: RefCounted<InnerType=Self> or type SharedType: RefCounted<Self>.

The `RefCounted` trait is the one that is implemented by `Arc<T>/Rc<T>`, and `InnerType`, whether it is a generic parameter or an associated type of `RefCounted`, should not implement any particular trait, just be the same of `T`. What am I missing?

Actually, now that I think more about that, I believe that the "associated type" version makes probably more sense.
If we implement `RefCounted` as a generic parameter:
```
pub trait RefCounted<T>: Clone + Sized + 'static {
    unsafe fn ref_(this: *const T) -> *const T;
    ...
}
```
we are potentially allowing `Arc/Rc<V>` to implement it in different ways (meaning the parameter `V` not necessarily the same as `T`), while I think it makes sense to enforce those to be the same.
What do you think?

glib/src/subclass/shared.rs
```diff
@@ -0,0 +1,226 @@
...
+///
+/// [`register_shared_type`]: fn.register_shared_type.html
+pub trait SharedType: Clone + Sized + 'static {
+    /// Shared type name.
+    ///
+    /// This must be unique in the whole process.
+    const NAME: &'static str;
+
+    /// The inner refcounted type
+    type RefCountedType: RefCounted;
```
sdroege, line 77:
```suggestion
    type RefCountedType: RefCounted<InnerType = Self>;
```

Is what you would do here in the current situation, or

```suggestion
    type RefCountedType: RefCounted<Self>;
```

if you make it a generic parameter. That would ensure that the types are actually matching up properly.

glib/src/subclass/shared.rs
sdroege: See my comment here at the code in question https://github.com/gtk-rs/gtk-rs/pull/452/files#r613899233

I don't think either of the implementations makes a practical difference if that is done at that place, so keeping it as associated type should also be fine.

sdroege: Oh, now I see my confusion. `type RefCountedType` is going to be something like `Arc<FooInner>`, so the things are actually not matching up in that way and that can't work. The outer and inner types are different.

So yes, this is actually all fine. Sorry!

glib/src/subclass/shared.rs
mbiggio: Sure, here it is.
Looking deeper into it, all the integration tests in the `tests` subfolder of `glib` are failing, because one cannot use the `crate::` prefix in integration tests to refer to the crate being tested.

[cargo_expand.txt](https://github.com/gtk-rs/gtk-rs/files/6317144/cargo_expand.txt)

sdroege: But in the integration tests the crate is available with its own name (`glib`). So I assume our problem is that in the unit tests with need `crate` and in the integration tests we need `glib`, and we can't distinguish which of the two cases it is from the macro?

In that case I would suggest to keep it returning `glib` and add a short explanatory comment into the unit tests about this. Still use the `crate_ident` magic inside the macro so it works outside of glib properly :)

mbiggio: Yes that's the problem. Sure, I will add that comment and double-check that I'm using `crate_ident` inside the macro everywhere :)

sdroege: Great, thanks :)

glib-macros/src/lib.rs
sdroege: Then there's only this one and we can get it merged once @bilelmoussaoui or @GuillaumeGomez reviewed it :)

sdroege:
Let me know when the above discussions are all resolved and this is ready for a final review :)

glib-macros/src/lib.rs
mbiggio: Ops! Sorry, I missed this one, should be fixed now

glib/src/subclass/shared.rs
mbiggio: Comments added, and macro checked - should be ok now

sdroege:
@GuillaumeGomez @bilelmoussaoui What do you think?

sdroege (approved):

glib/src/subclass/shared.rs
```diff
@@ -0,0 +1,229 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+//! Module for registering shared types for Rust types.
+
+use crate::translate::*;
+
+pub trait RefCounted: Clone + Sized + 'static {
```
sdroege, line 7: Hmm this trait should actually be `unsafe`. You can do an implementation that compiles fine but will fail at runtime, e.g. by making `ref_()` simply pas through the pointer.

```suggestion
pub unsafe trait RefCounted: Clone + Sized + 'static {
```

glib/src/subclass/shared.rs
```diff
@@ -0,0 +1,229 @@
...
+    unsafe fn ref_(this: *const Self::InnerType) -> *const Self::InnerType;
+
+    /// Converts the RefCounted object to a raw pointer to InnerType
+    unsafe fn into_raw(self) -> *const Self::InnerType;
+
+    /// Converts a raw pointer to InnerType to a RefCounted object
+    unsafe fn from_raw(this: *const Self::InnerType) -> Self;
+}
+
+impl<T> RefCounted for std::sync::Arc<T>
```
sdroege, line 21:
```suggestion
unsafe impl<T> RefCounted for std::sync::Arc<T>
```

glib/src/subclass/shared.rs
```diff
@@ -0,0 +1,229 @@
...
+    unsafe fn into_raw(self) -> *const Self::InnerType {
+        std::sync::Arc::into_raw(self)
+    }
+
+    unsafe fn from_raw(this: *const Self::InnerType) -> Self {
+        std::sync::Arc::from_raw(this)
+    }
+}
+
+impl<T> RefCounted for std::rc::Rc<T>
```
sdroege, line 41:
```suggestion
unsafe impl<T> RefCounted for std::rc::Rc<T>
```

glib/src/subclass/shared.rs
sdroege: Note: `unsafe fn`s in traits are unsafe to *call*, `unsafe trait`s are unsafe to implement.

sdroege (approved):

pbor:
Should `glib-macros/src/shared_derive.rs` be renamed to `glib-macros/src/gboxed_shared_derive.rs` ?

sdroege:
> Should `glib-macros/src/shared_derive.rs` be renamed to `glib-macros/src/gboxed_shared_derive.rs` ?

That would make sense, good point!

bilelmoussaoui (approved):
Looks good to me, the name still seems a bit weird. Although, I don't think we can do any better if we want to keep the Boxed reference.

PR #454, opened 2021-04-15
Value trait refactoring

### LINKED ISSUE #48: Improve glib::Value usage by merging FromValue/FromValueOptional traits
Currently we have
```rust
pub trait FromValueOptional<'a>: StaticType + Sized {
    unsafe fn from_value_optional(_: &'a Value) -> Option<Self>;
}

pub trait FromValue<'a>: FromValueOptional<'a> {
    unsafe fn from_value(_: &'a Value) -> Self;
}
```

This has the effect that we have

```rust
let v: Option<i32> = value.get::<i32>().unwrap();
```

while `v` can never ever be `None` here. There's `get_some()` which only exists for `FromValue` types.

I would suggest to instead have the following instead of the two traits

```rust
pub trait FromValue<'a> {
    type Target;
    unsafe fn from_value(_: &'a Value) -> Self::Target;
}

impl<'a> FromValue<'a> for i32 {
    type Target = i32;

    unsafe fn from_value(_: &'a Value) -> i32 { ... }
}

impl<'a> FromValue<'a> for String {
    type Target = Option<String>;

    unsafe fn from_value(_: &'a Value) -> Option<String> { ... }
}

etc.
```

This would allow this confusion to disappear, `get::<i32>()` would return `i32` and `get::<String>()` would return `Option<String>`.

`SetValue` and `SetValueOptional` would stay the same as there this problem doesn't really exist.

@GuillaumeGomez @EPashkin Opinions?

sdroege:
The main change is that it for non-nullable types `Value::get()` will
never ever return `None`, replacing the previous confusing
`Value::get_some()`.

In addition it is now possible to `Value::get()` a nullable type
directly without `Option` wrapping. If the value was actually `None`
this would return an `Err`.

There is also a new generic `ValueType` trait that is implemented by all
types that can be used inside a `Value` and that allows
storing/retrieving types of this value into/from a `Value`.

Fixes #48

----

* [x] Tests need updating
* [x] gdk/gtk need manual code to be updated
* [x] atk has a `ValueType` enum that conflicts with the new trait, so gir has to generate code that references it by absolute path instead of simply `use`ing it
* [ ] gir branch needs to be merged: https://github.com/gtk-rs/gir/pull/1100
* [x] Make PR for gtk4-rs: https://github.com/gtk-rs/gtk4-rs/pull/309
* [x] Make MRs for gstreamer-rs / gst-plugins-rs: https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/746


glib-macros/tests/test.rs
```diff
@@ -98,16 +89,19 @@ fn derive_gboxed_nullable() {
 
     let b = MyNullableBoxed(String::from("abc"));
     let v = b.to_value();
-    assert_eq!(&b, v.get::<&MyNullableBoxed>().unwrap().unwrap());
+    assert_eq!(&b, v.get::<Option<&MyNullableBoxed>>().unwrap().unwrap());
+    assert_eq!(b, v.get::<Option<MyNullableBoxed>>().unwrap().unwrap());
 
     let b = Some(MyNullableBoxed(String::from("def")));
-    let v = b.to_value();
+    let v = b.as_ref().to_value();
```
sdroege, line 96: This part is a bit of a change, but IMHO it makes more sense like this and allows more code than before.

Previously you could do ToValue::to_value(&Option<Object>)`, now you can do `ToValue::to_value(&Option<&Object>)`. You can easily get from `&Option<Object>` to `Option<&Object>` via `Option::as_ref()` but the other way around is not possible without cloning.

sdroege: One place where this was previously annoying was that e.g.

```rust
obj.set_property("foo", &Some(obj))
```

would consume `obj`, while now you can simply do `&Some(&obj)`.

Just as before, the following of course also still works

```rust
obj.set_property("foo", &obj)
```

ids1024: Taking `&Option<&Object>` seems a bit weird, and inconsistent with other gtk-rs function signatures, where it would be `Option<&Object>`.

sdroege: I agree but that's simply because `ToValue::to_value()` takes its argument by reference. You'd want it to take a `&Object`, for example, and not an `Object`.

Note that the first `&` in `set_property("foo", &Some(&obj))` is to turn it into a trait object.

ids1024: > to turn it into a trait object

Hm, doesn't look like `ObjectExt::set_property` uses a trait object; the function signature seems to use static dispatch:

```rust
fn set_property<'a, N: Into<&'a str>, V: ToValue>(
    &self,
    property_name: N,
    value: &V,   
) -> Result<(), BoolError>;
```

I suppose for it to work with `Option<&Object>`, `ToValue` would have to take `self` rather than `&self`, and be implemented for `&Object` rather than `Object`. Not sure if that would run into any issues.

sdroege: It would run into the issue that `Object` wouldn't be a `ValueType` :) The traits all would stop to fit together.

sdroege: This could now be

```rust
fn set_property<'a, N: Into<&'a str>, V: ToValue>(
    &self,
    property_name: N,
    value: V,   
) -> Result<(), BoolError>;
```
I.e. not `&V` but `V`. `ToValue` is implemented for `T` and `&T` and `&Option<T>` and `Option<&T>` so there are lots of combinations that can fit into that.

cairo/src/lib.rs
```diff
@@ -60,36 +60,77 @@ macro_rules! gvalue_impl {
...
+                    if ptr.is_null() {
+                        return Err(glib::value::WrongValueTypeOrNoneError::UnexpectedNone);
+                    }
+                }
+
+                Ok(())
+            }
+
+            fn from_value(value: &'a glib::Value) -> Result<Self, Self::Error> {
+                Self::check(value)?;
```
sdroege, line 86: To take this as an example, we could make `from_value()` unsafe again and have it return `Self` and require that `check()` is called before, always.

Opinions?

cairo/src/lib.rs
```diff
@@ -60,36 +60,77 @@ macro_rules! gvalue_impl {
...
-            unsafe fn set_value(v: &mut glib::value::Value, s: &Self) {
-                glib::gobject_ffi::g_value_set_boxed(
-                    v.to_glib_none_mut().0,
-                    s.to_glib_none().0 as glib::ffi::gpointer,
-                );
+        impl<'a> glib::value::FromValue<'a> for $name {
+            type Error = glib::value::WrongValueTypeOrNoneError;
+
+            fn check(value: &'a glib::Value) -> Result<(), Self::Error> {
+                glib::value::WrongValueTypeError::check::<$name>(value)?;
```
sdroege, line 71: `check()` could in theory be moved into the error type. `WrongValueTypeError` can generically be checked, `WrongValueTypeOrNoneError` can also be generically checked (check that `value.data[0]` is all zeroes). That way it would only have to be implemented once and not for each value type.

The only potential problem here is if we ever want to support types that don't use all-zeroes for the `None` value. @fengalin do you see a problem here with your plans around `gst::ClockTime`?

Could such a type use a new error type specifically for that type? Probably not as that would then conflict with the `impl<T: FromValue<Error = WrongValueTypeOrNoneError>> FromValue for Option<T>`?

sdroege: Easy: add a new trait that does only this check, has a default impl for normal `NULL` types and make `WrongValueTypeOrNoneError<T: ThatNewTrait>` so it can call into it.

glib-macros/tests/test.rs
sdroege: With some reorganizing I got the advantages of all the above *and* the old code. Pushing soon.

cairo/src/lib.rs
sdroege: Maybe better. This is not supposed to be called directly anyway and adding the check everywhere introduces quite a bit of boilerplate.

cairo/src/lib.rs
sdroege: Leaving the new trait for that for a later time when actually needed.

sdroege:
This is ready for review now but don't merge before the gir branch is merged.

@fengalin I can't assign you as reviewer, but a review from your would also be very useful here :)

cairo/src/lib.rs
fengalin: I think I preferred the `check` method on the `FromValue` `trait`. It feels more natural to me to read: `MyType::check(value)` than `WrongValueTypeError::check(value)`. As an ambiguity resolver, that would lead to `<MyType as glib::FromValue>::check(value)`.

We could also implement `check_type::<T: FromValue>() -> Result<(), <T as FromValue>::Error>` on `Value`, so that we could use `value.check_type<MyType>()`.

glib/src/value.rs
```diff
@@ -56,31 +51,262 @@ use crate::gstring::GString;
...
+pub trait GetValueError: std::error::Error + Sized {
+    fn check(value: &Value) -> Result<(), Self>;
 }
 
-impl GetError {
-    pub fn new_type_mismatch(actual: Type, requested: Type) -> Self {
-        GetError { actual, requested }
+/// An error returned from the [`get`](struct.Value.html#method.get) function
+/// on a [`Value`](struct.Value.html) for non-optional types an `Option`.
+pub struct WrongValueTypeError<T: StaticType> {
```
fengalin, line 89: Even if it would be a bit longer, I'd prefer `ValueTypeMismatchError` for this.

glib/src/value.rs
```diff
@@ -56,31 +51,262 @@ use crate::gstring::GString;
...
+                )))
+            } else {
+                Ok(())
+            }
+        }
+    }
+}
+
+impl<T: StaticType> WrongValueTypeError<T> {
+    pub fn type_mismatch(actual: Type) -> Self {
```
fengalin, line 122: Since this is a constructor, we'd probably prefer implementing `From`.

glib/src/value.rs
```diff
@@ -56,31 +51,262 @@ use crate::gstring::GString;
...
+
+/// Trait to convert a value to a `Value`.
+pub trait ToValue {
+    /// Convert a value to a `Value`.
+    fn to_value(&self) -> Value;
+
+    /// Returns the type identifer of `self`.
+    ///
+    /// This is the type of the value to be returned by `to_value`.
+    fn to_value_type(&self) -> Type;
```
fengalin, line 273: Suggestion: `value_type` or `type_`, since this is not a conversion.
Also, shouldn't that be an associative function?

fengalin (commented):
Again, really nice improvements! :)

cairo/src/lib.rs
sdroege: >  think I preferred the check method on the FromValue trait. It feels more natural to me to read: MyType::check(value) than WrongValueTypeError::check(value). As an ambiguity resolver, that would lead to <MyType as glib::FromValue>::check(value).

I agree but that would require code duplication in all implementors (or at least I saw no way to get rid of that), and it's almost never directly called anyway but you go via the `Value` API (also note how `FromValue` is **not** part of the prelude).

> We could also implement check_type::<T: FromValue>() -> Result<(), <T as FromValue>::Error> on Value, so that we could use value.check_type<MyType>().

That's already there as `Value::is::<MyType>()` :)

glib/src/value.rs
```diff
@@ -56,31 +51,262 @@ use crate::gstring::GString;
...
+
+impl<'r, T: ValueTypeOptional + StaticType> ValueType for Option<T>
+where
+    Self: for<'s> FromValue<'s, Error = WrongValueTypeOrNoneError<T>>,
+{
+    type Type = T::Type;
+}
+
+/// Trait for errors returned from `Value::get`.
+pub trait GetValueError: std::error::Error + Sized {
```
sdroege, line 83: This trait should probably be `unsafe` because you can probably cause memory unsafety if implemented by simply returning `Ok(())` :)

glib/src/value.rs
sdroege: > Suggestion: value_type or type_, since this is not a conversion.

I've read it as "give the type that to_value returns". But `value_type()` is probably less confusing :)

Will change that, thanks :+1: 

>  Also, shouldn't that be an associative function?

No, then you could as well use `Self::static_type()`. This exists solely for trait objects so that on a `&dyn ToValue` you can check what the resulting type would be. An associated function is not object-safe, so you couldn't use `ToValue` as trait object anymore.

glib/src/value.rs
sdroege: No opinion here, I'll change it unless someone has a good reason not to :)

glib/src/value.rs
sdroege: Makes sense but then the `None` type should also be renamed. Do you have a suggestion for that one? `ValueTypeMismatchOrNoneError`?

cairo/src/lib.rs
fengalin: > I agree but that would require code duplication in all implementors (or at least I saw no way to get rid of that), and it's almost never directly called anyway but you go via the Value API (also note how FromValue is not part of the prelude).

I see. Then I have to say that I don't like `trait GetValueError` :), how about `trait ValueTypeChecker`?

fengalin: > That's already there as Value::is::<MyType>() :)

True, but that's more restrictive than `check_type` which can display helpful info when used with `dbg!` or `println!("{}", ...)`. We can still use `value.check_type().is_ok()` to get the boolean behaviour. Also we could use `?` to early exit, etc.

glib/src/value.rs
fengalin: Yes, that was my idea too :)

glib/src/value.rs
fengalin: I would allow using using things like `Err(type_expr)?` to return `Result<_, ValueTypeMismatchError>`, which I find neat :)

glib/src/value.rs
fengalin: > This exists solely for trait objects so that on a &dyn ToValue you can check what the resulting type would be.

Oh, I see!

glib/src/value.rs
fengalin: Hum, I guess this is also the case for `TryFromGlib`.

cairo/src/lib.rs
sdroege: > I see. Then I have to say that I don't like trait GetValueError :), how about trait ValueTypeChecker?

Works for me

> True, but that's more restrictive than check_type which can display helpful info when used with dbg! or println!("{}", ...). We can still use value.check_type().is_ok() to get the boolean behaviour. Also we could use ? to early exit, etc.

Do you have a use-case for this? I'd prefer to not add more API surface unless it's actually useful for something.

fengalin: > Do you have a use-case for this?

Not exactly, I was mostly thinking about dropping an { actual, requested } message while debugging. Like when an application developer uses an API, gets a type mismatch and dumps the error to get an insight of the type received.

fengalin: Mostly helping during development, not for production.

sdroege: For those cases, a simple `value.get::<ExpectedType>().unwrap()` should do already :) Let's worry about that then when it's a bit clearer what kind of API would be useful

glib/src/value.rs
sdroege: Neat but also a bit confusing :)

glib/src/value.rs
sdroege: Can you create an issue for that so we don't forget?

glib/src/value.rs
fengalin: > Neat but also a bit confusing :)

I thought it was a common pattern. Anyway, I like it.

cairo/src/lib.rs
fengalin: Still, there is my early exit point. With a boolean, you are forced to perform a C style `if value.is() then return ...` when you could just use Rust style `value.check_type()?`  ;)

glib/src/value.rs
fengalin: https://github.com/gtk-rs/gtk-rs/issues/462

glib/src/value.rs
sdroege: Returning a non-error type (`glib::Type`) via `Err(_)` just seems strange to me :)

cairo/src/lib.rs
sdroege: Can you give a somewhat realistic code example where you would do that but would not use `Value::get`?

glib/src/value.rs
fengalin: Well you can see it like this: the error type is `ValueTypeMismatchError` which requires the mismatched type to be constructed, so I return said type and let `rustc` take care about the boilerplates.

cairo/src/lib.rs
fengalin: I guess that would be the same use cases as for `is`. But I don't have an example for any off my mind.

sdroege: I don't have one for `is` either, but that existed before :)

fengalin: :) the more I use Rust, the more dislike functions returning a `boolean`. I find them too restrictive.

sdroege: I only ever needed these things in combination with `get()` so...

.gitmodules
```diff
@@ -1,6 +1,6 @@
 [submodule "gir"]
 	path = gir
-	url = https://github.com/gtk-rs/gir
+	url = https://github.com/sdroege/gir
```
GuillaumeGomez, line 3: We'll need to not commit that too. :) (you know that you have an option in `generator.py` to run another `gir` right? ;) )

sdroege: Yes but then the CI fails, and if a single CI job fails it's cancelling all the others so I wouldn't know if all are actually passing :)

glib/src/gstring.rs
```diff
@@ -505,32 +503,65 @@ impl StaticType for GString {
...
+    fn to_value(&self) -> crate::Value {
+        <&str>::to_value(&self.as_str())
+    }
+
+    fn to_value_type(&self) -> Type {
+        String::static_type()
+    }
+}
+
+impl crate::value::ToValueOptional for GString {
```
GuillaumeGomez, line 528: Why not the equivalent of `Into` and `TryInto` here? Like `ToValue` and `TryToValue`? That would seem more logical no?

sdroege: It's not `Into` because it's taking a reference instead of consuming the value, and it's not `Try` because it can't fail but instead it allows turning an `Option<T>` (that is: an optional) into a value.

A previous version of `FromValue` could've been a `TryFromValue` (it previously returned a `Result`), but the current version can't fail anymore either (and is `unsafe` because of that and the checking is done elsewhere, because otherwise there would be quite some code duplication for the checking).

glib/src/value.rs
bilelmoussaoui: I was just going to say that the `ValueTypeMismatchOrNoneError` name is probably not great, but then I have no better ideas and I it's probably better be explicit about the the difference between that & `WrongValueTypeError`

bilelmoussaoui:
Overall the changes looks good to me, other than a little comment on the gir PR :) really nice to see the get_some removed as it was very confusing.

sdroege:
Thanks, I'll later update the names then and then prepare the other PRs.

sdroege:
Updated, now the other crates.

sdroege:
> Updated, now the other crates.

gtk4-rs done, gstreamer-rs mostly done but that showed some problems that I'll have to solve first :)

sdroege:
The gstreamer-rs MR has some open issues that will need improvements here probably: https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/746

glib/src/value.rs
```diff
@@ -56,31 +51,264 @@ use crate::gstring::GString;
...
+
+impl<'r, T: ValueTypeOptional + StaticType> ValueType for Option<T>
+where
+    Self: for<'s> FromValue<'s, Error = ValueTypeMismatchOrNoneError<T>>,
+{
+    type Type = T::Type;
+}
+
+/// Trait for errors returned from `Value::get`.
+pub unsafe trait ValueTypeChecker: std::error::Error + Sized {
```
sdroege, line 83: Will become `pub unsafe trait ValueTypeChecker: std::error::Error + Sized + Send + Sync + 'static` {` so that this can be used painlessly

glib/src/value.rs
```diff
@@ -56,31 +51,264 @@ use crate::gstring::GString;
...
+    fn check(value: &Value) -> Result<(), Self>;
```
sdroege, line 84: Will become `fn check(value: &Value, type_: Type) -> Result<(), Self>` for the same reason

glib/src/value.rs
```diff
@@ -56,31 +51,264 @@ use crate::gstring::GString;
...
+}
+
+/// An error returned from the [`get`](struct.Value.html#method.get) function
+/// on a [`Value`](struct.Value.html) for non-optional types an `Option`.
+pub struct ValueTypeMismatchError<T: StaticType> {
```
sdroege, line 89: Will get the generic type removed and instead as part of the struct. That also simplifies the implementation here

glib/src/value.rs
sdroege: Or maybe rather `fn check<T: StaticType>(value: &Value) -> Result<(), Self>`, not sure...

glib/src/value.rs
```diff
@@ -614,253 +767,195 @@ pub trait ToSendValue: Send + ToValue {
...
     }
-}
 
-impl SetValue for str {
-    unsafe fn set_value(value: &mut Value, this: &Self) {
-        gobject_ffi::g_value_take_string(value.to_glib_none_mut().0, this.to_glib_full())
+    fn value_type(&self) -> Type {
+        String::static_type()
     }
 }
```
sdroege, line 800: There can also be a `impl ValueType for &'static str { ... }` now, which can be useful.

sdroege: And should maybe also add impls for `Cow<str>`

glib/src/value.rs
fengalin: I think `fn check<T: StaticType>(value: &Value) -> Result<(), Self>` is better if that causes no issues.

sdroege: Thanks, I couldn't decide between the two :) Neither causes problems from what I can see so far.

sdroege: That all doesn't fit very well either, no matter which of the two. Hm!

sdroege: Basically the error types need to be generic over the type or otherwise they can't actually implement any checking whether the value is compatible with it.

glib/src/value.rs
sdroege: One problem here is that `ValueTypeMismatch<T>` is only `'static` if `T: 'static` but that's not really what we want here (e.g. `&str`). It doesn't matter that the `T` is only used in a `PhantomData` here.

fengalin: Can you give an example of the pain if it wasn't static?

sdroege: For using it as `std::error::Error::source()` return value (the `gst::Structure` getter error)

sdroege: Generally, non-`'static` errors are not great for use in combination with other code.

sdroege: Which is of course also a problem for the structure getter error because it borrows the field name string

sdroege: Ok, so my plan now is

```rust
pub trait ValueTypeChecker {
  type Error = ...;
  fn check(&Value) -> Result<(), Self::Error>;
}

pub struct ValueTypeMismatchError { actual: Type, requested: Type }

pub struct GenericValueTypeChecker<T>;

impl<T: Type> ValueTypeChecker for GenericValueTypeChecker<T> {
  type Error = ValueTypeMismatchError;
  fn check...
}

pub trait FromValue {
  type Checker: ValueTypeChecker;
  ...
}

etc.
```

That way the error types and everything are `'static` etc.

Also, for GStreamer's structure getter: the field names are actually `GQuark`s so can be just `&'static str`.

glib/src/value.rs
```diff
@@ -56,31 +51,223 @@ use crate::gstring::GString;
...
+///
+/// Usually this would not be used directly but from the [`get`](struct.Value.html#method.get)
+/// function on a [`Value`](struct.Value.html)
+pub unsafe trait FromValue<'a>: Sized {
+    /// Value type checker.
+    type Checker: ValueTypeChecker<Error = Self::Error>;
+    /// Error returned from `Checker`.
+    // FIXME: This has to be a separate associated type because we can't do
+    // `where T: FromValue<Checker::Error = ValueTypeMismatchOrNoneError>`
+    type Error: std::error::Error + Send + Sized + 'static;
```
sdroege, line 201: This sucks a bit. @fengalin do you know a solution so the error type doesn't have to be effectively duplicated here?

fengalin: I think you can define a 2 steps constraint, see lines 32 & 33 in https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=9808b1508c31f01c1f525c9ce3a27908

sdroege: That works but you can't do

```rust
pub trait ValueTypeOptional:
    ValueType + ToValueOptional + FromValue<'static, Checker::Error = ValueTypeMismatchOrNoneError> + StaticType
{
}

unsafe impl<'a, T: FromValue<'a, Checker::Error = ValueTypeMismatchOrNoneError> + StaticType> FromValue<'a>
    for Option<T>
{
}
```

The `::` give parser errors in both positions there

fengalin: My solution addresses the 2d case AFAICT.

sdroege: Not really. Any `<` in those positions causes a parser error too :)

For example

```rust
unsafe impl<'a, T: FromValue<'a, <Checker as ValueTypeChecker>::Error = ValueTypeMismatchOrNoneError> + StaticType> FromValue<'a>
    for Option<T>
{
}
```

Gives

```
error: expected one of `(`, `,`, `::`, `:`, `<`, or `>`, found `=`
   --> glib/src/value.rs:211:71
    |
211 | unsafe impl<'a, T: FromValue<'a, <Checker as ValueTypeChecker>::Error = ValueTypeMismatchOrNoneError> + StaticType> FromValue<'a>
    |                                                                 ----- ^ expected one of `(`, `,`, `::`, `:`, `<`, or `>`
    |                                                                 |
    |                                                                 maybe try to close unmatched angle bracket
```

I'm not sure if this is intentional. In `where` clauses this works btw, so that would be an option for the `impl` blocks but doesn't work for super-traits.

sdroege: >  but doesn't work for super-traits.
where clauses for super-traits are not stable, that's why.

fengalin: That's why I used:

```rust
impl<T, C> FromValue for Option<T>
where
    T: FromValue<Checker = C>,
    C: TypeChecker<Error = GenericError>,
{
    type Checker = GenericChecker;
}
```

With your example, I guess, it would become:

```rust
unsafe impl<'a, T, C> FromValue<'a> for Option<T>
where
    T: FromValue<'a, Checker = C> + StaticType,
    C: ValueTypeChecker<Error = ValueTypeMismatchOrNoneError>,
{
}
```

But that doesn't address your first use case, maybe a solution would to use additional traits and blanket implementations like in https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=0d01fe15092b6e7557a1c0aa4e645712

I have to go for an hour or so.

sdroege: Hmm... yes that could potentially work. I'll give it a try.

sdroege: Nope, doesn't work. Your version doesn't work because `C` is not part of the trait or implementor type. And the following doesn't work because the constraint is in a `where` clause and not in the `impl` bounds:

```rust
pub trait OptionalValueTypeChecker: ValueTypeChecker {}
impl<T: ValueTypeChecker<Error = ValueTypeMismatchOrNoneError>> OptionalValueTypeChecker for T {}

/// Blanket implementation for all optional types.
unsafe impl<'a, T: FromValue<'a> + StaticType> FromValue<'a>
    for Option<T>
    where <T as FromValue<'a>>::Checker: OptionalValueTypeChecker,
{
    type Checker = GenericValueTypeChecker<T>;
    type Error = ValueTypeMismatchError;

    unsafe fn from_value(value: &'a Value) -> Self {
        if let Err(ValueTypeMismatchOrNoneError::UnexpectedNone) = T::Checker::check(value) {
            None
        } else {
            Some(T::from_value(value))
        }
    }
}
```

Gives

```
222 |         if let Err(ValueTypeMismatchOrNoneError::UnexpectedNone) = T::Checker::check(value) {
    |                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^    ------------------------ this expression has type `std::result::Result<(), <T as FromValue<'a>>::Error>`
    |                    |
    |                    expected associated type, found enum `ValueTypeMismatchOrNoneError`
    |
    = note: expected associated type `<T as FromValue<'a>>::Error`
                          found enum `ValueTypeMismatchOrNoneError`
```

I guess let's keep it like this for now and we can figure out a nicer way after it's merged.

fengalin: > Your version doesn't work because C is not part of the trait or implementor type.

Maybe the use case is different, but it runs in my playground. `C` can be used thanks to `impl<'a, T, C>`.
I wanted to try it on your branch, but it seems you found a solution.

bilelmoussaoui (approved):
After looking at the changes on gtk-rs/gtk4-rs along with the examples/tutorials in gstreamer-rs, it looks all good to me. If there's something that still can be improved, it can be done later so we don't block the rest of the changes with this PR & rebasing all the various projects isn't a fun thing to do neither.

Thanks a lot @sdroege for your work on this

glib/src/value.rs
sdroege: Something to look into at a later time :) Maybe you can take a look after it's merged? Worst case we can remove some more code from all the crates afterwards.

sdroege:
I'll merge this tomorrow morning unless there are any concerns in the
meantime, or possible improvements :)

fengalin:
See my proposal for getting rid of `FromValue::Error`: https://github.com/fengalin/gtk-rs/commit/8ebf2ac01186e3974c6d0260a84142849a8ba36e

sdroege:
Interesting, that looks less bad than I expected :) I'll merge that in a bit, update the generator and test things again. Thanks!

sdroege:
Ok, time to merge everything.

PR #455, opened 2021-04-16
Get removal round 2

fengalin:
## Note to applications developers

Applications developers should use [`fix-getters-calls` 0.3.0](https://crates.io/crates/fix-getters-calls/0.3.0) to ease migration of their applications. If you also wish to change your get functions definition to comply with the standards implemented here, use [`fix-getters-def` 0.3.0](https://crates.io/crates/fix-getters-def/0.3.0).

## Object

This PR addresses two concerns identified during https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/639:

- `enum` getters generated automatically required specific renames in `Gir.toml` because they were consuming `self`.
- `get` elision was not applied in the `subclass` infrastructure.

This second round addresses other important concerns. See the details in this [gir PR](https://github.com/gtk-rs/gir/pull/1102).


gtk/src/subclass/container.rs
```diff
@@ -32,7 +32,7 @@ pub trait ContainerImpl: ContainerImplExt + WidgetImpl {
         self.parent_child_type(container)
     }
 
-    fn get_path_for_child(&self, container: &Self::Type, widget: &Widget) -> WidgetPath {
+    fn path_for_child(&self, container: &Self::Type, widget: &Widget) -> WidgetPath {
         self.parent_get_path_for_child(container, widget)
```
bilelmoussaoui, line 36: I'm not sure we want to keep a `parent_get_X` here. It's confusing that everything is renamed except those

gtk/src/builder.rs
```diff
@@ -22,7 +22,7 @@ impl Builder {
 
 pub trait BuilderExtManual: 'static {
     #[doc(alias = "gtk_builder_get_object")]
-    fn get_object<T: IsA<Object>>(&self, name: &str) -> Option<T>;
+    fn object<T: IsA<Object>>(&self, name: &str) -> Option<T>;
```
bilelmoussaoui, line 25: Do we want to rename this one? not very sure

gio/src/subclass/action_group.rs
```diff
@@ -115,7 +111,7 @@ pub trait ActionGroupImplExt: ObjectSubclass {
         action_name: &str,
         value: &Variant,
     );
-    fn parent_get_action_enabled(&self, action_group: &Self::Type, action_name: &str) -> bool;
+    fn parent_action_is_enabled(&self, action_group: &Self::Type, action_name: &str) -> bool;
```
bilelmoussaoui, line 114: Interesting, some of the parent_get_x were renamed, but not all of them

gtk/src/subclass/container.rs
fengalin: It's leftover, will update shortly and relaunch a global search.

gtk/src/builder.rs
fengalin: I don't know.

@GuillaumeGomez @sdroege ?

gio/src/subclass/action_group.rs
fengalin: Yeah, those were manual changes, will check again.

gtk/src/builder.rs
sdroege: I don't know. Is it common for getters-by-name to omit the `get_` prefix?

fengalin: > Is it common for getters-by-name to omit the get_ prefix?

I guess, that depends, that's the reason why I stuck by the "conservative" approach initially. Here is another example:

https://github.com/fengalin/gtk-rs/blob/f5b0dd01859e9b43139b4329cd915adb20b59a81/gio/src/auto/action_group.rs#L49

fengalin: We can also consider usage: 

```
let header_bar = builder.get_object("header-bar").unwrap();
```

or

```
let header_bar = builder.object("header-bar").unwrap();
```

sdroege: Do we have something similar in some other well-written Rust library or even better in std that we could take as reference?

glib/src/subclass/object.rs
```diff
@@ -36,7 +36,7 @@ pub trait ObjectImpl: ObjectSubclass + ObjectImplExt {
     ///
     /// This is called whenever the property value of the specific subclass with the
     /// given index should be returned.
-    fn get_property(&self, _obj: &Self::Type, _id: usize, _pspec: &ParamSpec) -> Value {
+    fn property(&self, _obj: &Self::Type, _id: usize, _pspec: &ParamSpec) -> Value {
```
sdroege, line 39: This would be another such case. `fn property()` looks a bit strange to me at least, but `fn get_property()` seems to be nicer.

Not sure what the pattern here would be... `get_` prefix included if more than the `&self` parameter, and for subclass code more than the `&self, &Self::Type` parameters?

gtk/src/builder.rs
fengalin: Searching for `get` in `std` doesn't raise a lot: https://doc.rust-lang.org/std/?search=get.

The closet to the name thing would be the [`HashMap::get`](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.get).

glib/src/subclass/object.rs
fengalin: Well, there are many cases for which I think it's still preferable to elide `get`, like for the `ActionGroup::action_state` example above, or the many examples of `child_by_name`, `child_by_index`, `preferred_width`, etc... Like I mentioned in the `gir` PR, I believe now that *not* removing `get` is more the exception than the rule. I came to that conclusion because starting from round 1, I proceeded to remove the remaining `get` on a case by case basis and I found myself removing most of them. I browsed the result of the `get` removal from this round several time (and lost monthes of sight in the process :)). Now, I think deciding for the few for which `get` should be restored will be very subjective.

So what we need to choose, and by *we*, I mean you three :) since I already made my choice in these PR, is:

* Do we revert to round 1 and only apply subclass changes.
* Someone also browses through the changes in function definitions (the regen and fix-getters-def commits) and picks the elisions that should be reverted.
* We revert `get_object` & `get_property` and keep it like this until something else shows up.

fengalin: Forgot to mention that I'd preferred the 2d option :)

gtk/src/builder.rs
fengalin: I found this from an often praised Rust developer: https://docs.rs/grep-matcher/0.1.4/grep_matcher/trait.Matcher.html#method.capture_index

As a counter example, in this `struct`, `get` is used in some places and note in others: https://docs.rs/wgpu/0.7.1/wgpu/struct.Adapter.html

I found a similar situation in Druid. In a single `struct`, you can find functions retrieving something [with a `get` prefix](https://docs.rs/druid/0.7.0/druid/struct.Clipboard.html#method.get_format) and others [without `get`](https://docs.rs/druid/0.7.0/druid/struct.Clipboard.html#method.preferred_format).

I also searched in `iced`, but its model doesn't favour getters in the first place.

I believe this is really subjective.

fengalin: Continuing my quest: `bevy` seems to be using [`get` when provided with additional arguments](https://docs.rs/bevy/0.5.0/bevy/asset/struct.SourceInfo.html).

glib/src/subclass/object.rs
fengalin: I'd like to move forward with this. Since yesterday, I proceeded to check some crates for usage (see the end of [this discussion](https://github.com/gtk-rs/gtk-rs/pull/455#pullrequestreview-638212421)), but there doesn't seem to be a defined convention. If someone find courage to browse at least some of the changes in the regen and fix-getters-def commits, that would be helpful and we could discuss, otherwise I would adopt option 1.

sdroege: The cautious change would be 1), but IMHO this can also land as is. The difference in the remaining functions that are renamed here (outside `src/subclass`) is really small and I'd probably tell you a different opinion if you ask me on two different days.

We can do either depending on what others think. In the worst case we revert some functions again later, but IMHO in sum this is a positive change.

@bilelmoussaoui @GuillaumeGomez ?

fengalin: > The difference in the remaining functions that are renamed here (outside src/subclass) is really small

I'll update the commit stack later today so that subclass related changes are separated from the others, that would ease reviewing and identification of potential candidates for `get` restore.

.gitmodules
```diff
@@ -1,6 +1,6 @@
 [submodule "gir"]
 	path = gir
-	url = https://github.com/gtk-rs/gir
+	url = https://github.com/fengalin/gir
```
GuillaumeGomez, line 3: We'll need to be careful not to merge this change. ;)

fengalin: Hehe :)

gtk/src/builder.rs
bilelmoussaoui: Hmm, it's really difficult to take a decision here, but I guess let's remove the get prefix here anyway :) at worst, if someone complains, we can re-consider the rename here.

glib/src/subclass/object.rs
bilelmoussaoui: I think it's fine to rename the get_property to just property, it's a bit confusing at first but I think that comes from the habit of using `get_property` here.

sdroege: Ok, so it seems everybody agrees.

fengalin: @bilelmoussaoui do you still want to keep `get` for `get_object` though?

bilelmoussaoui: @fengalin nop, let's remove it :)

fengalin:
I reworked the commit stack to separate the different levels of changes so as to reduce the amount of things to review. The commits for relaxed get fn identification which could be source of concerns, are now:

- https://github.com/fengalin/gtk-rs/commit/c61f7b9ade14be126a6bfb91d0bb8b3b5124d720: regen after gir update with relax get fn identification.
- https://github.com/fengalin/gtk-rs/commit/8562dde4e4b44001988b0247b66ed9f6c4cf3004: migration tool applying same rules on manual code definition.

@sdroege @bilelmoussaoui @GuillaumeGomez please take a look and tell me anything strikes you.

@GuillaumeGomez : please take a look at [glib/src/auto/functions.rs](https://github.com/fengalin/gtk-rs/commit/c61f7b9ade14be126a6bfb91d0bb8b3b5124d720#diff-cf3e946adaf5f81ef984a00706c59e6fe900fbd6a7e789f92582977fc5af9789).

GuillaumeGomez (approved):

GuillaumeGomez:
Nothing caught my eyes, so I guess it's all good! :)

sdroege:
I looked at this and the gstreamer-rs MR again and it seems all fine to me. Some functions feel a bit strange but that's probably just a matter of getting used to it.

sdroege (approved):

Issue #430, opened 2021-04-10
ParamSpec trait should be unsafe


sdroege:
See https://github.com/gtk-rs/gtk-rs/pull/400#discussion_r605392775

sdroege:
Maybe also other traits, like the cairo pattern one.

pbor:
I do not see any pattern trait in cairo...

sdroege:
Indeed, that's solved differently nowadays. `gdk::FromEvent` is another one though. I think that's all :)

PR #457, opened 2021-04-17
glib: mark ParamSpecType trait as unsafe

pbor:
Implementing it requires ensuring that downcast() works.
Fixes https://github.com/gtk-rs/gtk-rs/issues/430


sdroege (commented):
Looks good but can you also add `gdk::FromEvent`?

sdroege (approved):

PR #458, opened 2021-04-17
glib: replace struct Continue with an enum

pbor:
Having `Continue(true)` and `Continue(false)` does not look very idiomatic. Using an `enum` is more natural.

However using an enum also means that we need to come up with a name for it... This patch proposes `Control`, but I am open to any better ideas. For now I did not add `Control` to the prelude.

### DIFF
--- a/examples/cairo_threads/main.rs
+++ b/examples/cairo_threads/main.rs
@@ -128,7 +128,7 @@ fn build_ui(application: &gtk::Application) {
 
         area.queue_draw_area(origins[thread_num].0, origins[thread_num].1, WIDTH, HEIGHT);
 
-        Continue(true)
+        glib::source::Control::Continue
     });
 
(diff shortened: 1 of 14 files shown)


pbor:
glib signal has another similar struct: `struct Inhibit`. I think we should use an enum for it as well, however there are various things to consider:

1. Do we really need a generic Propagate/Inhibit mechanism for all signals? `glib` itself only defines `GDK_EVENT_PROPAGATE`  and `STOP` at the Gdk level and only uses it for events.
2. If we decide to keep it, we have one more naming problem... `signal::Control` and require the caller to disambiguate with namespaces? Or we go with `SourceControl` and `SignalControl`? Or we use something different entirely?

GuillaumeGomez:
Needs to be (partially) re-opened on gtk-rs-core.

PR #459, opened 2021-04-17
glib: Remove un-generatable type UriParamsIter

MarijnS95:
This is a boxed record that cannot be generated due to missing memory management functions.  That was previously silently permitted (and logged somewhere along the rest of the warnings) but is now proactively disallowed thanks to https://github.com/gtk-rs/gir/pull/1103.  Remove this type from the `generate` list since it has no effect anyway.


bilelmoussaoui (approved):
Oups, that was added by me. Thanks!

PR #460, opened 2021-04-18
WIP: Add parent_fn macro

pbor:
This macro tries to cut down the boiler plate required to chain up methods when subclassing an object.

I am opening this PR to gauge whether this is something there is interest in. I converted just a bunch of callers in `widget` to show how it would work.

Also for now I put the macro in `lib.rs` just because it was the easiest place: clearly a better place is needed, but I did not want to fight macro resolution issues for now.


GuillaumeGomez:

PR #461, opened 2021-04-18
Return Options for new UnixMountEntry

sophie-h:


sdroege (approved):

gio/src/unix_mount_entry.rs
```diff
@@ -9,30 +9,28 @@ use std::mem;
...
         unsafe {
             let mut time_read = mem::MaybeUninit::uninit();
-            let ret = from_glib_full(ffi::g_unix_mount_at(
+            let ret: Option<UnixMountEntry> = from_glib_none(ffi::g_unix_mount_at(
                 mount_path.as_ref().to_glib_none().0,
                 time_read.as_mut_ptr(),
             ));
-            let time_read = time_read.assume_init();
-            (ret, time_read)
+            ret.map(|x| (x, time_read.assume_init()))
```
sdroege, line 19: Actually `time_read` is always valid, even if `NULL` is returned.

gio/src/unix_mount_entry.rs
```diff
@@ -9,30 +9,28 @@ use std::mem;
...
         unsafe {
             let mut time_read = mem::MaybeUninit::uninit();
-            let ret = from_glib_full(ffi::g_unix_mount_for(
+            let ret: Option<UnixMountEntry> = from_glib_none(ffi::g_unix_mount_for(
                 file_path.as_ref().to_glib_none().0,
                 time_read.as_mut_ptr(),
             ));
-            let time_read = time_read.assume_init();
-            (ret, time_read)
+            ret.map(|x| (x, time_read.assume_init()))
```
sdroege, line 33: Same here

sdroege (changes requested):
Otherwise seems correct :)

gio/src/unix_mount_entry.rs
sophie-h: Okay, was to lazy to check the C code

gio/src/unix_mount_entry.rs
```diff
@@ -9,10 +9,10 @@ use std::mem;
 
 impl UnixMountEntry {
     #[doc(alias = "g_unix_mount_at")]
-    pub fn new_at<P: AsRef<std::path::Path>>(mount_path: P) -> (UnixMountEntry, u64) {
+    pub fn new_at<P: AsRef<std::path::Path>>(mount_path: P) -> (Option<UnixMountEntry>, u64) {
         unsafe {
             let mut time_read = mem::MaybeUninit::uninit();
-            let ret = from_glib_full(ffi::g_unix_mount_at(
+            let ret = from_glib_none(ffi::g_unix_mount_at(
```
sdroege, line 15: It's `transfer full` though, or is that annotation wrong too?

sophie-h: Somehow I was thinking that none=Option. I will have a look if I did this mistake in other commits :|

sdroege (approved):

Issue #462, opened 2021-04-18
translate: trait TryFromGlib should be marked unsafe


fengalin:
See https://github.com/gtk-rs/gtk-rs/pull/454#pullrequestreview-638331872

PR #463, opened 2021-04-18
translate: make try_from_glib unsafe

fengalin:
A defunct implementation could lead to inconsistencies or UB due to
a valid Rust value being constructed from an invalid or undefined
(aka None) C value.

Fixes https://github.com/gtk-rs/gtk-rs/issues/462

### DIFF
--- a/glib-macros/src/genum_derive.rs
+++ b/glib-macros/src/genum_derive.rs
@@ -98,7 +98,7 @@ pub fn impl_genum(input: &syn::DeriveInput) -> TokenStream {
         impl #crate_ident::translate::TryFromGlib<i32> for #name {
             type Error = i32;
 
-            fn try_from_glib(value: i32) -> Result<Self, i32> {
+            unsafe fn try_from_glib(value: i32) -> Result<Self, i32> {
                 let from_glib = || {
                     #from_glib
                 };
--- a/glib/src/translate.rs
+++ b/glib/src/translate.rs
@@ -53,7 +53,7 @@
 //! # }
 //! impl TryFromGlib<libc::c_uint> for SpecialU32 {
 //!     type Error = GlibNoneError;
-//!     fn try_from_glib(val: libc::c_uint) -> Result<Self, GlibNoneError> {
+//!     unsafe fn try_from_glib(val: libc::c_uint) -> Result<Self, GlibNoneError> {
 //!         if val == SpecialU32::GLIB_NONE {
 //!             return Err(GlibNoneError);
 //!         }
@@ -73,7 +73,7 @@
 //! struct U32(u32);
 //! impl TryFromGlib<libc::c_long> for U32 {
 //!     type Error = TryFromIntError;
-//!     fn try_from_glib(val: libc::c_long) -> Result<Self, TryFromIntError> {
+//!     unsafe fn try_from_glib(val: libc::c_long) -> Result<Self, TryFromIntError> {
 //!         Ok(U32(u32::try_from(val)?))
 //!     }
 //! }
@@ -1172,7 +1172,7 @@ impl FromGlib<i32> for Ordering {
 /// Translate from a Glib type which can result in an undefined and/or invalid value.
 pub trait TryFromGlib<G: Copy>: Sized {
     type Error;
-    fn try_from_glib(val: G) -> Result<Self, Self::Error>;
+    unsafe fn try_from_glib(val: G) -> Result<Self, Self::Error>;
 }
 
 /// Error type for [`TryFromGlib`] when the Glib value is None.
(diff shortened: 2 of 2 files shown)


sdroege:
Can you think of a case where this could cause actual unsafe behaviour without the implementation requiring an `unsafe` block? I think this here is actually safe.

The difference to the error checking trait from the other PR is that that one is called from `unsafe` code and is supposed to check things that following `unsafe` code relies on.

fengalin:
> Can you think of a case where this could cause actual unsafe behaviour without the implementation requiring an unsafe block? I think this here is actually safe.

So far, we have only three real cases for this:

- `*Return` to `Result<*Success, *Error>` conversion (in `gst`). A defunct function would interpret the return value incorrectly and could perform an operation when it should be prevented by the result. A buggy code could also do this kind of things, without the conversion being defunct, but the application developer would be mislead by the defunct.
- While proceeding with this PR, I noticed that someone used this trait in the `genum` macro. The problem would be similar to the above. The `enum` specific conversion usually takes care of variants unknown to the bindings, so that shouldn't be a problem. But inconsistencies could arise in a transitively defunct implementation.
- `gst::ClockTime`. A defunct implementation would be one that fails to identify the `GST_CLOCKTIME_NONE` value which is `u64::MAX`. This would be a valid large `gst::ClockTime` in Rust code, but once passed back to C, it would be considered as `GST_CLOCKTIME_NONE`, which would be inconsistent. We might not be talking about UB here, but it seems worth pointing out that this implementation requires extra care.

fengalin:
Note this mention in the [nomicon](https://doc.rust-lang.org/nomicon/what-unsafe-does.html?#what-unsafe-rust-can-do), about what could lead to Undefined Behavior:

> * Producing invalid values (either alone or as a field of a compound type such as enum/struct/array/tuple): 
> [...]
>   o a type with custom invalid values that is one of those values, such as a `NonNull` that is null. (Requesting custom invalid values is an unstable feature, but some stable libstd types, like `NonNull`, make use of it.)

fengalin:
On the other hand, [`std::convert::TryFrom`](https://doc.rust-lang.org/std/convert/trait.TryFrom.html) is not marked `unsafe`...

GuillaumeGomez:
My little opinion here: an item is marked unsafe only if using it can panic. In rust, there should never be UB. If something doesn't go as expected and cannot be recovered, we panic. If `TryFromGlib` (or whatever the name) can introduce UB or panic, maybe it shouldn't be called as such.

sdroege:
What does panicking have to do with unsafe?

If the trait is unsafe or not depends on the guarantees it wants to uphold. It would have to be unsafe if an (otherwise safe! If the implementation uses unsafe code all this doesn't matter) implementation has to hold up more guarantees than what can be checked by the compiler, and otherwise would cause UB. I don't think that's the case here: it will either return a valid value or fail cleanly with an `Err`. What following code does with those values is their problem really.

GuillaumeGomez:
So is there a point into making this unsafe?

fengalin:
> So is there a point into making this unsafe?

That's the point of this discussion.

I gave it additional thinking while working on `ClockTime` & co in `gstreamer-rs`. As a reminder, `GST_CLOCK_TIME_NONE` equals `u64::MAX`. As we will able to use `Option<ClockTime>`, it's actually not a problem in Rust to embed `u64::MAX` in `ClockTime`. It could even silently occur before after an arithmetic operation for instance. The real problem is crossing `ffi` toward C. In a way, this is were any check should be performed. But I guess we don't really want a `TryToGlib` for this kind of stuff. So what should we do, `panic`?

sdroege:
> I gave it additional thinking while working on `ClockTime` & co in `gstreamer-rs`. As a reminder, `GST_CLOCK_TIME_NONE` equals `u64::MAX`. As we will able to use `Option<ClockTime>`, it's actually not a problem in Rust to embed `u64::MAX` in `ClockTime`. It could even silently occur before, after an arithmetic operation for instance. The real problem is crossing `ffi` toward C. In a way, this is where any check should be performed.

Btw, does the new value API help here for getting a `Option<gst::ClockTime>` from/into a `GValue`?

>  But I guess we don't really want a `TryToGlib` for this kind of stuff. So what should we do, `panic`?

I would silently have it wrap to `GST_CLOCK_TIME_NONE` as it can't be represented by the original value, and other calculations overflowing would also map to `NONE`.

fengalin:
> Btw, does the new value API help here for getting a Option<gst::ClockTime> from/into a GValue?

I didn't look too closely, but it should help yes.

> I would silently have it wrap to GST_CLOCK_TIME_NONE as it can't be represented by the original value, and other calculations overflowing would also map to NONE.

Then I guess the question how soon do we want to detect that:

- Should things like `ClockTime::from_nsecond` return `Option<ClockTime>`? Probably not.
- It's easy to implement for the `checked_*` op, but should a regular arithmetic operation return an `Option`? 

I think it might be quite difficult a problem for users to debug: a `ClockTime` resulting from an operation that is treated as `None` when passed to C.

fengalin:
Maybe we should just log a warning in `ToGlib` in that case. At least, that would help users spot the problem.

sdroege:
Hmm I don't know!

GuillaumeGomez:
Needs to be re-open on gtk-rs-core.

PR #466, opened 2021-04-19
Update to proc-macro-crate 1.0

sdroege:
@bilelmoussaoui Also applies to gtk4-macros I guess


bilelmoussaoui (approved):

examples/Cargo.toml
```diff
@@ -11,6 +11,9 @@ futures-channel = "0.3"
 futures-util = "0.3"
 once_cell = "1.2.0"
 
+[dependencies.glib]
+path = "../glib"
```
sdroege, line 15: Good that this became an assertion instead of silently doing random things :)

gtk3-macros/src/lib.rs
```diff
@@ -21,7 +21,7 @@ use syn::{parse_macro_input, DeriveInput};
 /// Specify that `MyWidget` is using a composite template and load the
 /// template file the `composite_template.ui` file.
 ///
-/// ```
+/// ```compile_fail
```
GuillaumeGomez, line 24: Was it supposed to fail at compilation?

sdroege: It could've never compiled so I'd assume yes. It's also not tested by the CI.

Let me create an issue for that, thanks for the reminder.

sdroege: https://github.com/gtk-rs/gtk-rs/issues/468

sdroege: Note that the macros themselves are tested by the examples

GuillaumeGomez: Not the problem here. ;)

I explained it a bit more on the issue you opened.

bilelmoussaoui:
The update is not great. as we can't really use the re-exported glib anymore and everyone has to explicitly add glib as a dependency :( see https://github.com/gtk-rs/gtk4-rs/pull/308

examples/Cargo.toml
bilelmoussaoui: Except glib is re-exported by gtk, not sure if that's a good thing actually :)

sdroege: The problem with that is that it worked accidentally before. If the dependency was not found then it was using simply the name `glib`. This required `glib` manually to be put into the scope by the user, and it wouldn't even be guaranteed that it's the correct `glib`.

PR #471, opened 2021-04-20
glib-macros: Try harder to find glib

mehmooda:
Also fallback to leaving it to user choice

### DIFF
--- a/glib-macros/src/utils.rs
+++ b/glib-macros/src/utils.rs
@@ -148,24 +148,21 @@ pub fn parse_item_attributes(attr_name: &str, attrs: &[Attribute]) -> Result<Vec
     Ok(v)
 }
 
-const KNOWN_GLIB_EXPORTS: [&str; 14] = [
+const KNOWN_GLIB_EXPORTS: [&str; 10] = [
     // Current Re-exports from gtk-rs
     "gtk",
     "gio",
     "gdk",
-    "gdk-pixbuf",
-    "gdkx11",
-    "graphene",
     "pango",
-    "pangocairo",
+    "graphene-rs",
     // Current Re-exports from gtk4-rs
     "gtk4",
     "gsk4",
     "gdk4",
-    "gdk4-wayland",
-    "gdk4-x11",
     // Special Request
     "gstreamer",
+    // Re-export may fail
+    "cairo-rs",
 ];
 
 pub fn crate_ident_new() -> TokenStream {
--- a/glib-macros/src/utils.rs
+++ b/glib-macros/src/utils.rs
@@ -148,15 +148,63 @@ pub fn parse_item_attributes(attr_name: &str, attrs: &[Attribute]) -> Result<Vec
     Ok(v)
 }
 
-pub fn crate_ident_new() -> Ident {
+const KNOWN_GLIB_EXPORTS: [&str; 14] = [
+    // Current Re-exports from gtk-rs
+    "gtk",
+    "gio",
+    "gdk",
+    "gdk-pixbuf",
+    "gdkx11",
+    "graphene",
+    "pango",
+    "pangocairo",
+    // Current Re-exports from gtk4-rs
+    "gtk4",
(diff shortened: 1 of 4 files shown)


glib-macros/src/utils.rs
```diff
@@ -148,15 +148,63 @@ pub fn parse_item_attributes(attr_name: &str, attrs: &[Attribute]) -> Result<Vec
...
+    "gdkx11",
+    "graphene",
+    "pango",
+    "pangocairo",
+    // Current Re-exports from gtk4-rs
+    "gtk4",
+    "gsk4",
+    "gdk4",
+    "gdk4-wayland",
+    "gdk4-x11",
```
sdroege, line 166: Does it make sense to include these two (and `gdkx11` above)? You'll always have `gdk4` / `gdk` in your dependencies anyway.

Same for `pangocairo` and `pango`. They're more "extension" crates but you'd use the base one too always. Same thing with e.g. `libhandy` and `gtk` / `libadwaita` and `gtk4` and `gstreamer-*` and `gstreamer`.

sdroege (commented):
Please also remove the `glib` dependency from `examples/Cargo.toml` then. That should be unneeded now.

sdroege:
> Please also remove the `glib` dependency from `examples/Cargo.toml` then. That should be unneeded now.

That said, IMHO the plain GLib and GIO examples shouldn't have to do `use gtk::gio` etc but just use the plain crate... otherwise it might look to people like you always have to use GTK (which you don't!).

bilelmoussaoui:
> > Please also remove the `glib` dependency from `examples/Cargo.toml` then. That should be unneeded now.
> 
> That said, IMHO the plain GLib and GIO examples shouldn't have to do `use gtk::gio` etc but just use the plain crate... otherwise it might look to people like you always have to use GTK (which you don't!).

Indeed. It only makes sense for gtk specific examples

GuillaumeGomez:
What happens in case you renamed of those dependencies?

sdroege:
> What happens in case you renamed of those dependencies?

That's exactly the point of `proc_macro_crate`. It detects that :)

glib-macros/src/utils.rs
bilelmoussaoui: Yes, you can drop most of those and only keep gtk4/gdk4/gsk4 for gtk4-rs crates; gio/gdk for gtk-rs crates.

mehmooda: Will do just to confirm new list: 
    "gtk",
    "gio",
    "gdk",
    "graphene",
    "pango",
    "gtk4",
    "gsk4",
    "gdk4",
    "gstreamer",

bilelmoussaoui: looks fine to me, @sdroege ?

sdroege: Those plus `cairo` (note that it's called `cairo-rs`). Also `graphene` is called `graphene-rs`.

mehmooda: I didn't put cairo in as glib was feature gated. If someone depended on cairo but not "use-glib" then it would try to use cairo and they wouldn't be able to fix it.

I guess they could directly depend on glib as a catchall

sdroege: The glib feature is enabled by default, so if they use `glib-macros` they probably didn't explictly disable the glib feature of cairo :)

sdroege: But maybe put cairo last because of this.

sdroege (approved):

bilelmoussaoui:
> Please also remove the glib dependency from examples/Cargo.toml then. That should be unneeded now.

This is still needed. Otherwise looks good to me!

sdroege:
> This is still needed. Otherwise looks good to me!

Well, we said it should gain a `gio` (and whatever) dependency and have the non-GTK examples changed accordingly :) But that can happen in a separate PR.

PR #483, opened 2021-04-25
gio: Rename TlsConnectionManualExt to ExtManual and export from prelude

MarijnS95:
Noticed in [1] that this trait has ManualExt written the wrong way around, and is exported from the crate root instead of the prelude where all other traits are reexported.

[1]: https://github.com/gtk-rs/gir/pull/1111#issuecomment-826319140

CC @bilelmoussaoui

### DIFF
--- a/gio/src/lib.rs
+++ b/gio/src/lib.rs
@@ -99,7 +99,6 @@ pub use crate::read_input_stream::ReadInputStream;
 mod write_output_stream;
 pub use crate::write_output_stream::WriteOutputStream;
 mod tls_connection;
-pub use crate::tls_connection::TlsConnectionManualExt;
 
 pub mod task;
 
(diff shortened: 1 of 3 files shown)


bilelmoussaoui:
The rename should he done in Gir.toml as well or otherwise added to the manual_traits list

MarijnS95:
@bilelmoussaoui It was already named the correct way around in `Gir.toml`:

https://github.com/gtk-rs/gtk-rs/blob/c72ce07ce8aaa0e8584e27ad88cbbfe89d6acf72/gio/Gir.toml#L1063-L1066

^^

bilelmoussaoui (approved):

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #485, opened 2021-04-25
GH workflows: Use actions-rs/toolchain `components` to install clippy

MarijnS95:
This is a bit more elegant than manually invoking `rustup component add`.  Afaik it is synonymous to `rustup toolchain install <toolchain> --components <components>` which is why it has not been applied to the clippy step in CI.yml, as that would limit available nightlies when clippy is only used in the `stable` matrix runs.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -138,7 +138,7 @@ jobs:
           profile: minimal
           toolchain: stable
           override: true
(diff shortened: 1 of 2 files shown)


bilelmoussaoui (approved):

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #487, opened 2021-04-26
Don't re-export traits from crate root

sdroege:


sdroege:
* gstreamer-rs MR: https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/751
* gtk4-rs PR: https://github.com/gtk-rs/gtk4-rs/pull/323

gio/src/unix_socket_address.rs
```diff
@@ -1,10 +1,10 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
+use crate::prelude::*;
 use crate::SocketAddress;
 use crate::UnixSocketAddress;
-use crate::UnixSocketAddressExt;
 use crate::UnixSocketAddressType;
-use glib::object::{Cast, IsA};
+use glib::prelude::*;
```
bilelmoussaoui, line 7: i believe gio prelude contains glib's prelude already no?

bilelmoussaoui (approved):

gio/src/unix_socket_address.rs
sdroege: Indeed, removed here and in a few other places.

bilelmoussaoui:
Something seems to be wrong with glib when generating the docs

sdroege:
That passed before though... did we maybe get a new nightly in the last 3 hours and that broke things? :upside_down_face:

sdroege:
Apparently just a case of cosmic rays.

PR #490, opened 2021-04-26
Add rename doc aliases

fengalin:
This is the result for `gtk-rs` of an attempt at fixing https://github.com/gtk-rs/gir/issues/1105.

It also addresses concerns described by @sdroege regarding some doc aliases which didn't match the name in upstream doc due to name mangling or property naming conventions. It also adds doc aliases for child properties which were not generated before.

Note that aliases from properties are no longer generated when they match the final name, hence the reduction in the "prop name alias" step below.

This tab shows the variations in size for `search-index.js`:
| step                  | size      | delta (1) |
| --------------------- |:---------:| ---------:|
|   master              | 7 420 982 |           |
| prop name alias (2)   | 7 406 753 |     -0.2% |
| gir fn name alias  | 7 447 813 |     +0.4% |
| fix-getters-def alias | 7 453 449 |     +0.4% |

(1) Delta on `search-index.js` size / `master`.
(2) Includes doc-aliases for child properties which were not generated previously.


GuillaumeGomez:
Do you also have a live version we could test? :)

fengalin:
How / where could I push the doc?
Otherwise, just checkout my branch :)

GuillaumeGomez:
I'll check out locally later then.

gtk/src/auto/accel_group.rs
```diff
@@ -84,14 +87,17 @@ pub trait AccelGroupExt: 'static {
         f: F,
     ) -> SignalHandlerId;
 
+    #[doc(alias = "accel-changed")]
     fn connect_accel_changed<F: Fn(&Self, u32, gdk::ModifierType, &glib::Closure) + 'static>(
```
fengalin, line 91: @sdroege: I also added doc aliases for signals. Thought that would be consistent with your proposal for properties getter / setter (see other comment below)...

gtk/src/auto/accel_group.rs
```diff
@@ -84,14 +87,17 @@ pub trait AccelGroupExt: 'static {
...
         &self,
         detail: Option<&str>,
         f: F,
     ) -> SignalHandlerId;
 
+    #[doc(alias = "notify::is-locked")]
     fn connect_property_is_locked_notify<F: Fn(&Self) + 'static>(&self, f: F) -> SignalHandlerId;
```
fengalin, line 98: @sdroege : ... For some signal, the name is a bit different though. This is the name we pass to the C API, so I guess it's ok.

fengalin (commented):
A little update since yesterday. This adds a few doc aliases, but I think they are worth it.

gtk/src/auto/accel_group.rs
sdroege: Good idea, yes. Forgot about signals. That was another case where we diverged from the C API before already (CC @GuillaumeGomez )

gtk/src/auto/accel_group.rs
sdroege: Different in what sense? But yes, always put the C signal name there.

However: the specific case of the notify signal should IMHO just get the doc alias for the property name. It's part of the property.

fengalin: > However: the specific case of the notify signal should IMHO just get the doc alias for the property name. It's part of the property.

This is the difference I was referring too :) I'm on it.

sdroege: Good that you mention this. This also needs another rename :) `connect_property_is_locked_notify` -> `connect_is_locked_notify`. No point in having the word property in there if we don't have it for the getter/setter anymore.

fengalin: Indeed!

fengalin: I just hope, this will not lead me to perform name conflict resolution like for getters...

sdroege: That would surprise me very much.

gtk/src/auto/accel_group.rs
```diff
@@ -256,7 +261,8 @@ impl<O: IsA<AccelGroup>> AccelGroupExt for O {
         }
     }
 
-    fn connect_property_is_locked_notify<F: Fn(&Self) + 'static>(&self, f: F) -> SignalHandlerId {
+    #[doc(alias = "is-locked")]
+    fn connect_is_locked_notify<F: Fn(&Self) + 'static>(&self, f: F) -> SignalHandlerId {
```
fengalin, line 265: It went smoothly. :)

gtk/src/auto/accel_group.rs
fengalin: You were right. It went smoothly :)

We get this now:

```rust
    #[doc(alias = "is-locked")]
    fn connect_is_locked_notify<F: Fn(&Self) + 'static>(&self, f: F) -> SignalHandlerId {
```
[link to code](https://github.com/fengalin/gtk-rs/blob/9451abcdb734624e5b2b888facab6f9f746b63d8/gtk/src/auto/accel_group.rs#L264)

sdroege: That's awesome

atk/src/auto/document.rs
```diff
@@ -119,6 +130,7 @@ impl<O: IsA<Document>> DocumentExt for O {
         }
     }
 
+    #[doc(alias = "load-stopped")]
```
GuillaumeGomez, line 133: dash?

fengalin: Yeah, we discussed this before. See comments above. This is intended to help people find the methods from the name of the signal from C doc.

bilelmoussaoui:
Thanks a lot @fengalin !

PR #493, opened 2021-04-27
Impl IntoGlib for Result...

fengalin:
... when both the `Ok` and `Error` types implement `IntoGlib` to the same type and they are bound to a `Result` type due to a `TryFromGlib` impl. This helps the implementation of trampolines which return a type which `impl` `TryFromGlib`.

Also, rename `OptionToGlib` as `OptionIntoGlib` for consistency with `ToGlib` which was renamed as `IntoGlib`

### DIFF
--- a/glib/src/translate.rs
+++ b/glib/src/translate.rs
@@ -18,7 +18,7 @@
 //!     }
 //! ```
 //!
-//! Implementing [`OptionToGlib`] on a Rust type `T` allows specifying a sentinel to indicate
+//! Implementing [`OptionIntoGlib`] on a Rust type `T` allows specifying a sentinel to indicate
 //! a `None` value and auto-implementing [`FromGlib`] for `Option<T>`, which would not be
 //! possible in dependent crates due to the [orphan rule](https://doc.rust-lang.org/book/ch10-02-traits.html#implementing-a-trait-on-a-type).
 //! In the example below, [`IntoGlib`] is auto-implemented for `Option<SpecialU32>`.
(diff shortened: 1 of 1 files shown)


glib/src/translate.rs
```diff
@@ -340,12 +340,29 @@ impl IntoGlib for Ordering {
     }
 }
 
+impl<O, E, G> IntoGlib for Result<O, E>
+where
+    G: Copy,
+    O: IntoGlib<GlibType = G>,
+    E: IntoGlib<GlibType = G>,
```
sdroege, line 347: My main concern here would be that this requires both variants to be disjoint. If you have a `Result<SomeEnum, SomeOtherEnum>` this would not work correctly, for example :)

fengalin: Initially, I wanted to tie this to both types implementing `From<FlowReturn>` for instance and forcing the result as the `GlibType` for `FlowReturn`. But then the trait system considers it is not constrained enough. Which I vaguely understood as there might be several candidates for this which would not allow to select one and only impl.

So this is the best solution I came up with.

fengalin: To clarify :). I wanted this:

```rust
impl<O, E, F> IntoGlib for Result<O, E>
where
    F: IntoGlib + From<Result<O, E>>,
{
    type GlibType = <F as IntoGlib>::GlibType;

    #[inline]
    fn into_glib(self) -> Self::GlibType {
        F::from(self).into_glib()
    }
}
```

And it leads to that:

```
error[E0207]: the type parameter `F` is not constrained by the impl trait, self type, or predicates
   --> glib/src/translate.rs:362:12
    |
362 | impl<O, E, F> IntoGlib for Result<O, E>
    |            ^ unconstrained type parameter
```

sdroege: Yeah that's exactly why I don't understand why your simplification of `FromValue` worked :)

fengalin: Yes, I guess the difference is that I used the "free" type parameter in an expression such as `IntoGlib<GlibType = G>`.

sdroege (approved):

Issue #491, opened 2021-04-26
glib: Ergonomics of 'Char' seem questionalble


ids1024:
The [docs](https://docs.rs/glib/0.10.3/glib/char/struct.Char.html) suggest this is meant to be more ergonomic that taking `libc:c_char`. But is it an improvement over just accepting `u8`? Rust has a `b''` syntax.

For the `set_list_separator` example in the documentation:
```rust
key_file.set_list_separator(glib::Char::new(';').unwrap());
key_file.set_list_separator(b';');
key_file.set_list_separator(';' as u8);
```

Apparently `u8` doesn't implement `TryFrom<char>` (https://github.com/rust-lang/rfcs/issues/2854), which would be useful for panicking behavior similar to `glib::Char::new(...).unwrap()` when out of range. But on the whole, accepting a `u8` seems more ergonomic for most uses.

ids1024:
For reference, `libc::c_char` seems to be `i8` or `u8` depending on the platform. I believe under the C standard a `char` has to be at least 8 bits, and `sizeof(char)` has to be `1`, but it could technically be more than 8 bits. I doubt Rust or gtk are really usable on some bizarre hardware with bytes that aren't 8 bits though.

So it should be fine to use `u8`, which is the most natural type in Rust (given that `b''` is interpreted as a `u8`).

sdroege:
I agree, and also if we keep `Char` we should remove its constructor and make it a `TryFrom` impl instead.

@bilelmoussaoui @GuillaumeGomez ?

GuillaumeGomez:
I'm fine with replacing `Char` constructor.

sdroege:
@GuillaumeGomez What advantage over using `u8` do you see that justifies the existence of the type? Not using it and directly using `u8` seems easier in all cases I can think of.

GuillaumeGomez:
`Char` is more explicit than `u8`. That's the only reason. We could use `libc::c_char` too but I find it less "nice".

ids1024:
`libc::c_char` seems like an awkward type to use (except for ffi) since it could be a `u8` or `i8`, and the `b''` byte char syntax can only be used with `u8`.

So I guess `glib::Char(b';')` currently works on some platforms but fails to compile on others. So even if a newtype pattern like this is preferred, perhaps it should contain a `u8` rather than `c_char`.

GuillaumeGomez:
[...]

PR #494, opened 2021-04-28
glib: `From` and `TryFrom` implementations for working with `Char`

ids1024:
See https://github.com/gtk-rs/gtk-rs/issues/491

This leaves `fn new()` unchanged. Don't know if it would be better to change or remove that.

### DIFF
--- a/glib/src/char.rs
+++ b/glib/src/char.rs
@@ -3,6 +3,8 @@
 use crate::translate::FromGlib;
 use crate::translate::IntoGlib;
 use libc::{c_char, c_uchar};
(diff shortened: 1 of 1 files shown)


GuillaumeGomez:
Maybe better to simply remove `new`.

ids1024:
Updated to change `UChar` as well, and remove `new()` functions.

`glib::Char::from(b'a')` is still rather verbose compared to `b'a'`, but at least it's a single function call without any unwrapping.

sdroege (approved):

PR #495, opened 2021-04-29
gtk: rename StyleContext::property to StyleContext::style_property

### LINKED ISSUE #489: Multiple traits with a `fn property()`: `glib::ObjectExt` and `gtk::StyleContextExt`
This should probably be resolved by renaming the second one to `style_property()`.

Same probably applies to gtk4 too.

CC @bilelmoussaoui

bilelmoussaoui:
fixes #489

### DIFF
--- a/gtk/Gir.toml
+++ b/gtk/Gir.toml
@@ -2102,6 +2102,9 @@ manual_traits = ["StyleContextExtManual"]
     [[object.property]]
     name = "paint-clock"
     version = "3.8"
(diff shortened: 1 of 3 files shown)


bilelmoussaoui:
Ok i'm too tired, there's already another function named `style_property` which has different parameters. Not sure how to handle this one. Rename it back to get_property?

bilelmoussaoui:
Maybe rename this one to `style_property_for_state`?

PR #496, opened 2021-04-29
gtk3-macros: make the doc example building

### LINKED ISSUE #468: gtk3-macros doctests are `compile_fail`
See https://github.com/gtk-rs/gtk-rs/pull/466#discussion_r615895426

Is this intentional? @bilelmoussaoui @BrainBlasted ?

bilelmoussaoui:
fixes #468


bilelmoussaoui:
Interesting, the CI never runs gtk3-macros tests :) Also per my local tests, it seems like the doc tests don't like using a `imp` module & having `use super::*` on it. It never finds the defined public struct which makes getting this test to at least compile a bit harder than I thought. Any ideas?

gtk3-macros/Cargo.toml
```diff
@@ -25,3 +26,6 @@ proc-macro2 = "1.0"
 quote = "1.0"
 syn = { version = "1.0", default-features = false, features = ["full"] }
 proc-macro-crate = "1.0"
+
+[dev-dependencies]
+gtk = { path = "../gtk" }
```
sdroege, line 31:
```suggestion
gtk = { path = "../gtk" }
glib = { path = "../glib" }
```

So that this is fixed probably

```
 error[E0433]: failed to resolve: use of undeclared crate or module `glib`
  --> src/lib.rs:45:7
   |
19 |     #[glib::object_subclass]
   |       ^^^^ use of undeclared crate or module `glib`
```

gtk3-macros/src/lib.rs
```diff
@@ -21,38 +21,57 @@ use syn::{parse_macro_input, DeriveInput};
...
-///        }
-///
-///        fn instance_init(obj: &glib::subclass::InitializingObject<Self>) {
-///            obj.init_template();
-///        }
-///    }
+/// mod imp {
+///     use super::*;
+/// 
+///     #[derive(Debug, Default, CompositeTemplate)]
```
sdroege, line 36:
```suggestion
///     #[derive(Debug, Default, gtk::CompositeTemplate)]
```

So that this is probably fixed

```
 error: cannot find derive macro `CompositeTemplate` in this scope
  --> src/lib.rs:36:30
   |
10 |     #[derive(Debug, Default, CompositeTemplate)]
   |                              ^^^^^^^^^^^^^^^^^
```

gtk3-macros/src/lib.rs
```diff
@@ -21,38 +21,57 @@ use syn::{parse_macro_input, DeriveInput};
...
+///     #[template(file = "composite_template.ui")]
+///     pub struct MyWidget {
+///         #[template_child]
+///         pub label: TemplateChild<gtk::Label>,
```
sdroege, line 40:
```suggestion
///         pub label: gtk::TemplateChild<gtk::Label>,
```

Maybe? Or how does this work exactly?

Currently this fails

```
error[E0412]: cannot find type `TemplateChild` in this scope
  --> src/lib.rs:40:20
   |
14 |         pub label: TemplateChild<gtk::Label>,
   |                    ^^^^^^^^^^^^^ not found in this scope
```

gtk3-macros/src/lib.rs
```diff
@@ -21,38 +21,57 @@ use syn::{parse_macro_input, DeriveInput};
...
+///             obj.init_template();
+///         }
+///     }
+/// 
+///     impl ObjectImpl for MyWidget {}
+///     impl WidgetImpl for MyWidget {}
+///     impl ContainerImpl for MyWidget {}
+///     impl BoxImpl for MyWidget {}
+/// }
+///
```
sdroege, line 65:
```suggestion
///
```

and the same everywhere else here

gtk3-macros/src/lib.rs
bilelmoussaoui: The current code should work just fine. Even if I do as you suggest it fails to find super::MyWidget.

sdroege: Well, it doesn't find `glib` so `glib::wrapper` is not expanded and no `super::MyWidget` is defined :)

sdroege: Oh hm, but you do `use gtk::glib`. Then I have no idea what is wrong here. Seems like a bug

bilelmoussaoui: It does sound like one, any ideas @GuillaumeGomez ?

GuillaumeGomez: The example is run as a binary containing itself. `super` doesn't match anything in it.

bilelmoussaoui: but the example is has a module `imp`, super is used inside that module. Which would default to everything outside the `imp` module. Unless i'm mis-understanding?

bilelmoussaoui:
Blocked by https://github.com/rust-lang/rust/issues/85433

GuillaumeGomez:
rust-lang/rust#85433 has been merged so I guess it's not blocked anymore? :)

bilelmoussaoui:
Yes it's not, I fixed it already in gtk4 but never got the time to fix it here

bilelmoussaoui:
@GuillaumeGomez feel free to push against my branch if you want to get this one merged / #468 fixed

GuillaumeGomez:
Updated. :)

GuillaumeGomez:
@sdroege Any idea what's going on here? O.o

sdroege:
You need to update to the new emit_by_name() API

sdroege:
Well, gir. I forgot about that yesterday and won't get to it before this evening. Sorry about that

We should build gtk and gstreamer as part of the core CI

bilelmoussaoui:
> Well, gir. I forgot about that yesterday and won't get to it before this evening. Sorry about that

https://github.com/gtk-rs/gir/pull/1275 should unlock the situation :-)

bilelmoussaoui:
@GuillaumeGomez should be working now if you pull the latest changes from the master branch

PR #497, opened 2021-04-29
gtk: Add binding for `gtk_container_class_handle_border_width()`

ids1024:


gtk/src/subclass/container.rs
```diff
@@ -268,3 +268,19 @@ impl Callback {
...
+
+pub unsafe trait ContainerClassSubclassExt: ClassStruct {
+    #[doc(alias = "gtk_container_class_handle_border_width")]
+    fn handle_border_width(&mut self) {
+        unsafe {
+            let type_class = self as *mut _ as *mut glib::gobject_ffi::GTypeClass;
+            let widget_class = glib::gobject_ffi::g_type_check_class_cast(
+                type_class,
+                ffi::gtk_container_get_type(),
+            ) as *mut ffi::GtkContainerClass;
```
bilelmoussaoui, line 280:
```suggestion
            let widget_class = glib::gobject_ffi::g_type_check_class_cast(
                type_class,
                ffi::gtk_container_get_type(),
            ) as *mut ffi::GtkContainerClass;
```
Is this needed? the check is already implied by implementing the trait only for widgets that implement `ContainerImpl`. I suppose you can directly cast it to as `GtkContainerClass`

ids1024: I copied it from the methods of `WidgetClassSubclassExt`. But yeah, probably not necessary.

sdroege: Yes, not necessary and also not there :)

sdroege (approved):
Looks good to me otherwise. Can you replace that with a cast here (and in a separate commit for the others)?

ids1024:
> Can you replace that with a cast here (and in a separate commit for the others)

Done.

bilelmoussaoui (approved):

PR #498, opened 2021-04-29
Make cairo error handling slightly better

bilelmoussaoui:
This should handle most of the concerns @federicomenaquintero  had regarding the errors handling on Cairo's side

See #74 for details

### DIFF
--- a/cairo/src/context.rs
+++ b/cairo/src/context.rs
@@ -587,12 +587,12 @@ impl Context {
     }
 
     #[doc(alias = "get_font_options")]
-    pub fn font_options(&self) -> FontOptions {
-        let out = FontOptions::new();
+    pub fn font_options(&self) -> Result<FontOptions, Error> {
+        let out = FontOptions::new()?;
         unsafe {
             ffi::cairo_get_font_options(self.0.as_ptr(), out.to_raw_none());
         }
-        out
+        Ok(out)
     }
 
     pub fn set_font_face(&self, font_face: &FontFace) {
--- a/cairo/src/context.rs
+++ b/cairo/src/context.rs
@@ -233,7 +233,6 @@ impl Context {
 
     pub fn set_antialias(&self, antialias: Antialias) {
         unsafe { ffi::cairo_set_antialias(self.0.as_ptr(), antialias.into()) }
-        self.status().expect("Failed to set antialias");
     }
 
     #[doc(alias = "get_antialias")]
@@ -250,7 +249,6 @@ impl Context {
                 offset,
             )
         }
-        self.status().expect("Failed to set a dash"); //Possible invalid dashes value
     }
 
     #[doc(alias = "get_dash_count")]
@@ -287,7 +285,6 @@ impl Context {
         unsafe {
             ffi::cairo_set_fill_rule(self.0.as_ptr(), fill_rule.into());
         }
-        self.status().expect("Failed to set fill rule");
     }
 
     #[doc(alias = "get_fill_rule")]
@@ -297,7 +294,6 @@ impl Context {
 
     pub fn set_line_cap(&self, arg: LineCap) {
         unsafe { ffi::cairo_set_line_cap(self.0.as_ptr(), arg.into()) }
-        self.status().expect("Failed to set line cap");
     }
 
     #[doc(alias = "get_line_cap")]
@@ -307,7 +303,6 @@ impl Context {
 
     pub fn set_line_join(&self, arg: LineJoin) {
         unsafe { ffi::cairo_set_line_join(self.0.as_ptr(), arg.into()) }
-        self.status().expect("Failed to set line join");
     }
 
     #[doc(alias = "get_line_join")]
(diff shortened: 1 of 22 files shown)


federicomenaquintero:
This looks really good!  Thanks!

bilelmoussaoui:
This should be ready for a review whenever you have the time for that. The changes are mostly
- Re-expose status
- Don't panic on getters/constructors, instead return an Error so that the caller has to handle the error
- Don't panic on setters and don't return an error neither but instead handle that when the user writes the changes to the target surface

The only call that might panic is `cairo_font_options_copy`, I don't think there's a way around that one.

sdroege:
> The only call that might panic is `cairo_font_options_copy`, I don't think there's a way around that one.

Can that actually fail?! But yes, can't really do anything if that fails :)

cairo/src/context.rs
```diff
@@ -343,7 +342,7 @@ impl Context {
         unsafe { ffi::cairo_clip_preserve(self.0.as_ptr()) }
     }
 
-    pub fn clip_extents(&self) -> (f64, f64, f64, f64) {
+    pub fn clip_extents(&self) -> Result<(f64, f64, f64, f64), Error> {
```
sdroege, line 345: Why can this and `in_clip` fail? They're just querying or not?

cairo/src/context.rs
```diff
@@ -393,40 +394,49 @@ impl Context {
         unsafe {
             ffi::cairo_fill_extents(self.0.as_ptr(), &mut x1, &mut y1, &mut x2, &mut y2);
         }
-        (x1, y1, x2, y2)
+        self.status().map(|_| (x1, y1, x2, y2))
     }
 
-    pub fn in_fill(&self, x: f64, y: f64) -> bool {
-        unsafe { ffi::cairo_in_fill(self.0.as_ptr(), x, y).as_bool() }
+    pub fn in_fill(&self, x: f64, y: f64) -> Result<bool, Error> {
```
sdroege, line 400: Same question here

cairo/src/context.rs
```diff
@@ -435,19 +445,22 @@ impl Context {
...
-    pub fn copy_page(&self) {
-        unsafe { ffi::cairo_copy_page(self.0.as_ptr()) }
+    pub fn copy_page(&self) -> Result<(), Error> {
+        unsafe { ffi::cairo_copy_page(self.0.as_ptr()) };
+        self.status()
     }
 
-    pub fn show_page(&self) {
-        unsafe { ffi::cairo_show_page(self.0.as_ptr()) }
+    pub fn show_page(&self) -> Result<(), Error> {
```
sdroege, line 461: And here

cairo/src/paths.rs
```diff
@@ -132,7 +132,7 @@ mod tests {
     fn make_cr() -> Context {
         let surface = ImageSurface::create(Format::Rgb24, 1, 1).unwrap();
 
-        Context::new(&surface)
+        Context::new(&surface).expect("Can't create a Cairo context")
```
sdroege, line 135: For a future reviewer: these panics are all in test code :)

sdroege (commented):
Looks generally good to me, but apart from the ones I commented on there are also a few other functions that seem like things that shouldn't really be able to fail (by breaking the context/surface status) as they only retrieve information.

cairo/src/context.rs
bilelmoussaoui: Because if the the context is in an error state, the output wouldn't make much sense

sdroege: Hm ok!

cairo/src/context.rs
federicomenaquintero: If you have something like

```rust
cr.move_to(...);
cr.line_to(...);
cr.curve_to(...);
// big, complex path gets formed
// maybe cairo OOMs or hopefully detects out-of-range coordinates or whatever

// The result is meaningless if the cr went into an error state above.
// Also maybe if x/y are out of range, cairo could detect it and return an error.
let hit_test = cr.in_fill(x, y)?;
```

So this one and the following are about catching meaningless results in case your code didn't notice before that the cr had gone into an error state.

cairo/src/context.rs
federicomenaquintero: For PDF/PNG/etc. that are backed by streams, this is where the code gets to know if there was an I/O error (unless you specify a write_func and store an error flag there).

sdroege:
@federicomenaquintero Thanks for the explanations. How do we proceed with this? From my side this seems fine. @bilelmoussaoui ?

GuillaumeGomez (approved):

bilelmoussaoui:
> @federicomenaquintero Thanks for the explanations. How do we proceed with this? From my side this seems fine. @bilelmoussaoui ?

Let me rebase it and cleanup the commits a bit, but other than that it looks good to me. I will start a WIP PR on librsvg side as soon as this one gets merged

sdroege (approved):
Ok, let me know! From my side this is all good.

bilelmoussaoui:
@sdroege let's get this one merged :)

GuillaumeGomez:
Ok!

PR #500, opened 2021-05-01
gtk: Add run_future() for native dialogs

sophie-h:


sdroege:
@bilelmoussaoui Should probably be also done for gtk4 then

sdroege (approved):

gtk/src/native_dialog.rs
```diff
@@ -0,0 +1,60 @@
...
+    fn run_future<'a>(&'a self) -> Pin<Box<dyn Future<Output = ResponseType> + 'a>> {
+        Box::pin(async move {
+            let (sender, receiver) = futures_channel::oneshot::channel();
+
+            let sender = Cell::new(Some(sender));
+
+            let response_handler = self.connect_response(move |_, response_type| {
+                if let Some(m) = sender.replace(None) {
+                    let _result = m.send(response_type);
+                }
```
bilelmoussaoui, line 45: Shouldn't the `dialog.destroy()` be called here instead of hiding the dialog later on?

sophie-h: The `.hide()` is probably redundant because the window seems to be hidden by opening a file or canceling it.

`.destroy()` would imply that the `dialog.file()` in the example could fail, right? But you want to get information out of the dialog after response.

gtk/src/native_dialog.rs
```diff
@@ -0,0 +1,60 @@
...
+    /// use gtk::prelude::*;
+    ///
+    /// # async fn run() {
+    /// let dialog = gtk::FileChooserNativeBuilder::new()
+    ///    .title("Select a File")
+    ///    .build();
+    ///
+    /// dialog.run_future().await;
+    /// dialog.hide();
+    /// println!("Selected file: {:?}", dialog.file());
```
bilelmoussaoui, line 25:
```suggestion
    /// println!("Selected file: {:?}", dialog.file());
    /// dialog.destroy();
```
there's no point on hiding the dialog i guess, so let's provide a more realistic example

sophie-h: Yes, this should be removed!

sophie-h:
I probably need to manually squash the commits because this is not GitLab?

sdroege (approved):

sdroege:
> I probably need to manually squash the commits because this is not GitLab?

There's a "squash & merge" button, let's see if that does what it promises :)

PR #503, opened 2021-05-04
gio: manually implement Display for File

pbor:


sdroege (commented):
This is unfortunately incoherent. If you'd `Display` a `gio::File` as a `glib::Object` or ` gio::File` or a `gio::LocalFile` the results would be different.

Not sure if that's a good idea here.

pbor:
I am not sure I follow... what is the benefit of printing an object and just see "File"?

If I print a file I surely would like to see the path/uri and parse name is exactly this: "This is generally used to show the GFile as a nice full-pathname kind of string in a user interface"

sdroege:
> I am not sure I follow... what is the benefit of printing an object and just see "File"?

Nothing, I don't think any of these things should have a `Display` impl but others disagreed with that.

> If I print a file I surely would like to see the path/uri and parse name is exactly this: "This is generally used to show the GFile as a nice full-pathname kind of string in a user interface"

Yes, but you'd just get `Object` if you did the same on `file.upcast_ref::<glib::Object>()` for example. That's why it's incoherent.

PR #504, opened 2021-05-04
glib: bind [freeze|thaw]_notify

bilelmoussaoui:


glib/src/object.rs
```diff
@@ -1760,6 +1762,14 @@ impl<T: ObjectType> ObjectExt for T {
         }
     }
 
+    fn freeze_notify(&self) {
```
GuillaumeGomez, line 1765: Missing the doc aliases. ;)

bilelmoussaoui: oh good catch

GuillaumeGomez (approved):

GuillaumeGomez:
Might be worth it to send a PR later on to add the missing C aliases (there are a lot).

bilelmoussaoui:
> Might be worth it to send a PR later on to add the missing C aliases (there are a lot).

yeah i started adding them within this PR and gave up on it for now :-) maybe some other day

GuillaumeGomez:
Yes don't worry, let's keep this PR focused on your change. We can make PRs for that later on. I'm gonna open an issue so it's not forgotten.

bilelmoussaoui:
> Yes don't worry, let's keep this PR focused on your change. We can make PRs for that later on. I'm gonna open an issue so it's not forgotten.

that's definitely a good idea, there are also a bunch of missing doc aliases on manual cairo code as well

glib/src/object.rs
```diff
@@ -1372,6 +1372,13 @@ pub trait ObjectExt: ObjectType {
         f: F,
     ) -> SignalHandlerId;
     fn notify<'a, N: Into<&'a str>>(&self, property_name: N);
+
+    #[doc(alias = "g_object_freeze_notify")]
+    fn freeze_notify(&self);
```
sdroege, line 1377: It would be better to return a guard object here that `thaw`s on `Drop`. Otherwise it's easy to forget to do that in all return paths.

glib/src/object.rs
```diff
@@ -1073,6 +1073,14 @@ glib_object_wrapper!(@object
 );
 pub type ObjectClass = Class<Object>;
 
+pub struct ObjectGuard<'a, T: ObjectType>(&'a T);
```
sdroege, line 1076:
```suggestion
#[must_use = "if unused the property notifications will immediately be thawed"]
pub struct PropertyNotificationFreezeGuard<'a, T: ObjectType>(&'a T);
```

Also not sure if this should take a reference or clone the object. There are probably cases where you want to e.g. send such a thing through a channel or keep it during async operations until they're completely done. A reference would make that all a bit tricky.

glib/src/object.rs
```diff
@@ -1372,6 +1380,13 @@ pub trait ObjectExt: ObjectType {
         f: F,
     ) -> SignalHandlerId;
     fn notify<'a, N: Into<&'a str>>(&self, property_name: N);
+
+    #[doc(alias = "g_object_freeze_notify")]
+    fn freeze_notify(&self) -> ObjectGuard<Self>;
+
+    #[doc(alias = "g_object_thaw_notify")]
+    fn thaw_notify(&self);
```
sdroege, line 1388: Not needed anymore with that guard.

glib/src/object.rs
sdroege: Also you can make it non-generic by storing a `glib::Object` internally. That keeps things a bit simpler too.

sdroege:
@bilelmoussaoui What's with this? :)

GuillaumeGomez:
Needs to be re-open on gtk-rs-core.

Issue #502, opened 2021-05-03
gio:File::parse_name conflict


pbor:
In C we have `get_parse_name()` and `parse_name()` constructor.

As far as I can see in rust both are called `parse_name` one must specify the trait to disambiguate

sdroege:
The constructor should really be `from_parse_name()`. CC @fengalin @bilelmoussaoui

pbor:
In general I would agree... but `parse` is a verb so the whole api is kind of weird.

Another possibility is to drop the getter entirely and implement `Display`.
Implementing `Display` is a good idea on its own, so in the mean time I opened https://github.com/gtk-rs/gtk-rs/pull/503

sdroege:
In `GFile` there's a concept of a "parse name", so I wouldn't consider it as a verb here.

fengalin:
> The constructor should really be `from_parse_name()`.

I agree, this is a `File` constructor, not a `Path` name parser.

While we're at it, we should also probably rename other constructors:

* `new_for_commandline_arg`.
* `new_for_commandline_arg_and_cwd`.
* `new_for_path`.
* `new_for_uri`.

... either to `for_*` or `from_*`. Kind of weird they didn't get caught by the [constructor auto-renaming](https://github.com/gtk-rs/gir/blob/b5ff6f8dd8713c3a8d4864f2f975374f0612976b/src/analysis/functions.rs#L558) stuff btw.

PR #505, opened 2021-05-04
Fix gio file constructors names

fengalin:
WARNING: don't merge as is. Need to update the gir commit.

Fixes https://github.com/gtk-rs/gtk-rs/issues/502

Depends on https://github.com/gtk-rs/gir/pull/1126.


gio/src/auto/file.rs
```diff
@@ -56,7 +57,8 @@ impl File {
     }
 
     #[doc(alias = "g_file_new_for_commandline_arg_and_cwd")]
-    pub fn new_for_commandline_arg_and_cwd<P: AsRef<std::ffi::OsStr>, Q: AsRef<std::path::Path>>(
```
sdroege, line 59: Should maybe autodetect `new_for` like we do `new_from`?

sdroege:
The results look good to me though.

gio/src/auto/file.rs
fengalin: We already do https://github.com/gtk-rs/gir/blob/b5ff6f8dd8713c3a8d4864f2f975374f0612976b/src/analysis/functions.rs#L558

But that applies only to constructors.

sdroege: Oh I see!

bilelmoussaoui (approved):

bilelmoussaoui:
There are still few others in DBusProxy/DBusConnection/UnixMountEntry that are prefixed with `new_for`, I wonder if it makes sense to handle them here?

fengalin:
`grep -rn new_for **/src/ | wc -l` returns `82` on `gtk-rs`...

bilelmoussaoui:
> `grep -rn new_for **/src/ | wc -l` returns `82` on `gtk-rs`...

most of those are from -sys no?

fengalin:
> most of those are from -sys no?

No, I grepped on `**/src`. I mean you are right, but I thought I was through this. I'll see what I can do later.

sdroege:
> I'll see what I can do later.

Let me know if you want this to be merged first, or do it all at once :)

fengalin:
Some are a bit tricky:

| pkg | struct | funct | comment | proposal |
| ---- | ------- | ------- | ----------- | -------------- |
| `gdk-pixbuf` | `PixbufAnimation` | `new_from_stream_async` | Takes a closure which receives a `Result`, no ret.  | `try_for_stream_async`? |
| '' | '' | `new_from_stream_async_future` | Ret. a `Future` which resolves to a `Result`  | `try_from_stream_async_future`? |
| `gio` | `DbusConnection` | `new_for_address` | Takes a closure which receives a `Result`, no ret.  | `try_for_address`? |
| '' | '' | `new_for_address_future` | Ret. a `Future` which resolves to a `Result`  | `try_for_address_future`? |
| '' | `DBusProxy` | `new_for_bus` | Takes a closure which receives a `Result`, no ret.  | `try_for_bus`? |
| '' | '' | `new_for_bus_future` | Ret. a `Future` which resolves to a `Result`  | `try_for_bus_future`? |
| '' | `Icon` | `new_for_string` | Ret. a `Result`  | `try_for_string`? |
| '' | `UnixMountEntry` | `new_for` | Arg. is a `Path`, Ret. a tuple `(Option<UnixMountEntry>, u64)`  | `try_for`? |
| `pangocairo` | `FontMap` | `new_for_font_type` | Ret. an `Option`  | `try_for_font_type`? |

Note: `gio::DBusConnection::for_address_sync`, which is marked "constructor" and so automatically renamed, also returns a `Result`. We might want `gir` to automatically pick the `try` prefix, not sure if it will be straight forward though...

sdroege:
We don't use the `try_` prefix in autogenerated code for constructors either, so if we start doing that here this also needs some gir changes.

> `UnixMountEntry`

`::try_for_path()` maybe?

fengalin:
> We don't use the try_ prefix in autogenerated code for constructors either, so if we start doing that here this also needs some gir changes.

Ok, I'm not willing to go through this now, so I'll just keep `for_` and `from_`.

> ::try_for_path() maybe?

Ok.

sdroege:
> Ok, I'm not willing to go through this now, so I'll just keep `for_` and `from_`.

Yeah it's all rabbit holes :) Can you create an issue about that so we don't forget?

fengalin:
> Yeah it's all rabbit holes :)

Yeah, even without going through this, I still need another `gir` change :)

> Can you create an issue about that so we don't forget?

https://github.com/gtk-rs/gir/issues/1129

About `UnixMountEntry`, I didn't realized there was also:

```rust
fn new_at<P: AsRef<std::path::Path>>(mount_path: P) -> (Option<UnixMountEntry>, u64)
```

So we might want to keep this one as is for consistency:

```rust
fn new_for<P: AsRef<std::path::Path>>(file_path: P) -> (Option<UnixMountEntry>, u64)
```

What do you think?

sdroege:
> About `UnixMountEntry`, I didn't realize there was also:

I think the C API is badly named :) That should've been `new_for_mount_path()` and `new_for_file_path()` in C because otherwise it's just confusing.

fengalin:
Ok, that was the last issue I raise about these. I'll deliberately ignore any other inconsistencies for now on :)

sdroege:
So let's get this in as-is and worry about the others some other time?

fengalin:
Let me push what I changed and then we'll be all good for now :)

sdroege (approved):

fengalin:
* Depends on https://github.com/gtk-rs/gir/pull/1130
* I'll update the "gir update" commit before we merge.

fengalin:
Ready to merge! :)
Thanks.

PR #508, opened 2021-05-04
glib: fix some nullables

sophie-h:
relevant c code

```c
      if (!e.user_name)
        e.user_name = g_strdup ("somebody");
      if (!e.real_name)
        e.real_name = g_strdup ("Unknown");
```

```c
  /* If we have been denied access to /etc/passwd (for example, by an
   * overly-zealous LSM), make up a junk value. The return value at this
   * point is explicitly documented as ‘undefined’. */
  if (home_dir == NULL)
    {
      g_warning ("Could not find home directory: $HOME is not set, and "
                 "user database could not be read.");
      home_dir = g_strdup ("/");
    }

```

and the rest according to docs and `auto/unix_mount_entry.rs`

### DIFF
--- a/gio/src/unix_mount_entry.rs
+++ b/gio/src/unix_mount_entry.rs
@@ -61,7 +61,7 @@ impl UnixMountEntry {
 
     #[doc(alias = "g_unix_mount_get_device_path")]
     #[doc(alias = "get_device_path")]
-    pub fn device_path(&self) -> Option<std::path::PathBuf> {
+    pub fn device_path(&self) -> std::path::PathBuf {
         unsafe {
             from_glib_none(ffi::g_unix_mount_get_device_path(mut_override(
                 self.to_glib_none().0,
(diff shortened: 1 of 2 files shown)


gio/src/unix_mount_entry.rs
```diff
@@ -61,7 +61,7 @@ impl UnixMountEntry {
 
     #[doc(alias = "g_unix_mount_get_device_path")]
     #[doc(alias = "get_device_path")]
-    pub fn device_path(&self) -> Option<std::path::PathBuf> {
+    pub fn device_path(&self) -> std::path::PathBuf {
```
sdroege, line 64: Why are all these manually written? For `UnixMountPoint` (which is more or less the same API!) almost all these functions are auto-generated.

Can you check? That would automatically do the right thing with regards to nullability also.

sdroege (commented):
The changes themselves look good to me in any case.

gio/src/unix_mount_entry.rs
sophie-h: Does the gir matching work via function names? The functions are called `g_unix_mount_point_*()` (correct) and `g_unix_mount_*()` (incorrect). I didn't found a way to manually add functions to a object in gir or to assign functions to an object. When generated naively the `g_unix_mount_*()` end up in `functions`.

sdroege: Uh, ok. So that's the reason why :( Not sure why that happens, it looks correct to me in the gir. Can you create an issue about that, and then let's get this one in?

PR #509, opened 2021-05-04
remove the deprecated functions/reduce the number of disabled clippy linters

bilelmoussaoui:


gtk/src/auto/enums.rs
```diff
@@ -7686,116 +7506,6 @@ impl ToValue for StackTransitionType {
     }
 }
 
-#[deprecated = "Since 3.14"]
```
sdroege, line 7689: Should maybe update the minimum supported version before this release... 3.14 is from 2014. That's ancient :) Maybe 3.20 would be a good next step, that's 2016 so would still cover Ubuntu 16.04 for example.

@GuillaumeGomez @bilelmoussaoui What do you think? That would be a separate PR of course.

graphene/src/frustum.rs
```diff
@@ -10,7 +10,7 @@ impl Frustum {
     #[doc(alias = "get_planes")]
     pub fn planes(&self) -> [Plane; 6] {
         unsafe {
-            let mut out: [ffi::graphene_plane_t; 6] = std::mem::uninitialized();
+            let mut out: [ffi::graphene_plane_t; 6] = std::mem::zeroed();
```
sdroege, line 13: Yeah it's a bit annoying that `MaybeUninit` doesn't work nicely for arrays yet.

gdk/src/lib.rs
```diff
@@ -3,7 +3,6 @@
 #![cfg_attr(feature = "dox", feature(doc_cfg))]
 #![allow(clippy::missing_safety_doc)]
 #![allow(clippy::upper_case_acronyms)]
-#![allow(deprecated)]
```
sdroege, line 6: Because of this we get many warnings like

```
use of deprecated static `auto::constants::STYLE_REGION_TAB`: Since 3.20
```

now. Which is of course correct but also... why would it warn about that inside the crate that defines these things? :) Somehow you need to keep the deprecated API alive, and that requires using it inside the crate.

What should we should do about that?

sdroege (commented):
Looks good to me otherwise, nice cleanup!

graphene/src/box_.rs
```diff
@@ -10,7 +10,7 @@ impl Box {
     #[doc(alias = "get_vertices")]
     pub fn vertices(&self) -> [Vec3; 8] {
         unsafe {
-            let mut out: [ffi::graphene_vec3_t; 8] = std::mem::uninitialized();
+            let mut out: [ffi::graphene_vec3_t; 8] = std::mem::zeroed();
```
GuillaumeGomez, line 13: Considering that we "fill" the array just below, is there any point in filling it with zeroes? Performance wise, it doesn't seem like a good idea...

gtk/src/auto/enums.rs
GuillaumeGomez: Sounds good to me.

graphene/src/box_.rs
sdroege: It's not UB like before :)

GuillaumeGomez: The content is all initialized, but then we call a function which might put random data, so back to square one. In my opinion, this is not an improvement in this context.

sdroege: There's nothing about opinions here. The old code is UB because you are not allowed to get a reference to uninitialized memory. The new code is fine because it initializes it with zeroes (which is a valid value in this case!) before creating a reference that is passed to the C function as a pointer.

What the C functions does or doesn't do is irrelevant in this context.

GuillaumeGomez: Now that I think about it: why aren't we using `MaybeUninit` here?

gdk/src/lib.rs
bilelmoussaoui: That's annoying indeed, but only happens when generating docs it seems. Maybe a rustdoc issue/behaviour?

sdroege: Ok :)

graphene/src/box_.rs
GuillaumeGomez: Ok nevermind, I thought it was an array of structs whereas it's an array of pointers.

gtk/src/auto/enums.rs
bilelmoussaoui: No opinions on that, but seems like everyone has 3.20 nowadays

GuillaumeGomez (approved):

PR #512, opened 2021-05-05
Update minimum versions to versions available in Ubuntu 16.04

sdroege:


sdroege:
@GuillaumeGomez @bilelmoussaoui This is now ready. Please check the manual commits if I forget something or removed too much :)

glib/tests/log.rs
```diff
@@ -67,7 +67,6 @@ fn check_log_handlers() {
     //
     // log_set_handler check part
     //
-    #[cfg(feature = "v2_46")]
     {
```
bilelmoussaoui, line 70: nitpick: I suppose there's no need for the extra {} around the test here?

pango/Gir.toml
```diff
@@ -33,8 +33,6 @@ generate = [
     "Pango.CoverageLevel",
     "Pango.Direction",
     "Pango.EllipsizeMode",
-    "Pango.EngineLang",
```
bilelmoussaoui, line 36: Shouldn't those be moved to the ignore list? otherwise gir -m not_bound will complain about them i guess?

pango/src/auto/font.rs
```diff
@@ -34,8 +33,8 @@ pub trait FontExt: 'static {
     #[doc(alias = "pango_font_describe_with_absolute_size")]
     fn describe_with_absolute_size(&self) -> Option<FontDescription>;
 
-    #[doc(alias = "pango_font_find_shaper")]
-    fn find_shaper(&self, language: &Language, ch: u32) -> Option<EngineShape>;
+    //#[doc(alias = "pango_font_find_shaper")]
+    //fn find_shaper(&self, language: &Language, ch: u32) -> /*Ignored*/Option<EngineShape>;
```
bilelmoussaoui, line 37: Heh, how those functions were not deprecated as well

sdroege: You'd automatically get warnings if you used them (also in C), so maybe just laziness or was forgotten :)

PR #513, opened 2021-05-06
glib-macros: do not export ::type_() for boxed types

pbor:
The proper way is to call ::static_type()

### DIFF
--- a/glib-macros/src/gboxed_derive.rs
+++ b/glib-macros/src/gboxed_derive.rs
@@ -1,8 +1,9 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
+use heck::SnakeCase;
 use proc_macro2::{Ident, TokenStream};
 use proc_macro_error::abort_call_site;
-use quote::quote;
+use quote::{format_ident, quote};
 
 use crate::utils::{crate_ident_new, find_attribute_meta, find_nested_meta, parse_type_name};
 
@@ -94,6 +95,7 @@ pub fn impl_gboxed(input: &syn::DeriveInput) -> TokenStream {
             e
         ),
     };
+    let get_type = format_ident!("{}_type", name.to_string().to_snake_case());
 
     let crate_ident = crate_ident_new();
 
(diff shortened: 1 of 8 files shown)


glib/src/subclass/boxed.rs
```diff
@@ -17,14 +17,6 @@ pub trait BoxedType: Clone + Sized + 'static {
     ///
```
sdroege, line 17: Needs to become `pub trait BoxedType: StaticType + Clone + Sized + 'static {`

glib/src/subclass/shared.rs
```diff
@@ -76,14 +76,6 @@ pub trait SharedType: Clone + Sized + 'static {
     /// The inner refcounted type
```
sdroege, line 76: And also here needs `StaticType` to be added to the trait bounds

sdroege (commented):
Looks good to me otherwise, but also needs some more tests to be updated.

sdroege (approved):

glib-macros/src/gboxed_derive.rs
```diff
@@ -116,28 +118,11 @@ pub fn impl_gboxed(input: &syn::DeriveInput) -> TokenStream {
...
-                    assert!(TYPE_.is_valid());
-                    TYPE_
-                }
-            }
         }
 
         impl #crate_ident::StaticType for #name {
             fn static_type() -> #crate_ident::Type {
-                <#name as #crate_ident::subclass::boxed::BoxedType>::type_()
+                #get_type()
```
bilelmoussaoui, line 125: Any reason why we have to call `#get_type()` here instead of just having the function content here? is it useful for something else?

sdroege: That's how it's done in all the other derive macros... but good point. Let's just inline it here

sdroege: (and in the other macros)

pbor: I blatantly copied from GEnum :)

sdroege (approved):

PR #514, opened 2021-05-07
glib: Make use of g_source_set_dispose_function() in 2.64+ to fix a r…

sdroege:
…ace condition during MainContext channel finalization

It is possible for a channel sender to get access to the source while it
is being finalized, which then can cause critical warnings or worse
memory corruption.

Solving this cleanly is not possible in older versions of GLib.

----

CC @pbor

### DIFF
--- a/glib/src/main_context_channel.rs
+++ b/glib/src/main_context_channel.rs
@@ -249,13 +249,41 @@ unsafe extern "C" fn dispatch<T, F: FnMut(T) -> Continue + 'static>(
     ffi::G_SOURCE_CONTINUE
 }
 
+#[cfg(feature = "v2_64")]
+unsafe extern "C" fn dispose<T, F: FnMut(T) -> Continue + 'static>(source: *mut ffi::GSource) {
+    let source = &mut *(source as *mut ChannelSource<T, F>);
+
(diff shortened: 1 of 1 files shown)


pbor:
I confirm I do not repro with this patch

glib/src/main_context_channel.rs
```diff
@@ -249,13 +249,41 @@ unsafe extern "C" fn dispatch<T, F: FnMut(T) -> Continue + 'static>(
...
 
+    #[cfg(feature = "v2_64")]
+    {
+        let _ = source.channel.take().expect("Receiver without channel");
+    }
+    #[cfg(not(feature = "v2_64"))]
     {
+        let channel = source.channel.take().expect("Receiver without channel");
+
+        // FIXME: This is the same as would otherwise be done in the dispose() function but
```
pbor, line 280: Does marking this as `FIXME` make sense? It is not going to be fixed... that is why in new glib we have the proper fix :)

sdroege: Yes, when we have newer glib as minimum version we can fix it be removing the old code :) That was my thought here.

PR #516, opened 2021-05-09
glib: add VariantTy::any()

pbor:


glib/src/variant_type.rs
```diff
@@ -160,6 +160,12 @@ impl VariantTy {
         }
     }
 
+    /// Returns an indefinite type that is a supertype of every type (including itself).
+    #[doc(alias = "G_VARIANT_TYPE_ANY")]
+    pub fn any() -> &'static VariantTy {
+        Self::new("*").unwrap()
```
sdroege, line 166:
```suggestion
        &*("*".as_bytes() as *const [u8] as *const VariantTy)
```

or so. No need to do checking if the string is a valid variant type. And then you can also make this an actual constant instead of a function.

pbor: is making it and actual const really better? given that for other cases we use `::static_variant_type()` etc that are all methods

sdroege: See e.g.

```rust
impl StaticVariantType for bool {
    fn static_variant_type() -> Cow<'static, VariantTy> {
        unsafe { VariantTy::from_str_unchecked("b").into() }
    }
}
```

This returns basically a `&'static VariantTy`. It only does allocations/etc for more complicated types like containers.

So having it as a constant would make sense IMHO

sdroege (approved):

bilelmoussaoui (approved):

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #518, opened 2021-05-09
Add a variant prop to the example

pbor:
This is an example that shows the problems with Option<Variant>.

Not sure if we want to add it or if you want to keep the example simple.

### DIFF
--- a/glib/src/subclass/mod.rs
+++ b/glib/src/subclass/mod.rs
@@ -13,6 +13,7 @@
 //! use glib::prelude::*;
 //! use glib::subclass;
 //! use glib::subclass::prelude::*;
+//! use glib::{Variant, VariantType};
 //!
 //! use std::cell::{Cell, RefCell};
 //!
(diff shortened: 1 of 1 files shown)


glib/src/subclass/mod.rs
```diff
@@ -87,11 +90,7 @@
 //!         // Called every time a new instance is created. This should return
 //!         // a new instance of our type with its basic values.
 //!         fn new() -> Self {
```
bilelmoussaoui, line 92: you could just drop the new method as well, it defaults to calling `Self::default()` already, but that might make it unclear how all of that works

glib/src/subclass/mod.rs
```diff
@@ -125,6 +124,14 @@
 //!                         MyFlags::default().bits(),
 //!                         glib::ParamFlags::READWRITE,
 //!                     ),
+//!                     glib::ParamSpec::new_variant(
+//!                         "secondary-action-target",
+//!                         "secondary-action-target",
+//!                         "secondary-action-target",
```
sdroege, line 130: Best to provide proper strings for the 3 arguments here :)

sdroege (approved):

PR #529, opened 2021-05-14
gdk: Fixes for keymap/keysym API

ids1024:


gdk/src/keys.rs
```diff
@@ -79,6 +92,19 @@ impl Key {
         skip_assert_initialized!();
         unsafe { ffi::gdk_keyval_to_lower(**self) }.into()
     }
+
+    #[doc(alias = "gdk_keyval_convert_case")]
+    pub fn keyval_convert_case(&self) -> (Self, Self) {
```
bilelmoussaoui, line 97: Maybe rename this one to just convert_case?

ids1024: Ah right, forgot to rename that.

gdk/Gir.toml
```diff
@@ -183,6 +183,34 @@ status = "generate"
...
+    #manual
+    ignore = true
+    [[object.function]]
+    name = "keyval_convert_case"
+    #manual
+    ignore = true
+    [[object.function]]
+    name = "keyval_is_lower"
+    #manual
+    ignore = true
```
sdroege, line 200:
```suggestion
    manual = true
```

etc

gdk/src/keys.rs
```diff
@@ -44,6 +45,18 @@ impl IntoGlib for Key {
 }
 
 impl Key {
+    #[doc(alias = "gdk_keyval_from_name")]
+    pub fn from_name(keyval_name: &str) -> Self {
```
sdroege, line 49: This can't fail?

gdk/src/keys.rs
```diff
@@ -44,6 +45,18 @@ impl IntoGlib for Key {
...
+        skip_assert_initialized!();
+        Self(unsafe { ffi::gdk_keyval_from_name(keyval_name.to_glib_none().0) })
+    }
+
+    #[doc(alias = "gdk_unicode_to_keyval")]
+    pub fn from_unicode(c: char) -> Self {
```
sdroege, line 55: This can't fail?

ids1024: It returns "the corresponding GDK key symbol, if one exists. or, if there is no corresponding symbol, wc | 0x01000000". I'm not really sure how that is supposed to make sense, but I guess it doesn't count as a "failure" and is a valid "keyval"?

gdk/src/keys.rs
ids1024: The documentation says it returns "`GDK_KEY_VoidSymbol` if the key name is not a valid key". So I guess not a failure? This is different from `gdk_keymap_lookup_key` which returns `0` if no mapping exists.

sdroege: Strange API, but ok :)

PR #537, opened 2021-05-18
gtk: Use MainContext::is_owner() to check if it's the owner instead o…

sdroege:
…f acquiring without releasing


sdroege:
https://github.com/gtk-rs/gtk3-rs/pull/536

PR #539, opened 2021-05-19
Add missing doc aliases

GuillaumeGomez:


gdk/src/device.rs
```diff
@@ -26,6 +26,7 @@ impl Device {
 
     #[doc(alias = "gdk_device_get_history")]
     #[doc(alias = "get_history")]
+    #[doc(alias = "gdk_device_free_history")]
```
bilelmoussaoui, line 29:
```suggestion
```
not needed

gtk/src/drag_context.rs
```diff
@@ -24,6 +24,7 @@ pub trait DragContextExtManual: 'static {
     fn drag_set_icon_name(&self, icon_name: &str, hot_x: i32, hot_y: i32);
 
     #[doc(alias = "gtk_drag_set_icon_pixubf")]
+    #[doc(alias = "gtk_drag_set_icon_pixbuf")]
```
bilelmoussaoui, line 27: hmm i can't tell the difference with the existing doc alias here?

gtk/src/rt.rs
```diff
@@ -103,6 +103,7 @@ pub unsafe fn set_initialized() {
 /// so will not cause the program to terminate if GTK could not be initialized.
 /// Instead, an Ok is returned if the windowing system was successfully
 /// initialized otherwise an Err is returned.
+#[doc(alias = "gtk_init_check")]
```
bilelmoussaoui, line 106: not needed, maybe make it  `gtk_init` as it's the right one to use?

gtk/src/drag_context.rs
GuillaumeGomez: I can't either, I guess it's a bug. :)

gtk/src/rt.rs
GuillaumeGomez: I'll make a replacement for it.

gtk/src/drag_context.rs
GuillaumeGomez: Oh I see it now! The `u` and `b` are not in the right place. That's pretty funny. :)

GuillaumeGomez:
Updated!

gtk/src/drag_context.rs
bilelmoussaoui: lol

bilelmoussaoui (approved):

PR #548, opened 2021-05-25
gir fixes

bilelmoussaoui:
As part of https://github.com/gtk-rs/gir/pull/1161, the docs are now based on analysis rather than the symbols list. With that, in order to avoid broken links we must mark missing(not automagically implemented ones) as such to avoid the analysis claiming a function exist and we end up with a broken link.

Do not merge until the gir PR is merged, there are still a bunch of other functions that needs to be handled properly and a ton of broken markdown in gtk3's gir file makes noticing those a bit harder.

Will get back to it in the future. But so far, most of the links are properly working :) no more breakage!


atk/src/attribute.rs
```diff
@@ -4,6 +4,17 @@ use glib::translate::*;
 use glib::GString;
 use std::fmt;
 
+/// AtkAttribute is a string name/value pair representing a generic
```
GuillaumeGomez, line 7: You'll need to add the rustdoc-stripper annotation, otherwise it'll get removed.

bilelmoussaoui: I didn't document it myself, it was always there :) I don't plan to fix gtk3, just the broken docs links

GuillaumeGomez: The diff tells otherwise.

bilelmoussaoui: probably caused by embed-docs, i will the changes there and regen.

sdroege:
@bilelmoussaoui What's the status here?

bilelmoussaoui:
I don't really plan to rebase this one, if someone wants to take over what I did here; feel free to do so

sdroege:
Something for @GuillaumeGomez I guess

PR #550, opened 2021-05-28
Ignore tmp emacs files

GuillaumeGomez:


sdroege (approved):

sdroege:
Same is also needed for gtk4-rs

GuillaumeGomez:
gtk4 is already the right version, unless I'm missing something?

GuillaumeGomez:
Oups never mind. Not the version update PR. :laughing:

PR #553, opened 2021-05-30
CI docs: use gir-rustdoc

bilelmoussaoui:
This generates a main page along with a popover to point to latest docs if the user is reading an old/git version of it

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -27,6 +27,8 @@ jobs:
         run: cargo build --release
       - run: cargo install rustdoc-stripper
       - run: python3 ./generator.py --embed-docs --yes ./
+      - run: git clone https://gitlab.gnome.org/World/Rust/gir-rustdoc/ # checkout action doesn't support random urls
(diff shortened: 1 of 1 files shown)


.github/workflows/docs.yml
```diff
@@ -27,6 +27,8 @@ jobs:
         run: cargo build --release
       - run: cargo install rustdoc-stripper
       - run: python3 ./generator.py --embed-docs --yes ./
+      - run: git clone https://gitlab.gnome.org/World/Rust/gir-rustdoc/ # checkout action doesn't support random urls
```
GuillaumeGomez, line 30: Would have make maybe more sense to put it in the release script? Well, not a blocker in any case.

GuillaumeGomez (approved):

.github/workflows/docs.yml
bilelmoussaoui: No, because it's used by gstreamer and in the future all the crates in World/Rust

GuillaumeGomez: Oh I see. Well then all good. :)

PR #561, opened 2021-06-03
Fix git URLs

GuillaumeGomez:


atk/README.md
```diff
@@ -1,4 +1,5 @@
 # atk 
+
 [Project site](https://gtk-rs.org/) | [Online documentation](https://gtk-rs.org/docs-src/)
 
 __Rust__ bindings and wrappers for __Atk__, part of [gtk-rs](https://github.com/gtk-rs/gtk3-rs).
```
bilelmoussaoui, line 5: would be nice to update those as well :)

GuillaumeGomez: Done.

bilelmoussaoui (approved):

PR #563, opened 2021-06-05
Add minimal crate info

sophie-h:
Also adds minimum required version of the library to every crate doc.

### DIFF
--- a/atk/src/lib.rs
+++ b/atk/src/lib.rs
@@ -1,10 +1,11 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
-//! # ATK bindings
+//! # Rust ATK bindings
 //!
-//! This library contains safe Rust bindings for [ATK](https://developer.gnome.org/atk/). It's
-//! a part of [Gtk-rs](https://gtk-rs.org/).
-
+//! This library contains safe Rust bindings for [ATK](https://developer.gnome.org/atk).
+//! It is a part of [Gtk-rs](https://gtk-rs.org/).
+//!
+//! ATK 2.18 is the lowest supported version for the underlying library.
 #![cfg_attr(feature = "dox", feature(doc_cfg))]
 
 pub use ffi;
--- a/gdk/src/lib.rs
+++ b/gdk/src/lib.rs
@@ -4,6 +4,13 @@
 #![allow(clippy::missing_safety_doc)]
 #![allow(clippy::upper_case_acronyms)]
 
+//! # Rust GDK 3 bindings
+//!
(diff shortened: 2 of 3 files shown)


atk/src/lib.rs
```diff
@@ -1,10 +1,11 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
-//! # ATK bindings
+//! # Rust ATK bindings
 //!
-//! This library contains safe Rust bindings for [ATK](https://developer.gnome.org/atk/). It's
-//! a part of [Gtk-rs](https://gtk-rs.org/).
-
+//! This library contains safe Rust bindings for [ATK](https://developer.gnome.org/atk).
+//! It is a part of [Gtk-rs](https://gtk-rs.org/).
```
bilelmoussaoui, line 6: Maybe it should mention it's part of gtk3-rs bindings instead of gtk-rs?

gdk/src/lib.rs
```diff
@@ -4,6 +4,13 @@
 #![allow(clippy::missing_safety_doc)]
 #![allow(clippy::upper_case_acronyms)]
 
+//! # Rust GDK 3 bindings
```
GuillaumeGomez, line 7: I think it needs to be before the other attributes.

GuillaumeGomez: Ah no, nevermind!

GuillaumeGomez (approved):

atk/src/lib.rs
GuillaumeGomez: We are talking about the organization there, not the repository (hence the URL).

bilelmoussaoui: sure, what i meant is it's maybe worth it to mention that atk is part of the specific gtk3-rs bindings, which are part of gtk-rs.org.

GuillaumeGomez: Not sure if this is worth it to mention it in the documentation. What do you think @sophie-h @sdroege ?

bilelmoussaoui: then what mentioning that atk is part of gtk-rs.org brings to the docs? you will already be reading them from gtk-rs.org/gtk3-rs/ 
but it's just a suggestion, probably doesn't worth the change, feel free to ignore it

sophie-h: I think there is too much room for improvement for all these missing introductory docs to start tweaking them now. Just wanted to have two sentences and a header everywhere.

bilelmoussaoui: Indeed, never mind my comment :-)

bilelmoussaoui (approved):

PR #571, opened 2021-06-11
CI: set gtk-rs logo for rust docs

bilelmoussaoui:


.github/workflows/docs.yml
```diff
@@ -34,7 +34,16 @@ jobs:
...
-      - run: echo ${{ env.DEST }}
+
+      - name: Grab gtk-rs LOGO
+        if: ${{ (github.event_name == 'push' && github.ref == 'refs/heads/master') || github.event_name == 'release' }}
+        run: |
+          sudo dnf install ImageMagick wget -y
+          wget https://raw.githubusercontent.com/gtk-rs/gtk-rs.github.io/master/logo.svg
+          convert -density 1200 -resize 256x256 -background none logo.svg ./target/doc/rust-logo.png
+          for size in 16 32 ; do convert -density 1200 -resize "${size}x${size}" -background none logo.svg "./target/doc/favicon-${size}x${size}.png" ; done
+          cp logo.svg ./target/doc/favicon.svg
```
GuillaumeGomez, line 45: Don't we already have a favicon that we can use instead of doing all this?

.github/workflows/docs.yml
```diff
@@ -34,7 +34,16 @@ jobs:
...
           args: -p atk -p atk-sys -p gdk -p gdk-sys -p gdkx11 -p gdkx11-sys -p gtk -p gtk3-macros -p gtk-sys --features dox --no-deps
       - run: echo "DEST=$(if [ "$GITHUB_EVENT_NAME" == "release" ]; then echo 'stable/${{ github.event.release.tag_name }}'; else echo 'git'; fi)" >> ${GITHUB_ENV}
-      - run: echo ${{ env.DEST }}
+
+      - name: Grab gtk-rs LOGO
+        if: ${{ (github.event_name == 'push' && github.ref == 'refs/heads/master') || github.event_name == 'release' }}
+        run: |
+          sudo dnf install ImageMagick wget -y
+          wget https://raw.githubusercontent.com/gtk-rs/gtk-rs.github.io/master/logo.svg
+          convert -density 1200 -resize 256x256 -background none logo.svg ./target/doc/rust-logo.png
```
GuillaumeGomez, line 43: Also: you can set an URL for the logo too, no need to go through all that...

bilelmoussaoui: That doesn't work for neither the index page nor the -sys crates :-)

.github/workflows/docs.yml
bilelmoussaoui: We use the the logo.svg as is for the favicon. rustdoc uses that along with a fallback to png ones in case the browser doesn't support SVG favicons

bilelmoussaoui: We could instead generate favicons for those sizes and use them on the website as well. then we can just grab them here but that's unrelated work to this change

GuillaumeGomez: Take a look there: https://doc.rust-lang.org/rustdoc/the-doc-attribute.html#html_favicon_url

bilelmoussaoui: As I said below, this only works for non-sys crates

GuillaumeGomez: Why? We simply need to add a new thing for gir to generate it and it's done.

bilelmoussaoui:
@GuillaumeGomez I updated the PR to use the new favicons. Feel free to squash and merge :)

GuillaumeGomez (approved):

GuillaumeGomez:
Let's see how it goes. Thanks!

PR #572, opened 2021-06-12
CI: use apt instead of dnf

bilelmoussaoui:
because it's an ubuntu based image instead of a fedora one

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -38,7 +38,7 @@ jobs:
       - name: Grab gtk-rs LOGO
         if: ${{ (github.event_name == 'push' && github.ref == 'refs/heads/master') || github.event_name == 'release' }}
         run: |
-          sudo dnf install wget -y
+          apt-get install wget -y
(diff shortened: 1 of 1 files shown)


.github/workflows/docs.yml
```diff
@@ -38,7 +38,7 @@ jobs:
       - name: Grab gtk-rs LOGO
         if: ${{ (github.event_name == 'push' && github.ref == 'refs/heads/master') || github.event_name == 'release' }}
         run: |
-          sudo dnf install wget -y
+          apt-get install wget -y
```
GuillaumeGomez, line 41: No `sudo`? Also: didn't pay attention to that...

bilelmoussaoui: Somehow the ubuntu image doesn't require one because you're sudoer by default

sdroege (approved):

.github/workflows/docs.yml
sdroege: Also `apt` instead of `apt-get` is the modern way, but doesn't make a difference really

bilelmoussaoui:
do not merge until #573 is merged and the image got updated

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #576, opened 2021-06-12
Fix icon-size property type issue

### LINKED ISSUE #568: [BUG] Regression after https://github.com/gtk-rs/gtk3-rs/pull/538
**Bug description**

Property `icon-size` of type `gtk::Image` has wrong type `IconSize` while it should be `i32`.

[gtk3](https://developer.gnome.org/gtk3/stable/GtkImage.html#GtkImage--icon-size)
[gtk4](https://developer.gnome.org/gtk4/stable/GtkImage.html#GtkImage--icon-size)

Creation panics like this:

```
Failed to create an instance of Image: BoolError { message: "property \'icon-size\' of type \'GtkImage\' can\'t be set from the given type (expected: \'gint\', got: \'GtkIconSize\')", filename: "/home/.../gtk-rs/gtk-rs-core/glib/src/object.rs", function: "glib::object", line: 2280 }
```

GuillaumeGomez:
Fixes #568.


bilelmoussaoui:
I suppose you also need to replace the other functions that take an IconSize with a manual implementation

bilelmoussaoui:
see https://github.com/gtk-rs/gtk3-rs/pull/538/files

GuillaumeGomez:
Updated!

gtk/src/stack_switcher.rs
```diff
@@ -0,0 +1,510 @@
...
+    #[doc(alias = "icon-size")]
+    fn connect_icon_size_notify<F: Fn(&Self) + 'static>(&self, f: F) -> SignalHandlerId;
+}
+
+#[cfg(any(feature = "v3_20", feature = "dox"))]
+#[cfg_attr(feature = "dox", doc(cfg(feature = "v3_20")))]
+impl<O: IsA<StackSwitcher>> StackSwitcherExtManual for O {
+    fn icon_size(&self) -> IconSize {
+        unsafe {
+            let mut value = glib::Value::from_type(<IconSize as StaticType>::static_type());
```
sdroege, line 51:
```suggestion
            let mut value = glib::Value::from_type(i32::static_type());
```

Or not?

gtk/src/image.rs
```diff
@@ -0,0 +1,492 @@
...
+    fn set_icon_size(&self, icon_size: IconSize);
+
+    #[doc(alias = "icon-size")]
+    fn connect_icon_size_notify<F: Fn(&Self) + 'static>(&self, f: F) -> SignalHandlerId;
+}
+
+impl<O: IsA<Image>> ImageExtManual for O {
+    fn icon_size(&self) -> IconSize {
+        unsafe {
+            let mut value = glib::Value::from_type(<IconSize as StaticType>::static_type());
```
sdroege, line 30:
```suggestion
            let mut value = glib::Value::from_type(i32::static_type());
```

GuillaumeGomez:
Updated!

PR #595, opened 2021-06-24
regen with latest gir

bilelmoussaoui:


bilelmoussaoui:
this is needed because i finally fixed the docs of 0.14 and we need the index page to be updated now.

GuillaumeGomez (approved):

PR #614, opened 2021-07-31
Acquire and leak main context in gtk::init()

ArekPiekarz:
Fixes partially gtk-rs/gtk-rs-core#186, because gtk4 still has that issue.

### DIFF
--- a/gtk/src/rt.rs
+++ b/gtk/src/rt.rs
@@ -117,6 +117,14 @@ pub fn init() -> Result<(), glib::BoolError> {
         let argv = ::std::env::args().take(1).collect::<Vec<_>>();
 
         if from_glib(ffi::gtk_init_check(&mut 1, &mut argv.to_glib_none().0)) {
+            // See https://github.com/gtk-rs/gtk-rs-core/issues/186 for reasoning behind
+            // acquiring and leaking the main context here.
+            let result: bool = from_glib(glib::ffi::g_main_context_acquire(
+                glib::ffi::g_main_context_default(),
+            ));
+            if !result {
+                return Err(glib::bool_error!("Failed to acquire default main context"));
+            }
             set_initialized();
             Ok(())
         } else {


ArekPiekarz:
The reason for the failed job seems to be unrelated to my change:
```
2021-07-31T15:59:46.8746626Z meson.build:1:0: ERROR: Value "false" (of type "string") for combo option "Enable GObject Introspection (depends on GObject)" is not one of the choices. Possible choices are (as string): "enabled", "disabled", "auto".
2021-07-31T15:59:46.8747658Z 
2021-07-31T15:59:46.8748628Z A full log can be found at C:\gtk-build\build\x64\release\graphene\_gvsbuild-meson\meson-logs\meson-log.txt
2021-07-31T15:59:46.9338617Z Traceback (most recent call last):
2021-07-31T15:59:46.9339631Z    same file (C:\gtk-build\build\x64\release\gettext\build\win32\vs15\gettextlib.vcxproj.filters)
2021-07-31T15:59:46.9340859Z   File "D:\a\gtk3-rs\gtk3-rs\gvsbuild\gvsbuild\utils\builder.py", line 492, in build
2021-07-31T15:59:46.9341530Z     if self.__build_one(p):
2021-07-31T15:59:46.9342584Z   File "D:\a\gtk3-rs\gtk3-rs\gvsbuild\gvsbuild\utils\builder.py", line 618, in __build_one
2021-07-31T15:59:46.9343378Z     skip_deps = proj.build()
2021-07-31T15:59:46.9344038Z   File "D:\a\gtk3-rs\gtk3-rs\gvsbuild\gvsbuild\projects.py", line 587, in build
2021-07-31T15:59:46.9344749Z     Meson.build(self, make_tests=True)
2021-07-31T15:59:46.9365782Z   File "D:\a\gtk3-rs\gtk3-rs\gvsbuild\gvsbuild\utils\base_builders.py", line 66, in build
2021-07-31T15:59:46.9366689Z     self.exec_vs(cmd, add_path=add_path)
2021-07-31T15:59:46.9367442Z   File "D:\a\gtk3-rs\gtk3-rs\gvsbuild\gvsbuild\utils\base_project.py", line 110, in exec_vs
2021-07-31T15:59:46.9368369Z     self.builder.exec_vs(cmd, working_dir=self._get_working_dir(), add_path=add_path)
2021-07-31T15:59:46.9369272Z   File "D:\a\gtk3-rs\gtk3-rs\gvsbuild\gvsbuild\utils\builder.py", line 865, in exec_vs
2021-07-31T15:59:46.9370168Z     self.__execute(self.__sub_vars(cmd), working_dir=working_dir, add_path=add_path, env=self.vs_env)
2021-07-31T15:59:46.9371192Z   File "D:\a\gtk3-rs\gtk3-rs\gvsbuild\gvsbuild\utils\builder.py", line 942, in __execute
2021-07-31T15:59:46.9372049Z     subprocess.check_call(args, cwd=working_dir, env=env, shell=True)
2021-07-31T15:59:46.9372969Z   File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\subprocess.py", line 363, in check_call
2021-07-31T15:59:46.9373802Z     raise CalledProcessError(retcode, cmd)
2021-07-31T15:59:46.9377061Z subprocess.CalledProcessError: Command 'C:\gtk-build\tools\python.3.7.7\tools\python.exe C:\gtk-build\tools\meson-fd6585ecc41e1c731224cac08a3f6ef6f49ab184\meson.py C:\gtk-build\build\x64\release\graphene C:\gtk-build\build\x64\release\graphene\_gvsbuild-meson --prefix C:\gtk-build\gtk\x64\release -Dbenchmarks=false -Dintrospection=false --buildtype debugoptimized' returned non-zero exit status 1.
2021-07-31T15:59:46.9379516Z Error: graphene build failed
```

gtk/src/rt.rs
```diff
@@ -117,6 +117,12 @@ pub fn init() -> Result<(), glib::BoolError> {
         let argv = ::std::env::args().take(1).collect::<Vec<_>>();
 
         if from_glib(ffi::gtk_init_check(&mut 1, &mut argv.to_glib_none().0)) {
+            let result: bool = from_glib(glib::ffi::g_main_context_acquire(
+                glib::ffi::g_main_context_default(),
```
sdroege, line 121:
```suggestion
                ptr::null_mut(),
```

Would also work

gtk/src/rt.rs
```diff
@@ -117,6 +117,12 @@ pub fn init() -> Result<(), glib::BoolError> {
         let argv = ::std::env::args().take(1).collect::<Vec<_>>();
 
         if from_glib(ffi::gtk_init_check(&mut 1, &mut argv.to_glib_none().0)) {
+            let result: bool = from_glib(glib::ffi::g_main_context_acquire(
```
sdroege, line 120: Can you add a comment here *why* this is done?

sdroege (commented):

sdroege:
> The reason for the failed job seems to be unrelated to my change:

This probably disappears on a retry... or needs to be reported to graphene.

gtk/src/rt.rs
ArekPiekarz: I tried to base that code on what `glib::MainContext::acquire` does. It's true that your version would work, but in my opinion it would be not obvious what the null means in this case. From the signature of the function one could see it's a context, but that alone doesn't say what are the semantics of acquiring a null context. The documentation of the function also doesn't mention it. I had to check the sources in GLib repository to confirm it behaves as you said. So I would prefer to leave it as is or create a named constant for that null.

gtk/src/rt.rs
ArekPiekarz: Would a short description like this suffice?
```
# See https://github.com/gtk-rs/gtk-rs-core/issues/186 for reasoning behind acquiring and leaking the main context here.
```
Or do you want a longer one?

gtk/src/rt.rs
sdroege: Ok with me to keep it as it is :)

gtk/src/rt.rs
ArekPiekarz: Done in https://github.com/gtk-rs/gtk3-rs/commit/7f36de09e73e4103a916b3d62f7f0af56131ab70

sdroege (approved):

sdroege:
@ArekPiekarz Thanks, you also do the same for GTK4?

ArekPiekarz:
@sdroege Yes, I will try to do that.

ArekPiekarz:
FYI now the Windows build failed for a different reason, again probably unrelated to this change:
```
[577/989] Generating test.gresource with a custom command (wrapped by meson to set PATH)
FAILED: gio/tests/test.gresource 
"C:/gtk-build/tools/python.3.7.7/tools/python.exe" "C:/gtk-build/tools/meson-fd6585ecc41e1c731224cac08a3f6ef6f49ab184/meson.py" "--internal" "exe" "--unpickle" "C:/gtk-build/build/x64/release/glib/_gvsbuild-meson/meson-private/meson_exe_glib-compile-resources_c9c94695beae3266dd669148af439d87e97211d7.dat"
../gio/tests/test.gresource.xml: Failed to execute helper program (No such file or directory).
while executing ['C:\\gtk-build\\build\\x64\\release\\glib\\_gvsbuild-meson\\gio\\glib-compile-resources.exe', '--target=gio/tests/test.gresource', '--sourcedir=C:/gtk-build/build/x64/release/glib/gio/tests', '--sourcedir=C:/gtk-build/build/x64/release/glib/_gvsbuild-meson/gio/tests', '../gio/tests/test.gresource.xml']
```

GuillaumeGomez:
Just want to confirm one thing: can we still get the context if we want (after calling `init` I mean)?

ArekPiekarz:
@GuillaumeGomez Yes, I tested it with a following small program:
```
use gtk::glib;

const DEFAULT_CONTEXT : Option<&glib::MainContext> = None;

fn main() {
    gtk::init().unwrap();
    let (_sender, receiver) = glib::MainContext::channel(glib::PRIORITY_DEFAULT);
    let handler = move |event: i32| {
        println!("got event: {}", event);
        glib::Continue(true)
    };
    receiver.attach(DEFAULT_CONTEXT, handler);  // panicked before this change
}
```
When we go into definition of `Receiver::attach()`, it calls `MainContext::ref_thread_default()` that retrieves the default context, asserts that it is its owner, then uses it.

GuillaumeGomez:
Thanks for the explanations! Would it be worth it to add such a test as well so we don't have a regression again?

ArekPiekarz:
@GuillaumeGomez Good idea, I added a unit test for that.

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

ArekPiekarz:
To save you time, Windows build failed again on graphene and "CI / gtk (rust nightly)" on linking Cairo:
```
note: /usr/bin/ld: /__w/gtk3-rs/gtk3-rs/target/debug/deps/libcairo-5e44ed6c27c39d69.rlib(cairo-5e44ed6c27c39d69.cairo.3157f652-cgu.0.rcgu.o): undefined reference to symbol 'cairo_gobject_extend_get_type'
```

GuillaumeGomez:
Not surprised, for whatever reason, windows CI is broken. Just waiting for @sdroege to take a look.

sdroege:
> @GuillaumeGomez Good idea, I added a unit test for that.

Actually a bad idea. Not only does it fail on the CI now (because `-lcairo-1.0 -lcairo-gobject-1.0` is missing for some reason) but also will it fail in various ways later (because you can initialize only from one thread per process but all tests are running from different threads).

@GuillaumeGomez If you insist on a test, how do you want to solve this? :)

GuillaumeGomez:
I guess it's not possible unless we run all tests in one thread... Well, let's remove the test and merge this then.

bilelmoussaoui:
> I guess it's not possible unless we run all tests in one thread... Well, let's remove the test and merge this then.

We have some code for doing so in gtk4 if you want to use that here

bilelmoussaoui:
Would be nice to backport this to gtk4 as well

GuillaumeGomez:
Backport is already scheduled I think. As for the test, please provide the code! :D

ArekPiekarz:
I've been using [rusty-fork](https://crates.io/crates/rusty-fork) crate for isolating tests in my apps, but I guess @bilelmoussaoui had [TEST_THREAD_WORKER](https://github.com/gtk-rs/gtk4-rs/blob/master/gtk4/src/lib.rs#L63) in mind.

sdroege:
Remove the test for now (or comment it out), then we can merge this here and do the test stuff independently.

GuillaumeGomez:
And open an issue in any case. :p

sdroege:
This is now in 0.14.1 on crates.io

PR #621, opened 2021-08-20
Make Clipboard::request_image pixbuf an Option

GuillaumeGomez:
Fixes https://github.com/gtk-rs/gtk3-rs/issues/620.

### DIFF
--- a/gtk/Gir.toml
+++ b/gtk/Gir.toml
@@ -742,6 +742,10 @@ status = "generate"
     name = "set_with_owner"
     #array with size
     ignore = true
+    [[object.function]]
+    name = "request_image"
+    # nullable callback parameter
+    ignore = true
 
 [[object]]
 name = "Gtk.ColorButton"
--- a/gtk/src/clipboard.rs
+++ b/gtk/src/clipboard.rs
@@ -90,4 +90,35 @@ impl Clipboard {
             );
         }
     }
+
+    #[doc(alias = "gtk_clipboard_request_image")]
+    pub fn request_image<P: FnOnce(&Clipboard, Option<&gdk_pixbuf::Pixbuf>) + 'static>(
+        &self,
+        callback: P,
+    ) {
+        unsafe extern "C" fn callback_func<
+            P: FnOnce(&Clipboard, Option<&gdk_pixbuf::Pixbuf>) + 'static,
+        >(
+            clipboard: *mut ffi::GtkClipboard,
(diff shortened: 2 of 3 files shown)


gtk/Gir.toml
```diff
@@ -742,6 +742,10 @@ status = "generate"
     name = "set_with_owner"
     #array with size
     ignore = true
+    [[object.function]]
+    name = "request_image"
+    # nullable callback parameter
+    ignore = true
```
sdroege, line 748: Can't this be configured via the toml file? If not, it can via the  gir file...

Please also send an MR upstream to gtk to fix the annotation :)

gtk/src/clipboard.rs
```diff
@@ -90,4 +90,37 @@ impl Clipboard {
...
+            let callback: Box_<P> = Box_::from_raw(data as *mut _);
+            (*callback)(
+                &clipboard,
+                Option::<gdk_pixbuf::Pixbuf>::from_glib_borrow(pixbuf)
+                    .as_ref()
+                    .as_ref(),
+            );
+        }
+        let callback = Some(callback_func::<P> as _);
+        let super_callback0: Box_<P> = callback_data;
```
sdroege, line 117: Why the 0? :p

sdroege: Also you can reduce the number of lines a bit

sdroege:
@bilelmoussaoui does this also apply to gtk4?

bilelmoussaoui:
> @bilelmoussaoui does this also apply to gtk4?

I don't think so, the API is slightly different and it does return a nullable texture already iirc

GuillaumeGomez:
I opened https://github.com/gtk-rs/gir/issues/1209 to allow that.

gtk/Gir.toml
sdroege: Did you create an MR for gtk?

sdroege (approved):

gtk/Gir.toml
GuillaumeGomez: No...

GuillaumeGomez: I actually don't know what I'm supposed to update there.

sdroege: Add an annotation that the callback parameter is nullable. I usually don't merge any PRs like this until there's an upstream MR or at least an issue.

PR #626, opened 2021-08-31
Don't require &mut self for gdk::Event getters

sdroege:
Fixes https://github.com/gtk-rs/gtk3-rs/issues/625


sdroege:
Also included the trivial gir update so that the diff view is not full of useless rustdoc warnings.

GuillaumeGomez (approved):

PR #627, opened 2021-08-31
0.14 backports

sdroege:


GuillaumeGomez:
Isn't this an API change? Even though the current code would work with it, is it compatible from a server standpoint?

sdroege:
I don't see why not? It's only loosening a constraint

GuillaumeGomez:
Well, I think you know better than me so I'll trust you. :)

GuillaumeGomez (approved):

PR #630, opened 2021-09-03
Nullable callback param

GuillaumeGomez:
Follow-up of https://github.com/gtk-rs/gir/pull/1217.


sdroege:
Did you check if there are more instances of this that can be autogenerated now?

GuillaumeGomez:
I didn't. First I need to fix the issue I created.

GuillaumeGomez:
@sdroege I checked if another `ignore = true` (in gtk) was in the same case but didn't find any.

sdroege (approved):

PR #638, opened 2021-10-03
Add buttons property for FileChooserDialogBuilder

### LINKED ISSUE #632: [FEATURE REQUEST] Consistency between FileChooserDialog and FileChooserDialogBuilder
As far as I can tell, the only way to specify the buttons for a `FileChooserDialog` is to call `FileChooserDialog::new()` (which has a `buttons` argument). `FileChooserDialogBuilder` has no equivalent method.

Similarly, `FileChooserDialogBuilder::use_header_bar()` provides control over using a header bar, and `FileChooserDialog` has no equivalent method.

Should these features be provided by both types?

GuillaumeGomez:
Fixes #632.

As for the header bar, you can get it from [DialogExt::header_bar](https://gtk-rs.org/gtk3-rs/stable/latest/docs/gtk/prelude/trait.DialogExt.html#tymethod.header_bar) so I'm not sure if there is any point in providing methods for it...


gtk/src/file_chooser_dialog.rs
```diff
@@ -94,3 +102,727 @@ impl FileChooserDialog {
...
+        self.show_hidden = Some(show_hidden);
+        self
+    }
+
+    pub fn use_preview_label(mut self, use_preview_label: bool) -> Self {
+        self.use_preview_label = Some(use_preview_label);
+        self
+    }
+
+    pub fn add_button(mut self, title: &str, response_type: ResponseType) -> Self {
```
sdroege, line 824:
```suggestion
    pub fn button(mut self, title: &str, response_type: ResponseType) -> Self {
```

Maybe? But if we do it for this dialog, shouldn't we do it for all the others too for consistency?

bilelmoussaoui: Does it make sense to have this as part of the Builder pattern actually ?

GuillaumeGomez: > Maybe? But if we do it for this dialog, shouldn't we do it for all the others too for consistency?

But then it feels like we're simply replacing the current one... We could replace with `buttons` and set them all at once though...

> Does it make sense to have this as part of the Builder pattern actually ?

You can't add buttons otherwise, so not sure how to do it if not by adding a manual thing...

bilelmoussaoui: Well, you can't set a filter neither for a FileChooser. The builders set properties not wrap function calls.

You can build the object from the builder and then call the methods to add/set the buttons on it I guess ?

GuillaumeGomez: There is no such method, that's the whole issue. You can only set them when creating the dialog apparently.

sdroege: What about https://gtk-rs.org/gtk3-rs/stable/0.14/docs/gtk/prelude/trait.DialogExtManual.html#tymethod.add_buttons and https://gtk-rs.org/gtk3-rs/stable/0.14/docs/gtk/prelude/trait.DialogExt.html#tymethod.add_button ?

GuillaumeGomez: Ah indeed! Then there is no point for this PR. :)

PR #658, opened 2021-10-31
gdk: add helpers for checking which display is being used

### LINKED ISSUE #651: Add helpers for checking which display is being used
Like what is done in https://github.com/gtk-rs/gtk4-rs/pull/534.

GuillaumeGomez:
Fixes #651.


bilelmoussaoui (approved):

gdk/src/display.rs
```diff
@@ -0,0 +1,70 @@
...
+    /// Get the currently used display backend
+    fn backend(&self) -> Backend;
+}
+
+impl<O: IsA<Display>> DisplayExtManual for O {
+    fn backend(&self) -> Backend {
+        match self.as_ref().type_().name() {
+            "GdkWaylandDisplay" => Backend::Wayland,
+            "GdkX11Display" => Backend::X11,
+            "GdkMacosDisplay" => Backend::MacOS,
```
sdroege, line 64:
```suggestion
            "GdkQuartzDisplay" => Backend::MacOS,
```

Please also double-check the others

GuillaumeGomez: Checking from https://gitlab.gnome.org/GNOME/gtk/-/tree/gtk-3-24/gdk, it seems to be the only one (took me a while to understand I was not looking at the right version...).

Issue #101, opened 2019-03-05
gdk wayland interaction


psnszsn:
Is it possible to use the functions from `gdkwayland.h`?
Specifically i would like to use `gdk_wayland_window_get_wl_surface()` and `gdk_wayland_window_set_use_custom_surface()`.

sdroege:
For GTK4 this exists now: https://github.com/gtk-rs/gtk4-rs/tree/master/gdk4-wayland

bilelmoussaoui:
I wrote a very small part of gdk3-wayland manually for a crate I'm working on and I was wondering what would be the best way to upstream it here.

I'm thinking of adding the ffi types to gdk-sys's manual.rs & have a feature flag for enabling the gdk-wayland bindings without an extra separate crate. I only have the `gdk_wayland_window_unexport_handle` & `gdk_wayland_window_export_handle` parts, more could be added manually in the future by interested people.

GuillaumeGomez:
Sounds like a good idea to bind them manually.

sdroege:
This can be closed then? @GuillaumeGomez @bilelmoussaoui

GuillaumeGomez:
No, a lot of items are still missing. I'll add them later.

PR #660, opened 2021-11-03
gdk: bind WaylandWindow manually

bilelmoussaoui:
Part of #101

### DIFF
--- a/gdk/src/wayland/mod.rs
+++ b/gdk/src/wayland/mod.rs
@@ -1,0 +1,4 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+mod wayland_window;
+pub use wayland_window::WaylandWindow;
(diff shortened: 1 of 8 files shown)


bilelmoussaoui:
Btw, the export handle part is something already "well" tested as it was part of my ashpd crate for a while https://github.com/bilelmoussaoui/ashpd/blob/master/src/window_identifier.rs#L291-L340 and it's what motivated my work here, so more people could make use of it hopefully :)

gdk/src/wayland/mod.rs
```diff
@@ -0,0 +1,2 @@
+mod wayland_window;
```
GuillaumeGomez, line 1: You need the header here as well.

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #661, opened 2021-11-03
implement subclassing for scrolledwindow

vhdirk:


sdroege (approved):

gtk/src/subclass/scrolled_window.rs
```diff
@@ -0,0 +1,110 @@
...
+            } else {
+                false
+            }
+        }
+    }
+}
+
+unsafe impl<T: ScrolledWindowImpl> IsSubclassable<T> for ScrolledWindow {
+    fn class_init(class: &mut ::glib::Class<Self>) {
+        <Bin as IsSubclassable<T>>::class_init(class);
```
GuillaumeGomez, line 73: Please use `parent_class_init`.

GuillaumeGomez (changes requested):

gtk/src/subclass/scrolled_window.rs
vhdirk: fixed

sdroege: Indeed, completely forgot that we have nicer things now :)

sdroege (approved):

PR #664, opened 2021-11-07
gtk: Generate Container::[get|set]_propert_child

### LINKED ISSUE #131: missing set|get_child_property on GtkContainer
I'm not sure if this are safe to generate ?

GuillaumeGomez:
Fixes #131.

It was mentioned in #131 that we needed to be careful how we would expose this API in rust. Is there anything special to do?

### DIFF
--- a/gtk/src/auto/container.rs
+++ b/gtk/src/auto/container.rs
@@ -42,6 +42,9 @@ pub trait ContainerExt: 'static {
     //#[doc(alias = "gtk_container_child_get")]
     //fn child_get(&self, child: &impl IsA<Widget>, first_prop_name: &str, : /*Unknown conversion*//*Unimplemented*/Fundamental: VarArgs);
 
+    #[doc(alias = "gtk_container_child_get_property")]
+    fn child_get_property(&self, child: &impl IsA<Widget>, property_name: &str) -> glib::Value;
+
     //#[doc(alias = "gtk_container_child_get_valist")]
     //fn child_get_valist(&self, child: &impl IsA<Widget>, first_property_name: &str, var_args: /*Unknown conversion*//*Unimplemented*/Unsupported);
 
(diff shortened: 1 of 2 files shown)


gtk/src/auto/container.rs
```diff
@@ -175,6 +186,19 @@ impl<O: IsA<Container>> ContainerExt for O {
     //    unsafe { TODO: call ffi:gtk_container_child_get() }
     //}
 
+    fn child_get_property(&self, child: &impl IsA<Widget>, property_name: &str) -> glib::Value {
```
bilelmoussaoui, line 189: Maybe we can make the gir produces code that make use of this once merged instead of going through ffi as well

sdroege: Yes please. Just like your PR for normal properties :)

gtk/src/auto/container.rs
```diff
@@ -203,6 +227,22 @@ impl<O: IsA<Container>> ContainerExt for O {
...
     //}
 
+    fn child_set_property(
+        &self,
+        child: &impl IsA<Widget>,
+        property_name: &str,
+        value: &glib::Value,
+    ) {
+        unsafe {
+            ffi::gtk_container_child_set_property(
```
sdroege, line 237: This needs checking if the property exists, if the value is of the correct type, if the property is writable

gtk/src/auto/container.rs
```diff
@@ -175,6 +186,19 @@ impl<O: IsA<Container>> ContainerExt for O {
     //    unsafe { TODO: call ffi:gtk_container_child_get() }
     //}
 
+    fn child_get_property(&self, child: &impl IsA<Widget>, property_name: &str) -> glib::Value {
+        unsafe {
+            let mut value = glib::Value::uninitialized();
+            ffi::gtk_container_child_get_property(
```
sdroege, line 192: This needs checking if the property exists and is readable. You return garbage from here if that's not the case :)

bilelmoussaoui: This basically needs something like there's in ObjectExt

PR #665, opened 2021-11-07
Add missing wayland types and create new gdkwayland crate

### LINKED ISSUE #101: gdk wayland interaction

GuillaumeGomez:
Fixes #101.


gdk/src/wayland/wayland_device.rs
```diff
@@ -0,0 +1,31 @@
...
+glib::wrapper! {
+    #[doc(alias = "GdkWaylandDevice")]
+    pub struct WaylandDevice(Object<ffi::GdkWaylandDevice>) @extends crate::Device;
+
+    match fn {
+        type_ => || ffi::gdk_wayland_device_get_type(),
+    }
+}
+
+impl WaylandDevice {
```
bilelmoussaoui, line 17: `gdk_wayland_device_get_node_path` is missing. There are also a bunch of other wayland specific functions which could be manually binded as well; see gdk4-wayland

gdk/src/wayland/wayland_display.rs
```diff
@@ -0,0 +1,67 @@
...
+    #[doc(alias = "GdkWaylandDisplay")]
+    pub struct WaylandDisplay(Object<ffi::GdkWaylandDisplay>) @extends crate::Display;
+
+    match fn {
+        type_ => || ffi::gdk_wayland_display_get_type(),
+    }
+}
+
+impl WaylandDisplay {
+    #[doc(alias = "gdk_wayland_selection_clear_targets")]
```
bilelmoussaoui, line 16: where does this one comes from? i don't see it in https://gitlab.gnome.org/GNOME/gtk/-/blob/gtk-3-24/gdk/wayland/gdkwaylanddisplay.h

also lacks two wayland specific getters

gdk/src/wayland/wayland_monitor.rs
```diff
@@ -0,0 +1,10 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+glib::wrapper! {
+    #[doc(alias = "GdkWaylandMonitor")]
+    pub struct WaylandMonitor(Object<ffi::GdkWaylandMonitor>) @extends crate::Monitor;
+
```
bilelmoussaoui, line 6: Could implement `gdk_wayland_monitor_get_wl_output`

gdk/src/wayland/wayland_display.rs
bilelmoussaoui: ah this one is a simple function, not a Display method iirc. same as for the other one in gdkwaylandselection.h header

gdk/src/wayland/wayland_window.rs
```diff
@@ -12,6 +13,18 @@ glib::wrapper! {
 }
 
 impl WaylandWindow {
```
bilelmoussaoui, line 15: This one also has some missing methods

bilelmoussaoui:
Otherwise looks good to me, thanks!

gdk/src/wayland/mod.rs
```diff
@@ -1,4 +1,15 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
+mod wayland_device;
```
sdroege, line 3: Why is all this not in a separate `gdk-wayland` crate?

gdk/src/wayland/wayland_display.rs
```diff
@@ -0,0 +1,67 @@
...
+    pub fn selection_clear_targets(&self, selection: &Atom) {
```
sdroege, line 17: Isn't this a memory managment function, i.e. shouldn't be bound?

gdk/sys/src/manual.rs
```diff
@@ -1,7 +1,24 @@
 #[cfg(any(feature = "wayland", feature = "dox"))]
 use glib_sys::{gconstpointer, GType};
 #[cfg(any(feature = "wayland", feature = "dox"))]
-use libc::{c_char, c_void};
+use libc::{c_char, c_int, c_uint, c_void};
+
+#[repr(C)]
+#[cfg(any(feature = "wayland", feature = "dox"))]
+pub struct GdkWaylandDevice(c_void);
```
sdroege, line 8: This is wrong, see the cairo PR we just merged from Federico

gdk/src/wayland/mod.rs
bilelmoussaoui: Could be moved there yes

gdk/src/wayland/wayland_device.rs
GuillaumeGomez: I don't think so? I looked at the gtk repository directly for all wayland functions and I didn't see any others...

gdk/src/wayland/mod.rs
GuillaumeGomez: Also the same for gtk4 I assume.

bilelmoussaoui: remove how? it can't be auto-generated as there's no gir file for gdk3-wayland.

GuillaumeGomez: From gtk crate and create another crate on the side? That's how I understood @sdroege's comment.

sdroege: Ignore me if what I said makes no sense. But ideally keep it consistent between gtk3/gtk4 :)

GuillaumeGomez: Oh ok. Well then, keeping things here then.

gdk/src/wayland/wayland_monitor.rs
GuillaumeGomez: I only implemented methods returning gobject types. I prefer not to bind functions returning pure wayland types as it would mean that we need to bind them. Or I could just declare a type alias to `c_void` and call it a day?

bilelmoussaoui: Yes, and you use Wayland-rs crate to provide safe wrapper for the Wayland types

GuillaumeGomez: Just to be sure we agree: we don't add dependency on wayland crates and I bind the methods by making them unsafe and returning the pointer. If users want to use it, they can use wayland wrappers to safely handle the returned pointer.

Did I understand you correctly?

bilelmoussaoui: Well, wayland-client already provides safe bindings wrappers for wayland types returned by gdk-wayland. So you would make the ffi functions return a c_void and provide safe wrappers by using wayland-client. That's what i did in gdk4-wayland anyway

gdk/src/wayland/mod.rs
bilelmoussaoui: Well, what @sdroege suggested is to have those as a separate crate even if it's manually binded instead of a sub-module of gdk. We already reserved the gdkwayland crate on crates.io for this

gdk/src/wayland/wayland_device.rs
bilelmoussaoui: https://gitlab.gnome.org/GNOME/gtk/-/blob/gtk-3-24/gdk/wayland/gdkwaylanddevice.h#L58

GuillaumeGomez:
Updated!

sdroege (approved):
You probably want to move `gdkx11` / `gdkwayland` out of the `default-members` of the workspace `Cargo.toml` and only leave them all in the `members`, but otherwise looks all good to me.

bilelmoussaoui:
Why gdkwayland is not part of the CI?

bilelmoussaoui:
I suppose you can disable the checker part https://github.com/gtk-rs/gtk3-rs/blob/87197687cd5e22f22f898da9e455a9f33e2234d6/.github/workflows/CI.yml#L39 for gdkwayland, you can add another config to the matrix for either enabling it or not

bilelmoussaoui (approved):
lgtm

PR #672, opened 2021-11-13
Fix some gir annotations

GuillaumeGomez:
Based on https://github.com/gtk-rs/gtk3-rs/pull/548.

### DIFF
--- a/atk/Gir.toml
+++ b/atk/Gir.toml
@@ -12,7 +12,6 @@ single_version_file = true
 generate = [
     "Atk.Component",
     "Atk.CoordType",
-    "Atk.Document",
     "Atk.EventListener",
     "Atk.EventListenerInit",
     "Atk.FocusHandler",
@@ -68,6 +67,13 @@ status = "generate"
 trait_name = "AtkActionExt" # duplicate with gio
 
 [[object]]
+name = "Atk.Document"
+status = "generate"
+    [[object.function]]
+    name = "get_attributes"
+    ignore = true # TODO: re-enable once AttributeSet is properly handled
+
+[[object]]
 name = "Atk.EditableText"
 status = "generate"
 manual_traits = ["EditableTextExtManual"]
@@ -90,6 +96,9 @@ trait_name = "AtkMiscExt" # duplicate with gtk
 name = "Atk.Object"
 status = "generate"
 trait_name = "AtkObjectExt"
+    [[object.function]]
+    name = "get_attributes"
+    ignore = true # TODO: re-enable once AttributeSet is properly handled
 
 [[object]]
 name = "Atk.Range"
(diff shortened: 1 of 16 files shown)


gtk/src/auto/functions.rs
```diff
@@ -135,11 +135,6 @@ pub fn accelerator_parse(accelerator: &str) -> (u32, gdk::ModifierType) {
     }
 }
 
-//#[doc(alias = "gtk_accelerator_parse_with_keycode")]
-//pub fn accelerator_parse_with_keycode(accelerator: &str, accelerator_codes: Vec<u32>) -> (u32, gdk::ModifierType) {
```
bilelmoussaoui, line 139: you can copy this one from gtk4 if you want :)

atk/Gir.toml
```diff
@@ -67,6 +66,13 @@ name = "Atk.Action"
 status = "generate"
 trait_name = "AtkActionExt" # duplicate with gio
 
+[[object]]
+name = "Atk.Document"
+status = "generate"
+    [[object.function]]
+    name = "get_attributes"
+    ignore = true # TODO: re-enable once AttributeSet is properly handled
```
sdroege, line 74: Why do you ignore them all if they can be implemented? Just leave them around, then at least we are aware that they still need to be bound if someone cares enough (and they should up in the "unbound" mode of gir)

bilelmoussaoui: Because of the docs, it will create broken links otherwise

sdroege: Annoying

bilelmoussaoui: Ideally gir would mark those functions/types as not possible to bound so we can ignore them when generating docs links

bilelmoussaoui: Currently they keep their generate state

gtk/src/auto/functions.rs
GuillaumeGomez: Done, thanks! :)

gtk/src/functions.rs
```diff
@@ -0,0 +1,43 @@
...
+            loop {
+                // zero-terminated array
+                let ret = std::ptr::read(accelerator_codes_ptr.add(i));
+                if ret == 0 {
+                    break;
+                }
+                accelerator_codes.push(ret);
+                i += 1;
+            }
+            glib::ffi::g_free(accelerator_codes_ptr as *mut _);
```
sdroege, line 33: If you feel like it

```rust
let mut len = 0;
if !accelerator_codes_ptr.is_null() {
    while ptr::read(accelerator_codes_ptr.add(len)) != 0 {
       len += 1;
    }
}
let accelerator_codes = glib::GSlice::from_glib_full_num(accelerator_codes_ptr, len);
```

That way you don't have to unnecessarily allocate twice but can re-use the allocation from GTK here.

GuillaumeGomez: :o

sdroege (approved):
Good for me otherwise

gtk/src/functions.rs
sdroege: Well, that's why I started adding `glib::GSlice` and things like that. I have some WIP changes locally for more things and having gir autogenerate that (in other cases, not here, this function is complicated).

GuillaumeGomez: Done!

PR #676, opened 2021-11-16
regenerate with latest gir

bilelmoussaoui:


gtk/src/auto/mod.rs
```diff
@@ -1142,3 +1142,167 @@ pub mod traits {
     pub use super::window::GtkWindowExt;
     pub use super::window_group::WindowGroupExt;
 }
+#[doc(hidden)]
+pub mod builders {
+    pub use super::about_dialog::AboutDialogBuilder;
```
sdroege, line 1147: Shouldn't this be re-exported from `src/lib.rs` like

```rust
pub mod builders {
    pub use foo::ManualBuilder;
    pub use auto::builders::*;
}
```

?

sdroege: Oh you do that here already. This commit is not just regenerating but also doing some manual changes.

sdroege (approved):

GuillaumeGomez (approved):

gtk/src/auto/mod.rs
bilelmoussaoui: Ah yeah, let me quickly split those

PR #682, opened 2021-11-21
gtk3-macros: make the doc example building

GuillaumeGomez:
Reopening of https://github.com/gtk-rs/gtk3-rs/pull/496 which was closed for whatever reasons...


gtk3-macros/src/lib.rs
```diff
@@ -21,38 +21,60 @@ use syn::{parse_macro_input, DeriveInput};
...
 ///
+/// ```no_run
+/// use gtk::prelude::*;
+/// use gtk::glib;
+/// use gtk::subclass::prelude::*;
+///
+/// mod imp {
+///     use gtk::prelude::*;
+///     use gtk::glib;
+///     use gtk::subclass::prelude::*;
```
bilelmoussaoui, line 36:
```suggestion
///     use super::*;
```
maybe?

GuillaumeGomez: Imports don't work this way. :p

You need to use `pub use` in order to be able to use another module's imports. Which is definitely not what we want in here. I originally used `pub use` inside the `imp` module but found it unnecessarily complicated so preferred to duplicated instead.

GuillaumeGomez:
CI passed. \o/

gtk3-macros/src/lib.rs
sdroege: For `super` this actually works. Submodules have full visibility of things in a supermodule

sdroege (commented):
Seems good to me otherwise

gtk3-macros/src/lib.rs
GuillaumeGomez: It doesn't in this case at least, it's the first thing @bilelmoussaoui tried in fact. :)

sdroege: Strange, that's how we do it everywhere in tests and elsewhere. Must be something strange with rustdoc / doctests.

bilelmoussaoui: it works fine in gtk4-rs btw https://github.com/gtk-rs/gtk4-rs/blob/master/gtk4-macros/src/lib.rs#L45. What I tried wasn't  work before, because of the missing `# fn main() {}`

GuillaumeGomez: Oh right, maybe it's working now. Sending another PR then!

PR #691, opened 2021-12-13
gtk/builder: manually implement several methods (custom return codes)

lucab:
This adds a manual implementation for all the GtkBuilder methods
that are using unconventional (non gboolean) return codes.

Closes: https://github.com/gtk-rs/gtk3-rs/issues/690

### DIFF
--- a/gtk/Gir.toml
+++ b/gtk/Gir.toml
@@ -527,6 +527,21 @@ manual_traits = ["BuilderExtManual"]
     #downcast
     ignore = true
     doc_trait_name = "BuilderExtManual"
+    [[object.function]]
+    pattern = "add_from_.+"
+    #non-boolean return code
+    manual = true
+    doc_trait_name = "BuilderExtManual"
+    [[object.function]]
+    pattern = "add_objects_from_.+"
+    #non-boolean return code
+    manual = true
+    doc_trait_name = "BuilderExtManual"
+    [[object.function]]
+    pattern = "extend_with_template"
+    #non-boolean return code
+    manual = true
+    doc_trait_name = "BuilderExtManual"
 
 [[object]]
 name = "Gtk.Button"
--- a/gtk/src/auto/builder.rs
+++ b/gtk/src/auto/builder.rs
@@ -3,7 +3,6 @@
 // DO NOT EDIT
 
 use crate::Application;
-use crate::Widget;
 use glib::object::Cast;
 use glib::object::IsA;
 use glib::signal::connect_raw;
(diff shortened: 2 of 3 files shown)


gtk/Gir.toml
```diff
@@ -527,6 +527,21 @@ manual_traits = ["BuilderExtManual"]
     #downcast
     ignore = true
     doc_trait_name = "BuilderExtManual"
+    [[object.function]]
+    pattern = "add_from_.+"
+    #non-boolean return code
+    ignore = true
```
GuillaumeGomez, line 533: Should be `manual`.

gtk/Gir.toml
```diff
@@ -527,6 +527,21 @@ manual_traits = ["BuilderExtManual"]
...
+    doc_trait_name = "BuilderExtManual"
+    [[object.function]]
+    pattern = "add_objects_from_.+"
+    #non-boolean return code
+    ignore = true
```
GuillaumeGomez, line 538: Should be `manual`.

gtk/Gir.toml
```diff
@@ -527,6 +527,21 @@ manual_traits = ["BuilderExtManual"]
...
+    doc_trait_name = "BuilderExtManual"
+    [[object.function]]
+    pattern = "extend_with_template"
+    #non-boolean return code
+    ignore = true
```
GuillaumeGomez, line 543: Should be `manual`.

GuillaumeGomez (approved):

lucab:
@GuillaumeGomez thanks for the very-very-quick feedback!

sdroege:
Why? You want to change the `gir` code to only work with bools? Looks good to me in that case :)

lucab:
@sdroege yes, in the longer term, possibly.
Not an urgent need as I don't think there is anything broken right now.
But I think the review comments were right, and being stricter on booleans will help ensuring we aren't unknowingly stepping on other strange cases (e.g. FDs as return codes).

sdroege:
Once this is merged this can be made more strict in `gir` or are the more things that need to be fixed first? I'd prefer to have this merged only together with the corresponding `gir` change.

lucab:
I don't have those insights right now, as I haven't yet cooked the `gir` change. But I'm ok in leaving this pending a bit longer until then.

lucab:
`gir` patch at https://github.com/gtk-rs/gir/pull/1287. Bump PRs for the usual suspects (gtk-rs-core, gtk3, gtk4, gstreamer) are all green. So yes I think this is the only manual adjustment needed.

PR #695, opened 2021-12-29
Auto generate with fixed slice of copy types

bilelmoussaoui:
This PR regenerates with the changes from https://github.com/gtk-rs/gir/pull/1296 along with a fix of GdkAtom's ToGlibContainerFromSlice implementation.

Instead of fixing the generated methods manually, the ToGlibContainerFromSlice was implemented for &Atom which doesn't make a lot of sense since it's a copy type.


GuillaumeGomez (approved):

GuillaumeGomez:
Thanks! Please merge once CI pass.

bilelmoussaoui:
> Thanks! Please merge once CI pass.

Can't be merged until the gir change lands

bilelmoussaoui:
Unless I implement the methods manually and then after the gir PR is merged drop them which is meh :/

gdk/src/atom.rs
```diff
@@ -79,13 +79,13 @@ impl<'a> ToGlibPtrMut<'a, *mut ffi::GdkAtom> for Atom {
...
-        Vec<Stash<'a, ffi::GdkAtom, &'a Atom>>,
+        Vec<Stash<'a, ffi::GdkAtom, Atom>>,
         Option<Vec<ffi::GdkAtom>>,
     );
 
-    fn to_glib_none_from_slice(t: &'a [&'a Atom]) -> (*mut ffi::GdkAtom, Self::Storage) {
+    fn to_glib_none_from_slice(t: &'a [Atom]) -> (*mut ffi::GdkAtom, Self::Storage) {
         skip_assert_initialized!();
 
         let v: Vec<_> = t.iter().map(|s| s.to_glib_none()).collect();
```
sdroege, line 91: This seems silly. Can't you just work with `t.as_ptr()` here without copying everything into a new `Vec` first?

sdroege: Would need a `#[repr(transparent)]` on the struct though

sdroege:
Seems good to me otherwise but this should probably come with some changes in `gtk-rs-core` too so that `BoxedInline` types can be converted to an array via `as_ptr()` directly?

sdroege:
And in `gtk-rs-core` this should allow autogenerating more functions in `graphene`, I think.

gdk/src/atom.rs
bilelmoussaoui: I didn't look into the implementation in details, i will give it a try later today

sdroege: We can't do that yet because we always generate a `NULL`-terminated array here. We need different traits, which is on my todo list for after the release already.

sdroege:
@bilelmoussaoui This needs updating to latest gir and should then be good to go

sdroege (approved):

PR #696, opened 2021-12-30
Fix `gdkwayland` pkg name in README.md snippets

Jaakkonen:
The folder name is `gdkwayland` so the `Cargo.toml` snippets should have `gdk-wayland = { git = "https://github.com/gtk-rs/gtk3-rs.git", package = "gdkwayland" }` instead of `gdk-wayland = { git = "https://github.com/gtk-rs/gtk3-rs.git", package = "gdk-wayland" }`.

Maybe a comment noting that the package hasn't been released yet and can only be used by the git style should be added too?

### DIFF
--- a/gdkwayland/README.md
+++ b/gdkwayland/README.md
@@ -13,8 +13,8 @@ Currently, the minimum supported Rust version is `1.56.0`.
 
 ## Documentation
 
- * [Rust API - Stable](https://gtk-rs.org/gtk3-rs/stable/latest/docs/gdk-wayland/)
- * [Rust API - Development](https://gtk-rs.org/gtk3-rs/git/docs/gdk-wayland)
+ * [Rust API - Stable](https://gtk-rs.org/gtk3-rs/stable/latest/docs/gdkwayland/)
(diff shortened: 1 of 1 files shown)


Jaakkonen:
The documentation links do not seem to be working either but I cannot find any working link for those. It looks like the dev docs do not have gdk-wayland but I wonder if that is on purpose.

bilelmoussaoui (approved):

bilelmoussaoui:
> The documentation links do not seem to be working either but I cannot find any working link for those. It looks like the dev docs do not have gdk-wayland but I wonder if that is on purpose.

there is no need to, we are planning to do a release in the upcoming days

gdkwayland/README.md
```diff
@@ -36,7 +36,7 @@ Avoid mixing versioned and git crates like this:
 # This will not compile
 [dependencies]
 gdk-wayland = "0.15"
```
sdroege, line 38: I'd suggest to rename it everywhere to `gdkwayland`. The crate is named without the dash too, so that's how it would end up on crates.io.

PR #698, opened 2022-01-06
regenerate with latest gir-files

bilelmoussaoui:


sdroege (approved):

atk/src/auto/enums.rs
```diff
@@ -1362,6 +1362,8 @@ pub enum StateType {
     HasTooltip,
     #[doc(alias = "ATK_STATE_READ_ONLY")]
     ReadOnly,
+    #[doc(alias = "ATK_STATE_COLLAPSED")]
+    Collapsed,
```
sdroege, line 1366:
```
tests/constant.c: In function 'main':
tests/constant.c:206:27: error: 'ATK_STATE_COLLAPSED' undeclared (first use in this function); did you mean 'ATK_STATE_FOCUSED'?
  206 |     PRINT_CONSTANT((gint) ATK_STATE_COLLAPSED);
      |                           ^~~~~~~~~~~~~~~~~~~
```
etc

This needs a version flag but also the -sys tests will continue failing because they assume that the newest possible version is available.

bilelmoussaoui: Meh, we should really figure out how to use that instead of having our own docker images. It works fine for gtk4 because i keep those updated but it doesn't scale that well

sdroege (approved):

PR #699, opened 2022-01-06
manually implement Container::(get|set)_property

bilelmoussaoui:
It requires a manual implementation to check if the property exists
and is readable/writeable and also of the correct type.
Fixes #664


bilelmoussaoui:
Once this is one is merged, I will land the necessary changes on gir to make uses of it

GuillaumeGomez (approved):

sdroege (approved):

PR #710, opened 2022-01-15
Point to git 0.15 versions of the gtk-rs-core dependencies

sdroege:
This makes sure we're building against the latest 0.15 versions while
still being able to publish to crates.io.


sdroege:
@GuillaumeGomez This is also something to add to your release script

sdroege:
@GuillaumeGomez Also it seems like the release script broke the -sys Cargo.toml indentation (see commit)

bilelmoussaoui (approved):

PR #714, opened 2022-01-16
gtk/{Text,Tree}Iter: derive Debug

fengalin:
So that a `struct` using these types can also auto derive `Debug`.

Another solution could be to add a `Debug` `impl` in manual code,
but since these are iterators, there's not much we can show.

Other `boxed_inline` candidates in `gtk-rs-core` and `gtk3-rs` use manual `Debug` `impl` AFAICT.

### DIFF
--- a/gtk/Gir.toml
+++ b/gtk/Gir.toml
@@ -2276,6 +2276,8 @@ manual_traits = ["TextBufferExtManual"]
 name = "Gtk.TextIter"
 status = "generate"
 boxed_inline = true
+    [[object.derive]]
+    name = "Debug"
     [[object.function]]
     name = "get_attributes"
     manual = true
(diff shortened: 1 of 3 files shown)


GuillaumeGomez (approved):

GuillaumeGomez:
I wonder if we should do the opposite and derive `Debug` by default unless set to false...

fengalin:
> I wonder if we should do the opposite and derive Debug by default unless set to false...

If I refer to `graphen`, which uses the most `boxed_inline`s AFAICT, it makes sense to manually implement. You would either get useless `inner` indirection or require to manually implement anyway to either call the C accessors or use the proper cast (see [`quad.rs`](https://github.com/gtk-rs/gtk-rs-core/blob/4199aff2d678934b2e487944aafb98430d281ed1/graphene/src/quad.rs#L64) for instance).

GuillaumeGomez:
Good point. :)

cc @sdroege

sdroege (approved):

PR #715, opened 2022-01-17
Replace `Foo::from_instance(foo)` with `foo.imp()`

sdroege:


examples/basic_subclass/simple_application/imp.rs
```diff
@@ -29,7 +29,7 @@ impl ApplicationImpl for SimpleApplication {
     /// aksed to present itself.
     fn activate(&self, app: &Self::Type) {
         let app = app.downcast_ref::<super::SimpleApplication>().unwrap();
-        let priv_ = SimpleApplication::from_instance(app);
+        let priv_ = app.imp();
```
bilelmoussaoui, line 32: Would be an opportunity to rename those priv_/self_ to imp as well

sdroege: Ah yes, I was just running sed over all this without thinking about that.

`sed -E -i 's;[A-Za-z0-9:]*::from_instance\(&?([a-z.]*)\);\1.imp\(\);g' **.rs` for future reference btw

GuillaumeGomez (approved):

bilelmoussaoui (approved):

PR #723, opened 2022-01-18
gtk: Set gtk-rs as initialized after the original `GtkApplication::st…

sdroege:
…artup` has returned

Fixes https://github.com/gtk-rs/gtk3-rs/issues/716

----

That old code could've never worked and would cause every application subclass to immediately panic.


GuillaumeGomez (approved):

sdroege:
Regression introduced in cb53101568b1

PR #730, opened 2022-01-30
0.15 backports

sdroege:


gtk/src/auto/header_bar_accessible.rs
```diff
@@ -0,0 +1,196 @@
...
+// rustdoc-stripper-ignore-next
+/// A [builder-pattern] type to construct [`HeaderBarAccessible`] objects.
+///
+/// [builder-pattern]: https://doc.rust-lang.org/1.0.0/style/ownership/builders.html
+#[must_use = "The builder must be built to be used"]
+pub struct HeaderBarAccessibleBuilder {
+    accessible_description: Option<String>,
+    accessible_name: Option<String>,
+    accessible_parent: Option<atk::Object>,
+    //accessible-role: /*Unknown type*/,
```
bilelmoussaoui, line 42: Could probably be generated as well if `atk::RelationType` is added to the manual list

sdroege: Yeah but not as part of the backport, that should happen first on master :)

sdroege: See https://github.com/gtk-rs/gtk3-rs/pull/731

PR #739, opened 2022-03-11
Regenerate gir for GString Deref

jf2048:
Corresponds to https://github.com/gtk-rs/gir/pull/1327


sdroege:
@jf2048 What's the status with this one?

PR #742, opened 2022-04-10
Update gir files

sdroege:
No real change other than some documentation, but things were quite outdated here compared to core/gtk4.


bilelmoussaoui (approved):

sdroege:
@GuillaumeGomez You probably want to port the Windows CI fixes from core to here too

PR #743, opened 2022-04-10
Fix windows CI

GuillaumeGomez:
As nicely suggested by @sdroege, I applied the patch from https://github.com/gtk-rs/gtk3-rs/pull/742.


sdroege (approved):
Yay, it even worked. I expected more GTK-related problems.

PR #744, opened 2022-04-22
subclass support for GtkEntry

pbor:


bilelmoussaoui (commented):
There are a bunch of other vfuncs, I guess you are planning to add those as well?

gtk/src/subclass/entry.rs
```diff
@@ -0,0 +1,83 @@
...
+            }
+        }
+    }
+}
+
+unsafe impl<T: EntryImpl> IsSubclassable<T> for Entry {
+    fn class_init(class: &mut glib::Class<Self>) {
+        Self::parent_class_init::<T>(class);
+
+        if !crate::rt::is_initialized() {
```
GuillaumeGomez, line 55: Isn't it checked in the parent class already?

pbor: I just blindly copied from button :)

pbor:
> There are a bunch of other vfuncs, I guess you are planning to add those as well?

I can add them, but a lot of the vfuncs are just action vfuncs related to moving with the keyboard... do we need them?

I copied from `button.rs` and I see that that one also just has a couple of methods will in C there are many more vfuncs...

gtk/src/subclass/entry.rs
pbor: All the other modules seem to do the same, but it is true that this is checked in widget.rs... Maybe we should drop it everywhere, but in that case it should go in a separate MR

GuillaumeGomez: Sounds like a plan. I'll send a PR for that.

sdroege: Maybe first `git blame` why they were added... there's a reason.

GuillaumeGomez: I was the one adding this check, just didn't think it through: https://github.com/gtk-rs/gtk3-rs/commit/cb53101568b1332acb48494ab5d82588bc04bd77

sdroege: Apparently you didn't :) See https://github.com/gtk-rs/gtk3-rs/issues/543

I think this code is correct and needed.

GuillaumeGomez: The issue is apparently about non-widget items. Oh well...

sdroege: @bilelmoussaoui can probably tell us for sure

bilelmoussaoui: I did that to panic as early as possible

GuillaumeGomez: Shouldn't it need to be before `Self::parent_class_init::<T>(class);` then?

pbor: I see you opened a PR, so I will resolve the conversation here for now.

sdroege (approved):

PR #745, opened 2022-04-22
Check subclass gtk check

GuillaumeGomez:
So in case we want to enforce this checks at all levels, this check enforces it.


sdroege (commented):
Seems fine to me if we just want to add it everywhere instead of deciding case by case. @bilelmoussaoui ?

GuillaumeGomez:
Still remains the question if we should put this check before running it in the parent.

PR #757, opened 2022-06-29
Add binding for Gtk.FileFilterInfo

kelnos:
This enables the `.add_custom()` and `.filter()` functions on `gtk::FileFilter`.

I'm not 100% sure about my implementations of `FromGlibPtr*`, but I've at least tested both a "regular" filter and an "add custom" filter, and they both seem to work (and not crash).


gtk/src/file_filter_info.rs
```diff
@@ -0,0 +1,71 @@
...
+pub struct FileFilterInfo {
+    pub contains: FileFilterFlags,
+    pub filename: Option<String>,
+    pub uri: Option<String>,
+    pub display_name: Option<String>,
+    pub mime_type: Option<String>,
+}
+
+#[doc(hidden)]
+impl From<*const ffi::GtkFileFilterInfo> for FileFilterInfo {
```
sdroege, line 18: This is not safe so shouldn't be marked as such :)

Why don't you implement the `FromGlibPtr*` and `ToGlibPtr*` traits here instead?

gtk/src/file_filter_info.rs
```diff
@@ -0,0 +1,71 @@
...
+
+    #[inline]
+    fn to_glib_none(&'a self) -> Stash<'a, *const ffi::GtkFileFilterInfo, Self> {
+        let contains = self.contains.into_glib();
+        let filename = self.filename.to_glib_none();
+        let uri = self.uri.to_glib_none();
+        let display_name = self.display_name.to_glib_none();
+        let mime_type = self.mime_type.to_glib_none();
+
+        let mut data = Box::new(ffi::GtkFileFilterInfo {
```
sdroege, line 58: No need to box this, it can just live directly inside the stash

gtk/src/file_filter_info.rs
```diff
@@ -0,0 +1,71 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use glib::translate::*;
+use libc::c_char;
+
+use crate::FileFilterFlags;
+
+#[derive(Clone)]
+pub struct FileFilterInfo {
```
sdroege, line 9: Another option would be to make this a `#[repr(transparent)]` struct around the FFI type and add a) a constructor, and b) getter functions that simply borrow the strings as `&str`.

That way you would reduce the number of copies here.

sdroege:
> Hmm, looks like there are clippy failures for code unrelated to my change.

Yes, don't worry about those :) They were introduced by the Rust beta being automatically updated.

gtk/src/file_filter_info.rs
kelnos: Yeah, that's a good point, would probably be much simpler too.

gtk/src/file_filter_info.rs
kelnos: Because I don't entirely know what I'm doing :-P

gtk/src/file_filter_info.rs
sdroege: That's also suboptimal then :)

gtk/src/file_filter_info.rs
kelnos: Having a little trouble here.  For the constructor:
```rust
    pub fn new(
        filename: Option<&str>,
        uri: Option<&str>,
        display_name: Option<&str>,
        mime_type: Option<&str>,
    ) -> Self {
        let mut contains = FileFilterFlags::empty();
        for (value, bit) in [
            (filename, FileFilterFlags::FILENAME),
            (uri, FileFilterFlags::URI),
            (display_name, FileFilterFlags::DISPLAY_NAME),
            (mime_type, FileFilterFlags::MIME_TYPE),
        ] {
            if value.is_some() {
                contains |= bit;
            }
        }

        Self(ffi::GtkFileFilterInfo {
            contains: contains.bits(),
            filename: filename.to_glib_full(),
            uri: uri.to_glib_full(),
            display_name: display_name.to_glib_full(),
            mime_type: mime_type.to_glib_full(),
        })
    }
```
For the struct members, if I use `filename.to_glib_none().0`, I end up with gibberish in them later (which makes sense).  But I'm pretty sure `.to_glib_full()` will cause a memory leak (valgrind seems to agree with me).  My first thought was to implement `Drop` and free them there, but I feel like that would cause problems for any `ffi::GtkFileFilterInfo` structs passed to us from C, which would expect to own those `gchar *` pointers, right?  Or do we end up owning anything that comes through that way?

Sorry, I think I just haven't quite wrapped my head around how ownership is supposed to work.

bilelmoussaoui: Are all those nullable? feels weird

kelnos: Also, for the getters, there seems to be a `from_glib_none()` impl for `Option<String>`, but not for `Option<&str>`.  Am I missing something?  Would be annoying to have to build a `String` for those getters.

kelnos: > Are all those nullable? feels weird

Yes.  The `contains` flags field indicates which are non-null.  I think the idea is that custom filters might not need all fields, so it's more efficient for filters to say what they need, and then for implementations to only provide what's needed.

gtk/src/file_filter_info.rs
sdroege: No problem :) Just ask if you have questions, this is all not entirely obvious

gtk/src/file_filter_info.rs
sdroege: You would use `to_glib_full()` for the strings in the constructor and implement `Drop`. When receiving such a struct from C in case of `FromGlibPtrNone` you would have to `strdup()` all the strings so that `Drop` can free them again.

However as you have a `#[repr(transparent)]` struct you can also borrow the values from C, e.g. in that callback. You can cast a `*const ffi::GtkFileFilterInfo` to `&FileFilterInfo` then via `&*(ptr as *const FileFilterInfo)`.

kelnos: Ah, that makes sense.  I was also looking at how `glib::GString` is implemented, and it looks like it handles the case where it was originally a C object vs. originally a Rust object, and can then presumably be a bit more efficient.  Do you think that's worth it for this?  `GtkFileFilterInfo` structs do seem to get created and dropped in large amounts since often one will get created for each row in a tree model.

(I've realized that `GtkFileFilterInfo` is anunusual case... gtk/glib has very few structs that aren't also GObjects, or aren't at _least_ registered as glib boxed types.  And it appears that none of the other few examples of this have been wrapped yet.  So I feel a little less bad about being fairly clueless here...)

kelnos: Ok, I think I have something that works (pushed), though I don't know if I can reduce string copying further.

One thing I'm having trouble with is that it'd probably be nicer to return borrowed `&str`s from the getters, rather than creating `String`s each time they're called, but I can't seem to figure out how to do that.  For example, I have:

```rust
    pub fn filename(&self) -> Option<String> {
        unsafe { from_glib_none(self.0.filename) }
    }
```

But it doesn't seem like there's an impl of `from_glib_none()` for `&str`, only `String`.  Maybe this isn't really feasible to do, though, since `&str` is fundamentally a borrowed type, and it's not particularly meaningful to create and return those on the fly.  I guess we could return them as `Option<CStr>` without needing any allocation, but maybe that's not worth it, as `CStr` is a bit more of a pain to work with.

kelnos: Argh, also I'm now confused about the `FromGlibPtrBorrow` impl (the generated code appears to need it).  I'm just doing simply `Borrowed::new(FileFilterInfo(*ptr))`, but now won't that cause the `Drop` impl to free someone else's strings when the `Borrowed` (and thus the `FilteFilterInfo`) wrappers get dropped?

sdroege: > [...] and it looks like it handles the case where it was originally a C object vs. originally a Rust object, and can then presumably be a bit more efficient.

I don't think that makes much difference here. When creating one from the Rust side, you will always pass it to C (requires copy because of the strings), whenever you get it from C you usually just borrow it and read the fields but that's it.

So having the constructor create a valid C object (`ffi::GtkFileFilterInfo` with owned C strings that are freed on `Drop`) plus passing a `&FileFilterInfo` into the callback seems the most efficient way here unless I'm missing something.

If you make the `FileFilterInfo` not just a `#[repr(transparent)]` around the FFI struct then you also can't do that cast-and-borrow trick for the callback FWIW.

> But it doesn't seem like there's an impl of from_glib_none() for &str, only String.

You would implement the string getters around `std::ffi::CStr` as otherwise they would still involve a copy of the string. You can find some example of that in the repository :)

kelnos: > You would implement the string getters around std::ffi::CStr as otherwise they would still involve a copy of the string.

Ok, done!  One thing I wasn't sure of was whether to `.expect()` on the result, so that if the string wasn't valid utf-8, it would panic.  I ended up settling on `.ok()` instead, as the fields are nullable and the getters have to return `Option` anyway.  I think I would personally prefer my program not crash due to some invalid data, but, on the other hand, this also silently ignores invalid data, which isn't super great either.

kelnos:
The little test I wrote is ending up causing
```
thread 'file_filter_info::test::custom_filter' panicked at 'GTK may only be used from the main thread.', gtk/src/auto/file_filter.rs:24:9
```
Is there a way to make the tests run on the main thread?  Alternatively I can just delete the test; mainly wrote it as a quick check that I wasn't doing anything weird with memory.

sdroege:
> Is there a way to make the tests run on the main thread? Alternatively I can just delete the test; mainly wrote it as a quick check that I wasn't doing anything weird with memory.

Check for the usage of `crate::TEST_THREAD_WORKER` in the gtk3 crate :)

gtk/src/file_filter_info.rs
sdroege: I would `expect()` here for invalid UTF-8 but of course `NULL` pointers are allowed and should be mapped to `None`. If we somehow get invalid input then that's a big problem and needs to be solved instead of being hidden, just like we do elsewhere for invalid UTF-8.

gtk/src/auto/file_filter.rs
```diff
@@ -34,10 +36,40 @@ impl FileFilter {
...
+        &self,
+        needed: FileFilterFlags,
+        func: P,
+    ) {
+        let func_data: Box_<P> = Box_::new(func);
+        unsafe extern "C" fn func_func<P: Fn(&FileFilterInfo) -> bool + 'static>(
+            filter_info: *const ffi::GtkFileFilterInfo,
+            data: glib::ffi::gpointer,
+        ) -> glib::ffi::gboolean {
+            let filter_info = from_glib_borrow(filter_info);
```
sdroege, line 50:
```suggestion
            let filter_info = &*(filter_info as *const FileFilterInfo);
```

gtk/src/auto/file_filter.rs
```diff
@@ -34,10 +36,40 @@ impl FileFilter {
...
+            let callback: &P = &*(data as *mut _);
```
sdroege, line 51:
```suggestion
            let callback: &P = &*(data as *const _);
```

gtk/src/file_filter_info.rs
```diff
@@ -0,0 +1,176 @@
...
+#[doc(hidden)]
+impl<'a> ToGlibPtr<'a, *const ffi::GtkFileFilterInfo> for FileFilterInfo {
+    type Storage = (ffi::GtkFileFilterInfo, u32, [*const c_char; 4]);
+
+    #[inline]
+    fn to_glib_none(&'a self) -> Stash<'a, *const ffi::GtkFileFilterInfo, Self> {
+        Stash(
+            &self.0,
+            (
+                self.0,
```
sdroege, line 133:
```suggestion
                &self,
```

And the corresponding other chane

gtk/src/file_filter_info.rs
```diff
@@ -0,0 +1,176 @@
...
+                self.0.contains,
+                [
+                    self.0.filename,
+                    self.0.uri,
+                    self.0.display_name,
+                    self.0.mime_type,
+                ],
```
sdroege, line 140: These are just numbers/pointers and keeping them in the `Stash` has no real effect. What will ensure that it stays alive is taking the `&self.0` or even `&self` into the `Stash`.

gtk/src/file_filter_info.rs
```diff
@@ -0,0 +1,176 @@
...
+    fn drop(&mut self) {
+        for ptr in [
+            self.0.filename,
+            self.0.uri,
+            self.0.display_name,
+            self.0.mime_type,
+        ] {
+            if !ptr.is_null() {
+                unsafe {
+                    glib::ffi::g_free(ptr as *mut _);
```
sdroege, line 95: `g_free()` can be called with `NULL` so that check is unneeded

gtk/src/file_filter_info.rs
```diff
@@ -0,0 +1,176 @@
...
+            unsafe { CStr::from_ptr(self.0.display_name).to_str().ok() }
+        }
+    }
+
+    #[doc(alias = "get_mime_type")]
+    pub fn mime_type(&self) -> Option<&str> {
+        if self.0.mime_type.is_null() {
+            None
+        } else {
+            unsafe { CStr::from_ptr(self.0.mime_type).to_str().ok() }
```
sdroege, line 80: The `unsafe` block should also cover the `NULL` check as that `CStr` calls is only OK because you first checked for `NULL` (also elsewhere).

sdroege (commented):
Looks good to me otherwise

gtk/src/auto/file_filter.rs
kelnos: These bits are autogenerated; won't any changes get overwritten the next time someone runs the generator script?

gtk/src/file_filter_info.rs
kelnos: Is that safe?  It appears `g_free()` just calls the system `free()`.  Glibc does promise that passing `NULL` won't do anything, but do we believe that is true for all libcs (like musl, or non-Linux OSes?).

gtk/src/file_filter_info.rs
kelnos: Ah, I see.  My general philosophy is to make the `unsafe` block as small as possible, to show exactly what is actually unsafe (checking if a pointer is null or not is not unsafe).  But I'm fine changing it if that's project policy/style.

gtk/src/file_filter_info.rs
sdroege: C99 (at least) allows passing `NULL` to `free()`.

gtk/src/file_filter_info.rs
sdroege: The `CStr::from_ptr()` is only safe because of the `NULL`-check though. We generally try to keep all unsafe code contained in the same block so that outside the block all Rust invariants are kept.

sdroege:
> Would it be ok if I pushed a PR to backport this to the 0.15 branch as well?

Yes but please check at the same time if there are other things worth backporting too :) Thanks!

kelnos:
> Yes but please check at the same time if there are other things worth backporting too :) Thanks!

Yep, was planning to take care of adding everything i need to to master, and then doing a single backport PR for everything to 0.15.

PR #770, opened 2022-09-04
Don't checkout submodules by default

sdroege:
This makes sure that cargo does not clone and checkout all the
submodules if pointing to this repository as a git dependency.

To checkout the submodules `git submodule update --checkout` can be
used.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -10,7 +10,7 @@ name: docs
 
 jobs:
   docs:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     name: build
     container:
       image: ghcr.io/gtk-rs/gtk3-rs/gtk3:latest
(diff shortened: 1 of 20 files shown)


.github/workflows/docs.yml
```diff
@@ -10,10 +10,8 @@ name: docs
 
 jobs:
   docs:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     name: build
-    container:
-      image: ghcr.io/gtk-rs/gtk3-rs/gtk3:latest
```
sdroege, line 16: This is required for building, but because of https://github.com/actions/checkout/issues/915 it doesn't work correctly

GuillaumeGomez:
It's strange that the docs build failed. An idea @bilelmoussaoui ?

sdroege:
I told you why it failed :P

sdroege:
Ah hm, apparently here it works fine with the different container

GuillaumeGomez:
> I told you why it failed :P

Sorry, just woke up. ^^'

Thanks for the PR!

GuillaumeGomez (approved):

PR #775, opened 2022-10-06
Activate dox feature wherever possible

pentamassiv:
Analogous to: https://github.com/gtk-rs/gtk-rs-core/pull/778

Whenever a crate is published on crates.io, its documentation is automatically built on docs.rs regardless of the documentation field in the Cargo.toml file. On docs.rs it is possible to look at all previous versions. This is why there is value in having working builds on docs.rs. Having broken builds on docs.rs also prevents crates using that library from providing their own documentation on docs.rs. I enabled the dox feature wherever it is available to ensure the docs always build, even if the dependencies are not available.

### DIFF
--- a/atk/Cargo.toml
+++ b/atk/Cargo.toml
@@ -10,9 +10,7 @@ readme = "README.md"
 version = "0.16.0"
 keywords = ["atk", "gtk-rs", "gnome", "accessibility"]
 repository = "https://github.com/gtk-rs/gtk3-rs"
-exclude = [
-    "gir-files/*",
-]
+exclude = ["gir-files/*"]
 edition = "2021"
 rust-version = "1.63"
 
@@ -20,7 +18,7 @@ rust-version = "1.63"
 name = "atk"
 
 [features]
-dox = ["ffi/dox"]
+dox = ["ffi/dox", "glib/dox"]
 v2_30 = ["ffi/v2_30"]
 v2_32 = ["v2_30", "ffi/v2_32"]
 v2_34 = ["v2_32", "ffi/v2_34"]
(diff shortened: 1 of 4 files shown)


GuillaumeGomez (approved):

pentamassiv:
For gtk3 I guess it isn't strictly necessary since I guess all dependencies are available on docs.rs for the builds to succeed. There were some failed builds for gtk4 though so I made the changes for all crates.

GuillaumeGomez:
Oh, interestingly enough, `gir` is messing with the Cargo file.

pentamassiv:
Is that the failed regen check? I was wondering why that failed.

pentamassiv:
https://github.com/gtk-rs/gtk-rs-core/pull/778#discussion_r989132506
I guess I have to teach it to gir instead.

pentamassiv:
The reason why I care is because I sometimes like to go back and forth between versions of the docs and compare the differences. It helps me when I change the version of a dependency. I wanted to do that with the documentation for relm4, but that's not possible because the builds on docs.rs failed. The reason for the failure is that some of it's dependencies such as libadwaita (thats [fixed](https://gitlab.gnome.org/World/Rust/libadwaita-rs/-/merge_requests/50) now) failed to build. I am sure there are other smaller projects out there that use some gtk crates as dependencies and would like to provide documentation. It should be as easy as possible for them to do so. Even if the documentation for gtk crates is not perfect if built on docs.rs, it helps other crates that depend on it.
I understand that it is not your priority, so I decided to work on it myself :)

pentamassiv:
While the newest version of gir activates the dox feature in all dependencies in the Cargo.toml, it does not activate them in the Cargo.toml of the safe wrapper crates. This commit activates the dox feature in the safe wrapper crates too.

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #779, opened 2022-10-18
Regenerate with latest gir submodules

GuillaumeGomez:


atk/sys/Cargo.toml
```diff
@@ -17,7 +17,7 @@ shell-words = "1.0.0"
 tempfile = "3"
 
 [features]
-dox = []
+dox = ["glib/dox", "gobject/dox"]
```
bilelmoussaoui, line 20: Would you mind making sure the dox features are properly set on the safe wrapper crates as well?

bilelmoussaoui (approved):

PR #783, opened 2022-10-19
Update link

GuillaumeGomez:


stable/latest/docs
```diff
@@ -1 +1 @@
-../0.15/docs
```
sdroege, line 1: This is wrong. It should be `../0.16/docs` instead of `..//docs`

GuillaumeGomez: Yes, which is why I mentioned that something is missing in the release description.

GuillaumeGomez:
Fixed the link!

PR #788, opened 2022-11-07
Update gtk crate version to 0.16.1

GuillaumeGomez:


gtk/Cargo.toml
```diff
@@ -7,7 +7,7 @@ name = "gtk"
 homepage = "https://gtk-rs.org/"
 authors = ["The gtk-rs Project Developers"]
 readme = "README.md"
-version = "0.16.0"
+version = "0.16.1"
```
sdroege, line 10: There are more crates in here. Best to update them all in sync (and only upload the ones that have actual changes). That's what we do elsewhere.

GuillaumeGomez: Done!

GuillaumeGomez: Not for the `sys` crates though, should I do it for them too?

sdroege: All

GuillaumeGomez: Done as well!

sdroege:
CI needs fixing but otherwise seems OK

GuillaumeGomez:
Should I fix it before merging this PR?

sdroege:
That would be good :)

PR #791, opened 2022-11-16
Skip init assertion for gdk::set_allowed_backends

nardoor:
To my understanding `gdk::set_allowed_backends()` should be called (when needed) before the `init()`  

https://docs.gtk.org/gdk3/func.set_allowed_backends.html
> This call must happen prior to gdk_display_open(), gtk_init(), gtk_init_with_args() or gtk_init_check() in order to take effect.

The way I added the assertion skip might not be the way intended.
Please let me know.

### DIFF
--- a/gdk/Gir.toml
+++ b/gdk/Gir.toml
@@ -220,6 +220,9 @@ status = "generate"
     [[object.function]]
     name = "init_check"
     ignore = true # TODO: manually implement
+    [[object.function]]
+    name = "set_allowed_backends"
+    assertion = "skip"
 
 [[object]]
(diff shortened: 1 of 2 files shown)


bilelmoussaoui:
Would you mind sending a similar PR for gtk4-rs?

GuillaumeGomez (approved):

nardoor:
> Would you mind sending a similar PR for gtk4-rs?

Sure, https://github.com/gtk-rs/gtk4-rs/pull/1183

sdroege (approved):

PR #792, opened 2022-11-26
Regenerate gir

jf2048:


GuillaumeGomez:
Windows CI is broken apparently. The gvsbuild version is maybe invalid?

jf2048:
It looks like the build.py script was removed from gvsbuild recently, looking into it

jf2048:
New version changes the cli a bit and needs to get installed with pip. It runs after updating the windows vm to get a new python version but now fails with some error when building fontconfig

sdroege:
The gvsbuild CI is broken since quite a while here, I wouldn't block this PR on it. It can be solved separately.

sdroege (approved):

jf2048:
The windows CI will also eventually need to be updated on gtk-rs-core with these changes if we ever bump the versions

sdroege:
Yeah, do you want to submit a PR for that?

PR #794, opened 2022-11-27
gtk: implement From<ResponseType> for Value

jf2048:
This looks like the only one here that needs a manual implementation

### DIFF
--- a/gtk/src/response_type.rs
+++ b/gtk/src/response_type.rs
@@ -137,3 +137,10 @@ impl ToValue for ResponseType {
         Self::static_type()
     }
 }
+
+impl From<ResponseType> for glib::Value {
+    fn from(t: ResponseType) -> Self {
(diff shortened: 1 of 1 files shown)


sdroege (approved):

gtk/src/response_type.rs
```diff
@@ -137,3 +137,9 @@ impl ToValue for ResponseType {
         Self::static_type()
     }
 }
+
+impl From<ResponseType> for glib::Value {
+    fn from(t: ResponseType) -> Self {
+        t.to_value()
```
sdroege, line 143:
```suggestion
        skip_init_asserts!();
        t.to_value()
```

or how the macro was called, I always forget :)

bilelmoussaoui:
A similar PR for gtk4-rs would be appreciated :)

sdroege (approved):

PR #797, opened 2022-12-28
examples: Fix compilation after `gio::SimpleAction` constructor took …

sdroege:
…the state by value

CC @pbor


sdroege:
It doesn't look like anybody makes sure that gtk3-rs continues compiling :)

PR #799, opened 2023-01-23
Regenerate with latest gir

GuillaumeGomez:
Needed by https://github.com/gtk-rs/gtk3-rs/pull/798.


GuillaumeGomez:
Updated with new gir fix and added missing manual types.

gtk/src/image.rs
```diff
@@ -126,134 +125,134 @@ impl ImageBuilder {
     /// Build the [`Image`].
     #[must_use = "Building the object from the builder is usually expensive and is not expected to have side effects"]
     pub fn build(self) -> Image {
-        let mut properties: Vec<(&str, &dyn ToValue)> = vec![];
+        let mut builder = glib::Object::builder::<Image>();
```
sdroege, line 128: This should be a plain wrapper around the `ObjectBuilder` like the autogenerated builders now. Same for the other manual builders

sdroege:
Looks good otherwise

gtk/src/image.rs
GuillaumeGomez: Oh nice!

GuillaumeGomez:
Updated builders.

sdroege (approved):

PR #806, opened 2023-02-08
Update crates version

GuillaumeGomez:


atk/Cargo.toml
```diff
@@ -31,7 +31,7 @@ features = ["dox"]
 libc = "0.2"
 bitflags = "1.0"
 ffi = { package = "atk-sys", path = "sys" }
-glib = { git = "https://github.com/gtk-rs/gtk-rs-core" }
+glib = "0.17"
```
sdroege, line 34: You want to keep git

```suggestion
glib = { git =  "https://github.com/gtk-rs/gtk-rs-core", branch = "0.17", version = "0.17" }
```

etc

GuillaumeGomez: Oh? Ok.

sdroege: @bilelmoussaoui also missed that in gtk4

GuillaumeGomez: Fixed. :)

sdroege (approved):

PR #816, opened 2023-03-14
Update to bitflags 2.0

sdroege:


GuillaumeGomez (approved):

sdroege:
Why did you merge this if the CI fails? :)

GuillaumeGomez:
Because the changes look good and it won't work until gtk-rs-core is merged and I'll forget about it in the meantime. :)

sdroege:
Yes but we need to regenerate this with latest gir so that it actually compiles again.

GuillaumeGomez:
Ah, my bad. Went too quickly.

PR #818, opened 2023-03-22
gtk3-macros: enable default features of syn

decathorpe:
The code in this crate uses features of "syn" which are only present
when the default features are enabled. This previously worked "by
accident" because other crates in the dependency tree pulled in the
"default" feature of syn, but these crates have started moving to
syn v2, so this is no longer the case.


GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #828, opened 2023-07-11
Upgrade syn to 2.0

jplatte:
There's also a `NestedMeta` API in syn 2.0, but it's callback based for some reason and I'm much more familiar with the custom parsing API. The error messages should be of similar quality to before, on average.


GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #840, opened 2023-08-12
Regen

GuillaumeGomez:


atk/sys/src/lib.rs
```diff
@@ -1819,6 +1824,11 @@ extern "C" {
     //=========================================================================
     pub fn atk_layer_get_type() -> GType;
 
+    //=========================================================================
+    // AtkLive
+    //=========================================================================
+    pub fn atk_live_get_type() -> GType;
```
bilelmoussaoui, line 1830: Needs a since annotation otherwise things will break, I opened a PR for it upstream so :/

gdk/sys/Cargo.toml
```diff
@@ -56,6 +56,7 @@ rust-version = "1.70"
 [package.metadata.docs.rs]
 rustdoc-args = ["--cfg", "docsrs"]
 features = []
```
bilelmoussaoui, line 58: this should be removed, actually if you removed the whole section gir will generate the correct thing

atk/sys/src/lib.rs
bilelmoussaoui: Also as there is a new type, maybe it is worth adding to the safe bindings as well & it would need a since annotation there as well

atk/sys/Cargo.toml
```diff
@@ -22,6 +22,7 @@ v2_32 = ["v2_30"]
 v2_34 = ["v2_32"]
 v2_38 = ["v2_34"]
 v2_46 = ["v2_38"]
+v2_50 = ["v2_46"]
```
bilelmoussaoui, line 25: the new feature has to be exposed in the safe wrapper

atk/sys/src/lib.rs
GuillaumeGomez: Gonna add the new type (even though it doesn't seem to be used anywhere?) and add the missing cfg.

gdk/sys/Cargo.toml
GuillaumeGomez: Let's keep things as is for now.

PR #844, opened 2023-09-03
Properly use the Propagation values

pbor:
In the shuffle from Inhibit(bool), to ControlFlow, to Propagtaion we ended up flipping some value. In particular in the widget subclass code and in the example.


bilelmoussaoui:
the ci failures are unrelated so let us land these

PR #847, opened 2023-11-01
Use async-channel in the examples

pbor:
glib's MainContext::channel was removed from glib-rs


examples/cairo_threads/main.rs
```diff
@@ -117,18 +117,18 @@ fn build_ui(application: &gtk::Application) {
         }),
     );
 
-    ready_rx.attach(None, move |(thread_num, image)| {
-        let (ref images, ref origins, ref workers) = *workspace;
+    glib::MainContext::ref_thread_default().spawn_local(async move {
```
sdroege, line 120: This is GTK, you should use the global default main context for scheduling UI things

examples/multi_threading_context/main.rs
```diff
@@ -28,27 +28,28 @@ fn build_ui(application: &gtk::Application) {
     scroll.set_policy(gtk::PolicyType::Automatic, gtk::PolicyType::Automatic);
     scroll.add(&text_view);
 
-    let (tx, rx) = glib::MainContext::channel(glib::Priority::default());
+    let (tx, rx) = async_channel::unbounded();
```
sdroege, line 31: Should probably not be unbounded?

examples/cairo_threads/main.rs
pbor: I know, but the result is the same, and I just use this out of habit... will change it

examples/multi_threading_context/main.rs
pbor: Dunno... I do not like to use a value pulled out of this air if I do not understand the implications and these are just examples. Anyway I guess I will just change that to 10

sdroege: I think here it does not matter much, but any bounded value is usually better than it growing infinitely.

PR #857, opened 2026-08-13
Update gtk-rs-core deps to 0.22, update gir/gir-files

kelnos:
This updates the code to be compatible with the 0.22.z releases of the various gtk-rs-core crates.  It also updates the gir tool and gir-files submodules, and regenerates all the source.

For now I've pinned the gtk-rs-core deps to the crates.io 0.22 releases; I notice that the convention here (and in gtk4-rs) seems to be to pin them to the git repository with a version specified.  If that's what I should do here instead, let me know, and I'll update.

This bumps the MSRV to 1.92 and fixes all clippy warnings.

I've also started migrating to a more "modern" way of declaring the dependencies in the workspace Cargo.toml, and then referring to them in the sub-crates via the workspace.  There's still more that can be done here with other properties that could be declared in the workspace, but it's a start.

Finally, this includes a bunch of updates to the github workflows to bump to modern versions and fix failures.


atk/sys/Cargo.toml
```diff
@@ -6,11 +6,11 @@ libc = "0.2"
 
 [dependencies.glib]
 package = "glib-sys"
-git = "https://github.com/gtk-rs/gtk-rs-core"
+version = "0.22"
```
sdroege: This is different to what we do in the other repos but maybe this is OK here as it's just in maintenance mode and not active development? @bilelmoussaoui ?

Elsewhere we track git deps in the main branch and the release branches track versions (but also via git, replaced at release time with plain crates.io via `cargo publish`).

examples/composite_template/example_application_window/mod.rs
```diff
@@ -10,7 +10,7 @@ glib::wrapper! {
 }
 
 impl ExampleApplicationWindow {
-    pub fn new<P: glib::IsA<gtk::Application>>(app: &P) -> Self {
+    pub fn new<P: glib::object::IsA<gtk::Application>>(app: &P) -> Self {
```
sdroege: Just `IsA` should be sufficient?

examples/Cargo.toml
```diff
@@ -9,25 +9,13 @@ async-channel = "2.0.0"
 chrono = "0.4"
 futures = "0.3"
 futures-util = "0.3"
+glib.workspace = true
+gtk = { path = "../gtk" }
+pangocairo = { version = "0.22", optional = true }
+cairo = { package = "cairo-rs", version = "0.22", features = ["png"], optional = true }
```
sdroege: Should move all this to workspace dependencies

gdk/src/display.rs
```diff
@@ -2,7 +2,7 @@
 
 use crate::Display;
 use glib::object::IsA;
-use glib::ObjectExt;
+use glib::prelude::ObjectExt;
```
sdroege: `prelude::*` maybe

gdk/src/functions.rs
```diff
@@ -2,7 +2,7 @@
 
 use glib::translate::*;
 
-#[repr(packed)]
+#[repr(Rust, packed)]
```
sdroege: This wants `repr(C, packed)` I think?

.github/workflows/docs.yml
```diff
@@ -36,14 +36,22 @@ jobs:
       - run: cargo install rustdoc-stripper
       - run: python3 ./generator.py --embed-docs --yes ./
       - run: git clone https://gitlab.gnome.org/World/Rust/gir-rustdoc/ # checkout action doesn't support random urls
+      - run: |
+          sed -i 's|archive.ubuntu.com|old-releases.ubuntu.com|g; s|security.ubuntu.com|old-releases.ubuntu.com|g' /etc/apt/sources.list
+          apt-get update
+          apt-get install -y python3-requests
```
sdroege, line 42: Did you check if the CI workflows (all of them) are matching approximately what is now in gtk4-rs? Might make sense to sync them

Cargo.toml
```diff
@@ -26,3 +26,19 @@ members = [
 ]
 
 exclude = ["gir", "checker"]
+resolver = "2"
+
+[workspace.dependencies]
```
sdroege, line 31: We also include a `Cargo.lock` in the repo now. Can you add one here?

sdroege (commented):
Looks mostly good to me, thanks :)

atk/sys/Cargo.toml
kelnos: Yeah, I wasn't sure about that.

I see that gtk4-rs pins to both the git repo and the current unreleased 0.23.0-alpha version, without specifying a git rev or branch.  I wasn't sure what that actually means from the perspective of what cargo will resolve, so I just looked it up... apparently that means it will pull the git source (obviously, heh) at whatever is the tip of the default branch, and then verify that the version there matches the specified version (and throw an error if not).

gtk3-rs has Cargo.lock gitignore'd, so it's not checked in (different from gtk4-rs, which does check it in), so people who clone the repo and do their first build at different times can end up building against a different git rev of the gtk-rs-core crates, which is not ideal.  So at the very least, I should change gtk3-rs to include Cargo.lock in git.

But even then, Cargo.lock only affects someone who pulls the git repo and builds the crate directly; gtk3-rs's Cargo.lock is ignored when another crate builds that depends on gtk3-rs.

Then there's the question of timing: when do y'all plan on making the first 0.23.0 stable release?  If that's within the next month or so, I guess that's fine, and my first gtk3-rs release could depend on 0.23.0 instead of 0.22.  If it's not going to be for a long time, then I'd rather pin to 0.22 so I can release against something stable, in a shorter amount of time.  (I'll be away traveling for a couple weeks, so realistically my first gtk3-rs release probably wouldn't be until later in September.)

Anyhow, still not sure here.  I'm fine switching it to the git source if that's what you'd prefer.

.github/workflows/docs.yml
kelnos: I started doing that, but I can't match them exactly, because the gtk3-rs build container is very old, and is based on Ubuntu 20.10.  It looks like y'all switched to Fedora at some point, and presumably it's currently a much more up-to-date Fedora.  My plan was to revisit that after I'm able to update the build container.

gdk/src/functions.rs
kelnos: Ah yes, since it's used for FFI it has to be...

.github/workflows/docs.yml
sdroege: Ok, that makes sense :)

examples/composite_template/example_application_window/mod.rs
kelnos: Ah yeah, this was probably just a result of `sed -e 's/glib::IsA/glib::object::IsA/g'`, but that particular line doesn't need it to be fully-qualified.

examples/Cargo.toml
kelnos: Ah yeah, at that point in my changes I didn't realize you could do `workspace = true` and also override `features`/`optional`.  I've moved the pangocairo/cairo deps to use the workspace deps, but for now I'll leave the other deps as-is and will take care of them in a future pass, since there are quite a few other deps that probably need to be moved as well.  (For this PR I just focused on consolidating the gtk-rs-core deps, plus cairo.)

Cargo.toml
kelnos: Ah right, I mentioned in my other comment about Cargo.lock not being checked in here, when it is for gtk4-rs.  Another thing I was planning to do for future work, but it's simple to do it here now.

atk/sys/Cargo.toml
bilelmoussaoui: I would still track git for the sake of easily testing any potential future gtk-rs-core breaking changes.

.github/workflows/docs.yml
bilelmoussaoui: yes we are using a container for gtk4-rs that builds on top of the image generated for gtk-rs-core, if you can do the same, that would be great.

bilelmoussaoui:
nothing to comment about, lgtm. It would be nice if you can work on bringing some of the features we worked on for gtk-rs-core/gtk4-rs like
- automatic regen after updates of gir/gir-files
- a base container image that all CI jobs use instead of reinstalling deps all over the place
- keeping up with our release cycle

atk/sys/Cargo.toml
kelnos: Sure thing, updated.  Fortunately there are no breaking changes yet, so I didn't have to make any other changes :)

sdroege: Once this is in and all cleanup is done, we can also do a 0.22 branch and do a release from that I guess.

examples/Cargo.toml
sdroege: Can be done after this is in IMHO, let's just not forget it

kelnos:
> nothing to comment about, lgtm. It would be nice if you can work on bringing some of the features we worked on for gtk-rs-core/gtk4-rs like
> 
>     * automatic regen after updates of gir/gir-files

Will do in a future commit.  I'd also like to harmonize the rest of the GH actions jobs where it makes sense.  Got a little burned out on CI for now after spending hours getting this PR green the other night, so I needed a break :-P

>     * a base container image that all CI jobs use instead of reinstalling deps all over the place

Yeah, I'd like to get this done as soon as possible; I have plenty of experience fighting with an out of date build container, so getting that fixed up soon will be a priority.

>     * keeping up with our release cycle

Let's chat about this on Matrix at some point, since I don't know what your release cycle is :)

examples/Cargo.toml
kelnos: Yep, I'll create some GH issues to track some of this work.

kelnos:
Ok, I think I've addressed all the comments.  I also created 3 issues to track future work that we've discussed here.  Anything else?

djc:
Should we conclude that gtk3 crates are now being maintained again? See

- https://github.com/rustsec/advisory-db/pull/3189

sdroege:
That's the goal. The plan is to release something with updated dependencies, and then (as long as @kelnos stays around) keep it updated at least.

PR #872, opened 2026-08-18
gdkx11: add raw pointer getters for libX11 types

### LINKED ISSUE #346: [gdkx11] Get pointer to xlib xdisplay
I'm trying to use [GDK_DISPLAY_XDISPLAY()](https://developer.gnome.org/gdk3/stable/gdk3-X-Window-System-Interaction.html#GDK-DISPLAY-XDISPLAY:CAPS) or [gdk_x11_display_get_xdisplay ()](https://developer.gnome.org/gdk3/stable/gdk3-X-Window-System-Interaction.html#gdk-x11-display-get-xdisplay) but I can't find a way to get a raw pointer to the xlib display using gdkx11.

The purpose of this is to use it in [raw-window-handle](https://docs.rs/raw-window-handle/0.3.3/raw_window_handle/unix/struct.XlibHandle.html)

How can I do it?

Thanks!

kelnos:
Closes #346

### DIFF
--- a/gdkx11/src/x11_display.rs
+++ b/gdkx11/src/x11_display.rs
@@ -1,0 +1,14 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use crate::{X11Display, ffi};
+
+use glib::translate::ToGlibPtr;
+use x11::xlib;
+
+impl X11Display {
+    #[doc(alias = "gdk_x11_display_get_xdisplay")]
+    #[allow(clippy::missing_safety_doc)]
+    pub unsafe fn xdisplay(&self) -> *mut xlib::Display {
+        unsafe { ffi::gdk_x11_display_get_xdisplay(self.to_glib_none().0) }
+    }
+}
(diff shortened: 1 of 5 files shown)


gdkx11/src/x11_display.rs
```diff
@@ -0,0 +1,14 @@
...
+
+use crate::{X11Display, ffi};
+
+use glib::translate::ToGlibPtr;
+use x11::xlib;
+
+impl X11Display {
+    #[doc(alias = "gdk_x11_display_get_xdisplay")]
+    #[allow(clippy::missing_safety_doc)]
+    pub unsafe fn xdisplay(&self) -> *mut xlib::Display {
```
bilelmoussaoui, line 11: I don't think those should be unsafe btw

kelnos: Ah hm, I was just copying what gdk4-x11 does.  I suppose they're not unsafe in the sense that simply calling it can't cause a crash or undefined behavior, but I think there is a precedent in Rust for marking functions unsafe as a sort of warning that the caller needs to be careful with what the function returns.

But I'm fine dropping the unsafe if you feel strongly about it.

